Issue #29, opened 2019-03-28
Azure VMs are created with OS disk of type Standard HDD 


sidoruka:
Currently Azure VM instances are always created with OS disk of type Standard HDD and data disk of type Premium SSD
HDD type of the OS disk may affect the node initialization performance
Would be nice to set OS disk to Premium SSD the same as data disk. 
E.g. with azure-cli it can be done via `--storage-sku Premium_LRS`, which will set Premium SSD to all disks

PR #35, opened 2019-03-28
SSD disk type for Azure OS disks

SilinPavel:
### Summary
This PR resolves Issue #29 
Now SSD disk will be applied as OS disk to the instance if instance supports it


sidoruka:
I've just tried that change and VMs are created correctly with SSD OS disk.
Though can't see much difference in the bootup performance, compared to HDD OS disk. But this can be a separate issue

Issue #10, opened 2019-03-25
Implement support for Google Cloud Storage in CP Java API


mzueva:
Provide implementation of `AbstractDataStorage` and `StorageProvider` classes to support all CP API operations with Google Cloud Storage.

mzueva:
Major part of work is done

PR #59, opened 2019-04-02
issue #10: Implement support for Google Cloud Storage in CP Java API …

ekazachkova:
Provides basic operations for data storage operations with Google Storage API

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPClient.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPClient.java
@@ -16,6 +16,7 @@
 
 package com.epam.pipeline.manager.cloud.gcp;
 
+import com.epam.pipeline.entity.datastorage.DataStorageException;
 import com.epam.pipeline.entity.region.GCPRegion;
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
@@ -25,6 +26,10 @@ import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.services.cloudbilling.Cloudbilling;
 import com.google.api.services.compute.Compute;
 import com.google.api.services.iamcredentials.v1.IAMCredentials;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPClient.java
```diff
@@ -74,6 +79,23 @@ public IAMCredentials buildIAMCredentialsClient(final GCPRegion region) throws I
                 .build();
     }
 
+    public Storage buildStorageClient(final GCPRegion region) {
+        if (StringUtils.isBlank(region.getAuthFile())) {
+            return StorageOptions.getDefaultInstance().getService();
+        }
+        try (InputStream stream = new FileInputStream(region.getAuthFile())) {
+            final GoogleCredentials sourceCredentials = ServiceAccountCredentials
```
mzueva: Can't we just use `buildCredentials` method here? I think it will work fine for both: application default credentials and auth_file?

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/gcp/GSBucketStorageHelper.java
```diff
@@ -55,35 +71,192 @@ public String createGoogleStorage(final GSBucketStorage storage) {
...
+        );
+        return new DataStorageListing(blobs.getNextPageToken(), items);
+    }
+
+    public DataStorageFile createFile(final GSBucketStorage storage, final String path, final byte[] contents) {
+        final Storage client = gcpClient.buildStorageClient(region);
+
+        final String bucketName = storage.getPath();
+        checkBucketExists(bucketName, client);
+        checkBlobDoesNotExist(bucketName, path, client);
```
mzueva: Please verify that other clients (S3/AZ) do not allow to overwrite file content, it think this method is also used for update operation.

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/gcp/GSBucketStorageHelper.java
```diff
@@ -55,35 +71,192 @@ public String createGoogleStorage(final GSBucketStorage storage) {
...
+                                        final Integer pageSize, final String marker) {
+        final List<AbstractDataStorageItem> items = new ArrayList<>();
+
+        String requestPath = Optional.ofNullable(path).orElse(EMPTY_PREFIX);
+        if (StringUtils.isNotBlank(requestPath)) {
+            requestPath = normalizeFolderPath(requestPath);
         }
+        final Storage client = gcpClient.buildStorageClient(region);
+        final String bucketName = storage.getPath();
+        checkBucketExists(bucketName, client);
```
mzueva: Do we need to check that bucket exists for each method? Maybe `StorageClient` will throw appropriate exception itself?

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/gcp/GSBucketStorageHelper.java
```diff
@@ -55,35 +71,192 @@ public String createGoogleStorage(final GSBucketStorage storage) {
...
+
+    public DataStorageFile createFile(final GSBucketStorage storage, final String path, final byte[] contents) {
+        final Storage client = gcpClient.buildStorageClient(region);
+
+        final String bucketName = storage.getPath();
+        checkBucketExists(bucketName, client);
+        checkBlobDoesNotExist(bucketName, path, client);
+
+        final BlobId blobId = BlobId.of(bucketName, path);
+        final BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType(TEXT_CONTENT_TYPE).build();
```
mzueva: Is `contentType` required? I'm not sure that it is OK to always set it to `TEXT`

PR #83, opened 2019-04-08
Issue #61: resolve azure blob service domain name to /etc/hosts

### LINKED ISSUE #61: Very slow operations on the BLOB Fuse mounted directories
* Region: westeurope
* BLOB contents: 1 file
* `ls -l`: ~1min+
* `echo 1 > 1.txt`: ~1min+

SilinPavel:
Summary
This RP resolves #61.
Now resolved ip addresses for Azure Blob will be automatically added to /etc/hosts during mounting storages task

### DIFF
--- a/workflows/pipe-common/scripts/mount_storage.py
+++ b/workflows/pipe-common/scripts/mount_storage.py
@@ -247,12 +247,13 @@ class AzureMounter(StorageMounter):
             AzureMounter.fuse_tmp = fuse_tmp
 
     def mount(self, mount_root, task_name):
-        super(AzureMounter, self).mount(mount_root, task_name)
         # add resolved ip address for azure blob service to /etc/hosts (only once per account_name)
         account_name, _, _ = self._get_credentials(self.storage)
-        command = "grep {account_name} /etc/hosts || getent hosts {account_name}.blob.core.windows.net ".format(account_name=account_name) \
-                  + "| awk '{ printf \"%s\t%s\\n\", $1, $3 }' >> /etc/hosts"
+        command = 'etc_hosts_clear="$(sed -E \'/.*{account_name}.blob.core.windows.net.*/d\' /etc/hosts)" ' \
+                  '&& cat > /etc/hosts <<< "$etc_hosts_clear" ' \
+                  '&& getent hosts {account_name}.blob.core.windows.net >> /etc/hosts'.format(account_name=account_name)
         common.execute_cmd_command(command, silent=True)
+        super(AzureMounter, self).mount(mount_root, task_name)
 
     def build_mount_params(self, mount_point):
         account_id, account_key, _ = self._get_credentials(self.storage)
--- a/workflows/pipe-common/scripts/mount_storage.py
+++ b/workflows/pipe-common/scripts/mount_storage.py
@@ -246,6 +246,14 @@ class AzureMounter(StorageMounter):
         if StorageMounter.create_directory(fuse_tmp, task_name):
             AzureMounter.fuse_tmp = fuse_tmp
 
+    def mount(self, mount_root, task_name):
+        super(AzureMounter, self).mount(mount_root, task_name)
+        # add resolved ip address for azure blob service to /etc/hosts (only once per account_name)
+        params = self.build_mount_params(mount_root)
+        command = "grep {account_name} /etc/hosts || getent hosts {account_name}.blob.core.windows.net ".format(**params) \
+                  + "| awk '{ printf \"%s\t%s\\n\", $1, $3 }' >> /etc/hosts"
+        common.execute_cmd_command(command, silent=True)
+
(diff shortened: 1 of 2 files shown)


workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -246,6 +246,14 @@ def init_tmp_dir(tmp_dir, task_name):
         if StorageMounter.create_directory(fuse_tmp, task_name):
             AzureMounter.fuse_tmp = fuse_tmp
 
+    def mount(self, mount_root, task_name):
+        super(AzureMounter, self).mount(mount_root, task_name)
+        # add resolved ip address for azure blob service to /etc/hosts (only once per account_name)
+        params = self.build_mount_params(mount_root)
+        command = "grep {account_name} /etc/hosts || getent hosts {account_name}.blob.core.windows.net ".format(**params) \
```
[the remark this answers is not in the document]
mzueva: Also please check, that this won't break anything if `getent` returns some error instead of expected output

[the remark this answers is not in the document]
SilinPavel: yes it will, since we initialize mounter for each storage individually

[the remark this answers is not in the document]
SilinPavel: I also check that if something going wrong with getent no line will be added to /etc/hosts but mount_storages.py will proceed

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -246,6 +246,14 @@ def init_tmp_dir(tmp_dir, task_name):
         if StorageMounter.create_directory(fuse_tmp, task_name):
             AzureMounter.fuse_tmp = fuse_tmp
 
+    def mount(self, mount_root, task_name):
+        super(AzureMounter, self).mount(mount_root, task_name)
+        # add resolved ip address for azure blob service to /etc/hosts (only once per account_name)
+        account_name, _, _ = self._get_credentials(self.storage)
+        command = "grep {account_name} /etc/hosts || getent hosts {account_name}.blob.core.windows.net ".format(account_name=account_name) \
```
sidoruka: What if ip address changes since last addition to the /etc/hosts? 
E.g.:
1. We ran some tool
2. IP was added to /etc/hosts
3. We paused/resumed a run
4. IP is changed, but we still have the old one in /etc/hosts

Is it handled?

workflows/pipe-common/shell/install_azure_fuse_blobfuse
```diff
@@ -51,6 +64,16 @@ function install_from_package_repository {
 _FUSE_TYPE=blobfuse
 _FUSE_BIN=/usr/bin/blobfuse
 
+command -v getent > /dev/null 2>&1
+_IS_GCLIB_INSTALLED=$?
+if [ $_IS_GCLIB_INSTALLED -ne 0 ]; then
```
sidoruka: typo: _IS_GCLIB_INSTALLED -> _IS_GLIBC_INSTALLED

sidoruka: Maybe `if [ $? -ne 0 ]; then` is sufficient?

workflows/pipe-common/scripts/mount_storage.py
SilinPavel: No it's not, we can do replacement here instead of current solution.

SilinPavel:
@sidoruka please take a look

workflows/pipe-common/shell/install_azure_fuse_blobfuse
```diff
@@ -19,6 +19,19 @@ function get_os_distribution_parameter {
     echo $(cat /etc/os-release | grep "^$_PARAMETER_NAME=" | awk -F '=' '{gsub(/"/, "", $2); print $2}')
 }
 
+function install_glibc_from_package_repository {
+    /usr/bin/rpm -q -f /usr/bin/rpm >/dev/null 2>&1
+    _IS_RPM_BASED=$?
+
+    if [ $_IS_RPM_BASED -eq 0 ]; then
```
sidoruka: Once again: maybe `$?` is sufficient here

workflows/pipe-common/shell/install_azure_fuse_blobfuse
```diff
@@ -51,6 +64,15 @@ function install_from_package_repository {
 _FUSE_TYPE=blobfuse
 _FUSE_BIN=/usr/bin/blobfuse
 
+command -v getent > /dev/null 2>&1
+if [ $? -ne 0 ]; then
+    install_glibc_from_package_repository
+    if [ $? -ne 0 ]; then
```
sidoruka: I this will work: `if ! install_glibc_from_package_repository; then`

PR #127, opened 2019-04-12
fix: GCP Bucket listing shows previous versions as deleted

### LINKED ISSUE #107: GCP Bucket listing shows previous versions as deleted
API returns previous file versions as `deleted`:
Screen Shot 2019-04-10 at 20 24 12

Steps to reproduce:

1. Create gs buckets with enabled versioning
2. Upload file
3. Change file and upload it again
4. Enable `Show files versions` check

CLI works fine for the same case

ekazachkova:
Resolves #107

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/datastorage/providers/gcp/GSBucketStorageHelper.java
+++ b/api/src/main/java/com/epam/pipeline/manager/datastorage/providers/gcp/GSBucketStorageHelper.java
@@ -172,7 +172,7 @@ public class GSBucketStorageHelper {
         }
 
         final Blob blob = checkBlobExistsAndGet(bucketName, path, client, version);
-        deleteBlob(bucketName, path, blob);
+        deleteBlob(blob, client, StringUtils.isNotBlank(version));
     }
 
     public void deleteFolder(final GSBucketStorage dataStorage, final String path, final Boolean totally) {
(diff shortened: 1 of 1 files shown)


mzueva:
@ekazachkova Problem still in place. If files is deleted, all it's versions are shown as delete markers. Since GS doesn't return a separate object for delete events (like `deletion marker` in AWS), I assume we should use the same logic as in cli: if current state of blob is deleted just add a synthetic `delete` version, while showing all other versions as non-deleted.

Also please check blob deletion: when I delete file from GUI the latest version is completely lost (not moved to previous versions), `pipe storage rm` works as expected: deleting file form a versioned bucket saves latest version as previous one.

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/gcp/GSBucketStorageHelper.java
```diff
@@ -471,11 +473,14 @@ private void deleteAllVersions(final String bucketName, final String path, final
         final Page<Blob> blobs = client.list(bucketName,
                 Storage.BlobListOption.versions(true),
                 Storage.BlobListOption.prefix(path));
-        blobs.iterateAll().forEach(blob -> deleteBlob(bucketName, path, blob));
+        blobs.iterateAll().forEach(blob -> deleteBlob(blob, gcpClient.buildStorageClient(region), true));
```
mzueva: Can't we reuse client from methods argument?

ekazachkova: fixed

Issue #151, opened 2019-04-16
BLOB Fuse cannot be installed to the Debian containers


sidoruka:
BLOB Fuse is not available as a pre-built binary for the `debian` distr.
`Debian` is used for some of the default docker images, e.g. `RStudio` and `Bioconductor`

Please also add some error logging to the storage mount routine, so we know if something goes wrong.

sidoruka:
Verified using Debian 9-based RStudio:
- Error logging added
- Installation is done correctly

PR #161, opened 2019-04-18
Issue 61: fix for ubuntu and debian images

SilinPavel:
## Summary
This PR addresses BLOB service name resolution issue #61 on Ubuntu and Debian images

### Implemented steps
- Extend logging
- In launch.sh added section that force create symbol link from /bin/sh to /bin/bash

### DIFF
--- a/workflows/pipe-common/scripts/mount_storage.py
+++ b/workflows/pipe-common/scripts/mount_storage.py
@@ -156,7 +156,10 @@ class StorageMounter:
 
     @staticmethod
     def execute_and_check_command(command):
-        install_check = common.execute_cmd_command(command, silent=False)
+        install_check, _, stderr = common.execute_cmd_command_and_get_stdout_stderr(command, silent=False)
+        if install_check != 0:
+            Logger.warn(
+                'Installation script {command} failed: \n {stderr}'.format(command=command, stderr=stderr))
         return install_check == 0
 
     @staticmethod
@@ -281,7 +284,7 @@ class AzureMounter(StorageMounter):
         exit_code, _, stderr = common.execute_cmd_command_and_get_stdout_stderr(command, silent=True)
         if exit_code != 0:
             Logger.warn(
-                'Azure BLOB service hostname resolution and writing to /etc/hosts failed:: \n {}'.format(stderr))
+                'Azure BLOB service hostname resolution and writing to /etc/hosts failed: \n {}'.format(stderr))
 
 
 class S3Mounter(StorageMounter):
--- a/workflows/pipe-common/scripts/mount_storage.py
+++ b/workflows/pipe-common/scripts/mount_storage.py
@@ -281,7 +281,7 @@ class AzureMounter(StorageMounter):
         exit_code, _, stderr = common.execute_cmd_command_and_get_stdout_stderr(command, silent=True)
         if exit_code != 0:
             Logger.warn(
-                'Something goes wrong with resolving azure blob service hosts to /etc/hosts: \n {}'.format(stderr))
+                'Azure BLOB service hostname resolution and writing to /etc/hosts failed:: \n {}'.format(stderr))
 
 
 class S3Mounter(StorageMounter):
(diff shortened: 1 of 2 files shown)


workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -277,6 +272,17 @@ def build_mount_command(self, params):
...
+    def __resolve_azure_blob_service_url(self):
+        # add resolved ip address for azure blob service to /etc/hosts (only once per account_name)
+        account_name, _, _ = self._get_credentials(self.storage)
+        command = 'etc_hosts_clear="$(sed -E \'/.*{account_name}.blob.core.windows.net.*/d\' /etc/hosts)" ' \
+                  '&& cat > /etc/hosts <<< "$etc_hosts_clear" ' \
+                  '&& getent hosts {account_name}.blob.core.windows.net >> /etc/hosts'.format(account_name=account_name)
+        exit_code, _, stderr = common.execute_cmd_command_and_get_stdout_stderr(command, silent=True)
+        if exit_code != 0:
+            Logger.warn(
+                'Something goes wrong with resolving azure blob service hosts to /etc/hosts: \n {}'.format(stderr))
```
sidoruka: 'Azure BLOB service hostname resolution and writing to /etc/hosts failed: ...'

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -277,6 +272,17 @@ def build_mount_command(self, params):
...
                '{mount_options}'.format(**params)
 
+    def __resolve_azure_blob_service_url(self):
+        # add resolved ip address for azure blob service to /etc/hosts (only once per account_name)
+        account_name, _, _ = self._get_credentials(self.storage)
+        command = 'etc_hosts_clear="$(sed -E \'/.*{account_name}.blob.core.windows.net.*/d\' /etc/hosts)" ' \
+                  '&& cat > /etc/hosts <<< "$etc_hosts_clear" ' \
+                  '&& getent hosts {account_name}.blob.core.windows.net >> /etc/hosts'.format(account_name=account_name)
+        exit_code, _, stderr = common.execute_cmd_command_and_get_stdout_stderr(command, silent=True)
+        if exit_code != 0:
```
sidoruka: Is it the only place when shall handle shell errors?
E.g. we've experienced an issue with blobfuse installation for Debian (in the background it threw 404) and it was hard to troubleshoot without any log message

SilinPavel: In this script we already have checks for other places where shell is executed:
 - execute_mount method
 - create_directory method
 - execute_and_check_command method

sidoruka: In this case, what log will I get if blobfuse installation fails for whatever reason? E.g. when the following is executed:
https://github.com/epam/cloud-pipeline/blob/2dd2eae60f3fb8779a74e7751014fc64a79b79ae/workflows/pipe-common/scripts/mount_storage.py#L237

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -281,7 +281,7 @@ def __resolve_azure_blob_service_url(self):
         exit_code, _, stderr = common.execute_cmd_command_and_get_stdout_stderr(command, silent=True)
         if exit_code != 0:
             Logger.warn(
-                'Something goes wrong with resolving azure blob service hosts to /etc/hosts: \n {}'.format(stderr))
+                'Azure BLOB service hostname resolution and writing to /etc/hosts failed:: \n {}'.format(stderr))
```
sidoruka: Would be nice if you double-check the spelling yourself... `::` -> `:`

workflows/pipe-common/scripts/mount_storage.py
SilinPavel: lines from stdout and stderr will be printed by python process to the current shell, this won't be appear in a pipeline logs. Do you refer me to #151 and propose to resolve this issue in the current PR?

sidoruka: Yes

SilinPavel: ok, done, now stderr of all installation script (install_s3_fuse_goofys, install_azure_fuse_blobfuse etc) will be passed to pipeline as warn logs

Issue #188, opened 2019-04-22
Allow to terminate paused runs


sidoruka:
**BACKGROUND**

At the moment, when a run is paused - the user cannot terminate/stop it before resuming. This is the expected behavior. But in certain "bad" cases - it is not possible to resume a run:
1. An underlying instance is terminated outside of the Cloud Pipeline
2. Docker image was removed from the registry
3. And other cases that are not yet uncovered

This introduces a number of stale runs, that just sit there in the PAUSED state and nobody can remove them.

**TODOs**

1. For the PAUSED runs we shall allow OWNER or the run and the ADMIN users to terminate it
2. Termination of the PAUSED run - shall drop the underlying cloud instance and mark the run as STOPPED
3. From the GUI perspective - we shall not hide the STOP button, but replace it with TERMINATE, when a run is PAUSED (it shall not be available for the interim phases, e.g. PAUSING/RESUMING)

mzueva:
@tcibinan Add method `run/${id}/terminate` that will terminate underlying cloud instance and change run status to `STOPPED`. Method shall be available only for runs in status `PAUSED`.

mzueva:
@rodichenko API method `POST /run/{runId}/terminate` implemented and deployed on dev

sidoruka:
@rodichenko, paused run termination is a relatively long running operation (~10 seconds).
While it is in progress - please disable the `Cancel`/`Terminate` controls, as this will confuse users.

E.g. I can click `Cancel` while run is being terminated
![image](https://user-images.githubusercontent.com/20998291/57798362-3b295200-7755-11e9-85c8-77b63cd2f7d7.png)

rodichenko:
@sidoruka `Cancel` and `Terminate` buttons are disabled now while termination in progress (via 863ae252f4fc94e123dae9858250f4c7fad62b76)

PR #222, opened 2019-04-29
Add support for paused runs termination

tcibinan:
Resolves issue #188 in backend.

The pull request brings support for paused runs termination. Run termination means termination of its instance node if it exists and changing the run status to `STOPPED`. See #188 for more info.

Also the pull request removes unused `endDate` argument in several run status update methods.


mzueva (changes requested):
Logic for termination shall be the following: if node is not found in k8s (this will be always true for properly paused runs), we should try to delete it using `instance` fields in `PipelineRun`

- check k8s
- check `PipelineRun`
- try to find instance by `runid` label from `CloudProvider`

api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
```diff
@@ -159,31 +160,90 @@ public NodeInstance getNode(String name) {
...
+        if (node.isPresent()) {
+            log.debug("Kubernetes node {} for run {} was found and will be terminated.", node.get().getId(),
+                    run.getId());
+            terminateNode(node.get());
+        } else {
+            log.debug("Kubernetes node for run {} wasn't found and its termination will be skipped.", run.getId());
+            final AbstractCloudRegion region = instance.map(RunInstance::getCloudRegionId)
+                    .map(regionManager::load)
+                    .orElseGet(regionManager::loadDefaultRegion);
+            final Optional<String> nodeId = instance.map(RunInstance::getNodeId)
```
mzueva, line 216: If nodeId is not set try to find and terminate VM by `runid` label

mzueva (changes requested):

mzueva (approved):

Issue #192, opened 2019-04-23
Consider Cloud Providers' resource limitations, when scheduling a job


sidoruka:
**BACKGROUND**

Cloud Pipeline supports queueing of the jobs, that cannot be scheduled to the existing or new compute nodes immediately.
Queueing occurs if:
1. `cluster.max.size` limit is reached
2. Cloud Provider limitations are reached

If (1) happens - run will sit in the `init` state, until a spare will be available or stopped manually. This is the correct behavior.

But if (2) occurs - `nodeup.py` will receive an error from AWS and raise an exception, treating this as an issue with the node creation. Autoscaler will resubmit the node creation task for `cluster.nodeup.retry.count` times (default: 5) and then fail the job run.

This behavior confuses users, as (2) shall behave in the same manner as (1) - job shall be kept in the queue until there will be free space for a new node.

**TODOs**

1. `nodeup.py` shall handle instance/VM type/cores/RAM/SSD/etc. limitation exception for all three cloud providers (consider that Spots/Preemtibles may produce a different exception)
2. If such an exception is thrown - `nodeup.py` shall gracefully report such status to the autoscaler
3. Autoscaler shall not decrease current `cluster.nodeup.retry.count` counter value for such statuses
4. `nodeup.py` shall still report the exception to job log, so we can troubleshoot limits if needed

sidoruka:
@mzueva , regarding the priorities across cloud providers - let's first implement for the `develop` (aws/azure) and then port the changes into GCP as well.

Hope there will be not much of such modifications, that require changes across different branches...

mzueva:
@ekazachkova Please provide implementation of limit handling for AWS. 

1. Introduce exit code to `nodeup` script that will indicate that cluster scaling failed due to exceeding AWS instance quotas/limits. If `nodeup` process exits with this exit code, `AutoscaleManager` shall not increase retry count for runid and continue attempts to scale node up. 
2. Implement corresponding  changes to `nodeup` scripts for AWS provider. Do not forget to handle spot and ondemand instances creation.

For testing purposes I've requested to set limit for on-demand an to spot instances with type **m4.large** to 2 instances each. Also please consider that we should handle limit for total number of spot requests: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-spot-limits.html

PR #238, opened 2019-05-06
Consider Cloud Providers' resource limitations, when scheduling a job - implementation for Azure

ekazachkova:
Implements issue #192 for Azure provider.  

- `nodeup.py` script logs error and returns exit code 6 if instance limit exceeded

Also, contains fix for AWS implementation: throw initial exception if an error occurred but not a quota problem.

### DIFF
--- a/scripts/autoscaling/azure/nodeup.py
+++ b/scripts/autoscaling/azure/nodeup.py
@@ -16,6 +16,8 @@ import argparse
 import functools
 import os
 import json
+import sys
+
 import pykube
 import logging
 import fnmatch
@@ -27,6 +29,7 @@ from azure.common.client_factory import get_client_from_auth_file
 from azure.mgmt.resource import ResourceManagementClient
 from azure.mgmt.compute import ComputeManagementClient
 from azure.mgmt.network import NetworkManagementClient
+from msrestazure.azure_exceptions import CloudError
 from pipeline import Logger, TaskStatus, PipelineAPI
 
 VM_NAME_PREFIX = "az-"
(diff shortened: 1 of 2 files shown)


scripts/autoscaling/azure/nodeup.py
```diff
@@ -388,15 +410,23 @@ def create_vm(instance_name, run_id, instance_type, instance_image, disk, user_d
...
+            instance_name,
+            vm_parameters
+        )
+        creation_result.result()
+
+        start_result = compute_client.virtual_machines.start(resource_group_name, instance_name)
+        start_result.wait()
+    except CloudError as client_error:
+        error_message = client_error.__str__()
+        if 'OperationNotAllowed' in error_message or 'ResourceQuotaExceeded' in error_message:
```
mzueva, line 425: Maybe we should check for `AND` condition instead of `OR` in an error message , I'm not sure whether  `OperationNotAllowed` will always refer to quotas problem?

mzueva (changes requested):

scripts/autoscaling/azure/nodeup.py
ekazachkova: According to azure [docs](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-common-deployment-errors) `OperationNotAllowed` error code means:
 `The deployment is attempting an operation that exceeds the quota for the subscription, resource group, or region. If possible, revise your deployment to stay within the quotas. Otherwise, consider requesting a change to your quotas.`
This way I suppose the problem will always connected with quota. 
`ResourceQuotaExceeded` error code means the same and according to this [article](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-quota-errors) we can face it too.

mzueva (approved):

PR #215, opened 2019-04-25
Issue 166 New type of notification High resource consuming

SilinPavel:
## Summary
This PR related to #166. 
Adds new type of notification.

As described in #166 
When memory or disk consuming will be higher that some threshold value for specified period of time (in average) notification will be sent, it will be sent again after `notificationSettings.resendDelay` if the problem is still in place.

f.e. 
```
notificationSettings:
{
    "keepInformedAdmins":true,
    "keepInformedOwner":true,
    "enabled":true,
    "resendDelay": 1800,
    "threshold": -1
}
```
```
"system.memory.consume.threshold" : 95
"system.disk.consume.threshold": 95
"system.monitoring.time.range": 30
```

According to this settings if from the last half hour memory usage rate **or** disk usage rate average value is higher or equals 95 % notification will be sent. Next time it will be sent after another half hour if the problem is in place.

Issue #166, opened 2019-04-18
Notify users if a compute node has high resources pressure


sidoruka:
When a compute-intensive job is run - compute node may start starving for the resources.

CPU high load is typically fine - will result in the SSH/metrics slowdown. But running low on memory and disk - will crash the node, i.e. kubelet will not be able to communicate to the master and node will be lost from the cluster.
In such cases - autoscaler will eventually terminate the cloud instance.

Probably, we cannot do much with the users' jobs resources consumption - to make it clear we shall notify user on the fact of Memory/Disk high load.

1. A new notification type shall be introduced
2. Corresponding settings shall be added to the preferences to control:
* Memory/Disk thresholds, which shall be treated as a risky level (%% from the total amount)
* Threshold timeout (if a real average resource consumption is higher then a threshold for N seconds - a notification shall be sent)

3. The notification template shall support metric name, threshold value, current value templating
4. Add a default email template to the `pipectl` deployment

PR #244, opened 2019-05-06
High resource comsuming notification

SilinPavel:
Update for issue #166

Earlier, #215 was merged, but it uses `pod_container` data from heapster to calculate `memory` and `disk` usage. It turns out that we should use `node` data for this propose.
This PR brings necessary changes.
It also brings convenient way to fetch performance data from heapser as `Map<String, Double>` with `MetricRequester` interface.


sidoruka:
@SilinPavel, could you please fix the naming with of the notification-related json/html? Missed that for #215.
If I remember right - deployment scripts assume that html/json files, applied to the notification, shall have the same name as the notification type.
Currently, they are different and a new deployment have empty config/content of the new notification type: files are named like `HIGH_RESOURCE_CONSUMING_RUN` while a notification is called `HIGH_CONSUMED_RESOURCES`

sidoruka (changes requested):
@SilinPavel, could you please fix the naming with of the notification-related json/html? Missed that for #215.
If I remember right - deployment scripts assume that html/json files, applied to the notification, shall have the same name as the notification type.
Currently, they are different and a new deployment have empty config/content of the new notification type: files are named like `HIGH_RESOURCE_CONSUMING_RUN` while a notification is called `HIGH_CONSUMED_RESOURCES`

SilinPavel:
@sidoruka, indeed, fixed with b40f6dc

mzueva (approved):

mzueva (approved):

Issue #158, opened 2019-04-17
Restrict users from manually installing nvidia tools into the running containers


sidoruka:
# BACKGROUND

It was uncovered in #128 that some of the GPU-enabled runs were not able to initialize due to an issue describe at https://github.com/NVIDIA/nvidia-docker/issues/825 

I.e. it is possible to perform the following operations:
1. Run a GPU-enabled instance
2. SSH into it and install nvidia drivers/toolset/etc. manually
3. Commit such a run into a personal registry
4. Run this new image again

As a result:
1. If running using a GPU-enabled instance (e.g. p/N/etc. instance types):
- run will fail with the following message in the Cloud Pipeline GUI:
```
failed to open log file "/var/log/pods/**.log": open /var/log/pods/**.log: no such file or directory
```
- and the following error in the node logs
```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "process_linux.go:424: container init caused \"process_linux.go:407: running prestart hook 1 caused \\\"error running hook: exit status 1, stdout: , stderr: exec command: [/usr/bin/nvidia-container-cli --load-kmods configure --ldconfig=@/sbin/ldconfig --device=all --compute --utility --require=cuda>=9.0 --pid=10733 
 nvidia-container-cli: mount error: file creation failed: /usr/bin/nvidia-smi: file exists\\\\n\\\"\"": unknown.
ERRO[0000] error waiting for container: context canceled
```

2.  If running using a CPU-only instance: run will succeed as there will be no nvidia runtime present

# TODOs

To limit a possiblity of producing such docker images (which will not be able to start using GPU instance types) - we need to implement a set of restrictions:
- Display a notification to the user (in the Web GUI), that will warn about the risks of installing any of the nvidia packages manually. And that all cuda-based dockers shall be built using [nvidia/cuda](https://hub.docker.com/r/nvidia/cuda/) base images instead
- Restrict users (to any possible extent) from installing those packages while running SSH/terminal session in the container

sidoruka:
@rodichenko please take care of this:
> Display a notification to the user (in the Web GUI), that will warn about the risks of installing any of the nvidia packages manually. And that all cuda-based dockers shall be built using nvidia/cuda base images instead

sidoruka:
[...]

PR #248, opened 2019-05-07
Package manager restrictions

tcibinan:
Resolves issue #158.

The pull request introduces package managers interceptor that checks for a restricted packages in the manager commands.

It creates wrappers for all found package managers when the `launch.sh` is executed. The pull request adds an additional bin path to the `PATH` variable to override the default package manager binaries.

The restricted packages are specified in `CP_RESTRICTED_PACKAGES` variable and package managers that should be restricted are specified in `CP_RESTRICTING_PACKAGE_MANAGERS` variable. Both variables are set in `launch.env.properties`.

For now all **nvidia-** packages are restricted and the following list of package managers are restricted **apt,apt-get,aptitude,yum,pip,pip2,pip3**


workflows/pipe-common/shell/package_manager_restrictor
```diff
@@ -0,0 +1,84 @@
...
+}
+
+if any_restricted_packages "$FORWARDING_ARGUMENTS" "$RESTRICTED_PACKAGES"
+then
+    echo "######################################################################################"
+    echo "##                                                                                  ##"
+    echo "##                    CLOUD PIPELINE PACKAGE MANAGEMENT RESTRICTION                 ##"
+    echo "##                                                                                  ##"
+    echo "##   You tried to install one of the packages that are restricted to be installed   ##"
+    echo "##   within Cloud Pipeline run container.                                           ##"
```
sidoruka:
```
    echo "##   You tried to install one of the packages that are restricted to be installed   ##"
    echo "##   within Cloud Pipeline run container. This may lead to an unusable docker       ##"
    echo "##   image, if committed (i.e. image will not be able to start).                    ##"
```

workflows/pipe-common/shell/package_manager_restrictor
```diff
@@ -0,0 +1,84 @@
...
+    echo "##                                                                                  ##"
+    echo "##   Please see the corresponding GitHub issue to get more information on why the   ##"
+    echo "##   packages are restricted:                                                       ##"
```
sidoruka:
```
    echo "##   Please refer to the following GitHub issue to for the details:   ##"
```

scripts/pipeline-launch/launch.sh
```diff
@@ -1052,6 +1052,36 @@ echo
...
+######################################################
+
+CP_USR_BIN="/usr/cpbin"
+
+mkdir -p "$CP_USR_BIN"
+IFS=',' read -r -a RESTRICTING_PACKAGE_MANAGERS <<< "$CP_RESTRICTING_PACKAGE_MANAGERS"
+
+for MANAGER in "${RESTRICTING_PACKAGE_MANAGERS[@]}"
+do
+    MANAGER_PATH=$(which "$MANAGER")
```
sidoruka: Are we 100% sure that which will be available at this moment?

scripts/pipeline-launch/launch.sh
```diff
@@ -1052,6 +1052,36 @@ echo
...
+    if [[ "$?" == 0 ]]
+    then
+        MANAGER_WRAPPER_PATH="$CP_USR_BIN/$MANAGER"
+        echo "$COMMON_REPO_DIR/shell/package_manager_restrictor \"$MANAGER_PATH\" \"\$@\"" > "$MANAGER_WRAPPER_PATH"
+        chmod +x "$MANAGER_WRAPPER_PATH"
```
sidoruka: Is `+x` enough? I.e. don't we need to set any other permissions for non-root users?

sidoruka (changes requested):

workflows/pipe-common/shell/package_manager_restrictor
tcibinan: Replaced in efa3dd90f3c47976370276fc6861bd4c01a4a10b.

workflows/pipe-common/shell/package_manager_restrictor
tcibinan: Replaced in efa3dd90f3c47976370276fc6861bd4c01a4a10b.

scripts/pipeline-launch/launch.sh
tcibinan: The usage of `which` was replaced with the usage of `command -v` which is a more common approach. Plain `ubuntu:16.04`, `ubuntu:18.04` and `centos:7` images was tested to have `command` available.

Done in 44f08c878f1637b5b845416f720565dbf31add13.

scripts/pipeline-launch/launch.sh
tcibinan: Probably, the most certain way to set permissions for the wrapper scripts is to use permissions of the original package manager binaries.

Done in 44f08c878f1637b5b845416f720565dbf31add13.

sidoruka (approved):

Issue #241, opened 2019-05-06
Implement VM monitor service


mzueva:
## Problem
For various reasons cloud VM instances may "hang" and become invisible to Cloud Pipeline services. E.g. VM was created but some error occurred during joining k8s cluster - in this case Autoscale service will be unable to find such instance and it won't be shut down. This problem leads unmonitored useless  resource consumption and billing.

## Solution
We shall implement VM monitoring service that will track all VM instances in the registered `CloudRegions`, determine whether instances is in "hang" state and notify a configurable set of users about an existing problem. User on receiving such a notification may check the actual state of VM in Cloud Provider console and shut down VM manually.

## Implementation details

- VM monitor shall be implemented as a new standalone service  to be deployed as pod
- VM service shall fetch all registered `CloudRegions` from main API service  and check state of available VMs on a scheduled basis 
- VM shall be considered "hang" in the following cases:
          - VM exists in Cloud provider and it is in `Running` state
          - VM is marked as `worker` tag (configurable)
          - VM is not available as `Node` in Cloud Pipeline OR corresponding node  is missing `RUN_ID` label
- VM service shall notify a set of users about VM "hang" state using main API Service endpoint

sidoruka:
> to be deployed as pod
I'm becoming a fan of the deployments. Let's do it as deployment with a 1 replica

> VM exists in Cloud provider and it is in Running state
Ho do we distinguish the `managed` and `unmanaged` instances?

mzueva:
@sidoruka 
> Ho do we distinguish the managed and unmanaged instances?

I've added comment to the issue description:
> VM is marked as worker tag (configurable)

sidoruka:
Thanks. Please consider adding deployment of this service into `pipectl` as well

mzueva:
Done and deployed

PR #276, opened 2019-05-14
VM Monitor service implementation

mzueva:
This PR provides implementation for #241 

`VMMonitor` is a service for tracking virtual machines in the Cloud provider and verifying that VMs are monitored correctly by Cloud Pipeline API.

### Implementation details 
Service is developed as Spring Boot application with a scheduled task which is invoked according to cron expression. 

The following algorithm is used to check VM and Nodes status:
- Fetch registered `Cloud regions` from Cloud Pipeline API
- Fetch running VMs corresponding to each `Cloud regions`
- For each VM check that corresponding `Node` is registered in Cloud Pipeline k8s cluster (VM and node are matched by private IP address)
- Additionally service checks if VM is associated with any `PipelineRun` via `Name` tag (configurable), if corresponding `PipelineRun` exists and it is in `RUNNING` status, VM is considered healthy. This check is added mostly for cases when VM is created, but not yet registered in k8s cluster during `NodeInitialization` task.
- If matching `Node` is not found `MISSING-NODE` notification is sent
- If matching `Node` exists, service checks that labels required for node management are present (`runid` label by default)
- If some of required labels is absent, `MISSING-LABELS` notification is sent


cloud-pipeline-common/model/src/main/java/com/epam/pipeline/entity/region/AwsRegion.java
```diff
@@ -46,4 +46,8 @@
     private String tempCredentialsRole;
     private Integer backupDuration;
     private boolean versioningEnabled;
+
+    public CloudProvider getProvider() {
```
tcibinan, line 50: It looks like `getProvider` method will be generated by lombok and the overriding can be omitted here.

cloud-pipeline-common/model/src/main/java/com/epam/pipeline/entity/region/AzureRegion.java
```diff
@@ -49,4 +49,8 @@
     private String meterRegionName;
     private String azureApiUrl;
     private String priceOfferId;
+
+    public CloudProvider getProvider() {
```
tcibinan, line 53: Same here.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/VMMonitorApplication.java
```diff
@@ -0,0 +1,112 @@
...
+        scheduler.setRemoveOnCancelPolicy(true);
+        scheduler.setPoolSize(SCHEDULED_TASKS_POOL_SIZE);
+        return scheduler;
+    }
+
+    @Bean
+    public NotificationSender notificationSender(
+            @Value("${notification.send-via-api:false}") final boolean useApi,
+            final CloudPipelineAPIClient apiClient) {
+        return useApi ? new ApiNotificationSender(apiClient) : new SMTPNotificationSender(apiClient, buildSmtpConfig());
```
tcibinan: Probably, this is a use-case for conditional beans. Both `ApiNotificationSender` and `SMTPNotificationSender` can be made regular beans with explicit `ConditionalOnProperty` annotation specified.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/VMMonitor.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @SuppressWarnings("unchecked")
+    private void checkVMs(final AbstractCloudRegion region) {
+        log.debug("Checking VMs in region {} {}", region.getRegionCode(), region.getProvider());
+        getVmService(region)
+                .ifPresent(service -> {
+                    final List<VirtualMachine> vms = service.fetchRunningVms(region);
+                    log.debug("Found {} running VM(s) in {} {}", vms.size(),
+                            region.getRegionCode(), region.getProvider());
+                    ListUtils.emptyIfNull(vms).forEach(this::checkVmState);
```
tcibinan: No need to check if `vms` is empty, because if it is then it will end up in `NullPointerException` on 87 line :)

Probably, `ListUtils.emptyIfNull` should be used to wrap `service.fetchRunningVms` call.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/impl/AWSMonitorService.java
```diff
@@ -0,0 +1,131 @@
...
+            clientBuilder.withCredentials(new ProfileCredentialsProvider(region.getProfile()));
+        }
+        return clientBuilder
+                .build();
+    }
+
+    private void fetchInstances(final AmazonEC2 ec2,
+                                final DescribeInstancesRequest request,
+                                final List<VirtualMachine> vms) {
+        final DescribeInstancesResult result = ec2.describeInstances(request);
```
tcibinan, line 83: Rather then mutating the argument `vms` list concatenated stream of vms can be returned.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/impl/AzureMonitorService.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public CloudProvider provider() {
+        return CloudProvider.AZURE;
+    }
+
+    private VirtualMachine toVM(final com.microsoft.azure.management.compute.VirtualMachine azureVM) {
+        return VirtualMachine.builder()
+                .instanceId(azureVM.computerName())
```
tcibinan: We use `azureVM.name()` in the Cloud Pipeline Api as instance id. Probably, the difference can lead to unexpected errors in the future.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/impl/AzureMonitorService.java
```diff
@@ -0,0 +1,92 @@
...
+    public AzureMonitorService(@Value("${monitor.instance.tag}") final String instanceTag) {
+        this.instanceTag = VMTag.fromProperty(instanceTag);
+    }
+
+    @Override
+    public List<VirtualMachine> fetchRunningVms(final AzureRegion region) {
+        final Azure azure = buildClient(region.getAuthFile());
+        final PagedList<GenericResource> resources = azure.genericResources()
+                .listByTag(region.getResourceGroup(), instanceTag.getKey(), instanceTag.getValue());
+        final Iterable<GenericResource> iterable = resources::listIterator;
```
tcibinan, line 56: Looks like only a single page of resources will be loaded. Please see `com.epam.pipeline.manager.cloud.azure.AzureVMService` for an example of a recursive loading.

tcibinan (changes requested):

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/impl/AzureMonitorService.java
mzueva: I've checked the docs for `PagedList` class: 
` Defines a list response from a paging operation. The pages are
 lazy initialised when an instance of this class is iterated.
`
So I assume that this code will handle multipage responses correctly.

cloud-pipeline-common/model/src/main/java/com/epam/pipeline/entity/region/AwsRegion.java
mzueva: There are some issues related to Jackson deserialisation of class hierarchy, so I'll leave this implementation as is for now.

Issue #277, opened 2019-05-14
E2E GUI tests problems from 14.05.19


kamyshova:
There are several problems with E2E GUI tests revealed on 14 of May:

**AZURE** stand:

- `DockerCommitTest` fails because pipeline pulls incorrect docker image (the bug is described in #269)
- `PipelineDockerCommitTest` presumably fails because the commit operation has been aborted:

`ExceptionHandlerAdvice - 
This operation has been aborted: uri=/pipeline/restapi/run/1216/commit;client="";session="";user="".  java.lang.NullPointerException: null 
at com.epam.pipeline.manager.cluster.KubernetesManager.getContainerIdFromKubernetesPod(KubernetesManager.java:126) ~[classes!/:0.15.0.874.0161935f08f5d0c067da30f7407efadc9a943b1b] at com.epam.pipeline.manager.docker.DockerContainerOperationManager.commitContainer(DockerContainerOperationManager.java:125)`

- `RoleModelTest` fails because the test click() on runs menu was performed but switch-over
to the runs page didn't happen
- `ToolsScanTest` fails because the the test click() on "OK" button was performed after saving preference in settings popup but it didn't work

NShaforostov:
`PipelineDockerCommitTest` very likely failed because of the bug described at #262. "InitializationEnvironment", "MountDataStorages", "InputData" tasks weren't be executed before commit trying.
Separate issues on other actual bugs/errors, mentioned above, were created.
This issue will be closed.

PR #278, opened 2019-05-14
Resolve E2E GUI test issues from 14.05.19

kamyshova:
Resolves problems with tests in #277 
Navigation menu buttons click checks was added

### DIFF
--- a/e2e/gui/src/test/java/com/epam/pipeline/autotests/ao/NavigationMenuAO.java
+++ b/e2e/gui/src/test/java/com/epam/pipeline/autotests/ao/NavigationMenuAO.java
@@ -15,44 +15,56 @@
  */
 package com.epam.pipeline.autotests.ao;
 
+import org.openqa.selenium.By;
+
 import static com.codeborne.selenide.Condition.exist;
 import static com.codeborne.selenide.Condition.not;
 import static com.codeborne.selenide.Condition.visible;
 import static com.codeborne.selenide.Selectors.*;
 import static com.codeborne.selenide.Selenide.$;
+import static com.epam.pipeline.autotests.utils.Conditions.selectedMenuItem;
+import static com.epam.pipeline.autotests.utils.Utils.click;
 import static com.epam.pipeline.autotests.utils.Utils.sleep;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class NavigationMenuAO {
 
     public PipelinesLibraryAO library() {
-        $(byId("navigation-button-pipelines")).shouldBe(visible).click();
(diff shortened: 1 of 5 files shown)


e2e/gui/src/test/java/com/epam/pipeline/autotests/utils/Utils.java
```diff
@@ -360,4 +361,8 @@ public static void restartBrowser(final String address) {
         Selenide.close();
         Selenide.open(address);
     }
+
+    public static void click(final By selector) {
+        $(selector).shouldBe(visible).click();
```
tcibinan, line 366: You can implement `AccessObject` interface if you need to use safe click. Probably, there is no need for `Utils.click`.

e2e/gui/src/test/java/com/epam/pipeline/autotests/ao/NavigationMenuAO.java
```diff
@@ -15,44 +15,64 @@
...
         $(byXpath("//*[.//*[text()[contains(.,'Library')]] and contains(@id, 'pipelines-library-content')]"))
                 .waitUntil(visible, 5000);
         return new PipelinesLibraryAO();
     }
 
     public RunsMenuAO runs() {
-        $(byId("navigation-button-runs")).shouldBe(visible).click();
+        final By runsPageSelector = byId("navigation-button-runs");
+        click(runsPageSelector);
+        if (!$(runsPageSelector).has(selectedMenuItem)) {
```
tcibinan: I suppose we should not use `has` because it returns immediately. According to the original issue we should wait until the menu element is active rather then trying to click second time. From my point of view usage of simple `should` can be of help here.

tcibinan: It is related to all similar use cases in the pull request.

tcibinan (changes requested):

e2e/gui/src/test/java/com/epam/pipeline/autotests/utils/Utils.java
kamyshova: @tcibinan It's impossible implement `AccessObject` for `Navigation` interface.

e2e/gui/src/test/java/com/epam/pipeline/autotests/ao/NavigationMenuAO.java
kamyshova: @tcibinan Thanks, fixed.

tcibinan (approved):

Issue #303, opened 2019-05-17
Support Azure low-priority VMs using scale sets


mzueva:
# BACKGROUND
Cloud Pipeline shall support low-priority VMs for Azure deployments. When user launches a run with `spot` parameter set to `true`, we shall run low-priority VM.

# Approach
Azure currently provides two options to create a low-priority VM: [Azure Batch](https://docs.microsoft.com/en-us/azure/batch/) и [VM Scale sets](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/overview). Scale sets looks like a more effective approach to support low-priority VMs in Cloud Pipeline.

# Implementation

- Autoscaling `nodeup` script shall respect `spot` argument. If `spot` is set to `true` we shall create a Scale Set size limited to `1`. Script shall wait until Scale set is initialised, corresponding instances is created and started. Any additional instances created during Scale Set initialisation shall be cleaned up from Azure and k8s.
- `terminate_node`, `node_reassign` and `nodedown` scripts shall support handling of Scale Sets in the same manner as handling on-demand VMs
- `AzureInstanceService` class shall support operations on `Scale Sets` in the same manner as handling on-demand VMs (excluding `startInstance` and `stopInstance`, as these methods can be used only for on-demand instances).

mzueva:
Implemented and deployed

PR #311, opened 2019-05-21
Issue 303 azure low priority instances

SilinPavel:
Implements issue #303 

This PR provides abilities to use low priority instances from azure cloud as `spot` instances in our cloud pipeline.

Implementation based on creation of Scale Set in Azure cloud with 1 low priority instance.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cloud/azure/AzureInstanceService.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cloud/azure/AzureInstanceService.java
@@ -243,8 +243,11 @@ public class AzureInstanceService implements CloudInstanceService<AzureRegion> {
                 .kubeToken(kubeToken)
                 .region(region.getRegionCode());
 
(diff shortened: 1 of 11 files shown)


api/src/main/java/com/epam/pipeline/manager/cloud/azure/AzureInstanceService.java
```diff
@@ -240,6 +243,10 @@ private String buildNodeUpCommand(final AzureRegion region, final Long runId, fi
                 .kubeToken(kubeToken)
                 .region(region.getRegionCode());
 
+        final Boolean clusterSpotStrategy = preferenceManager.getPreference(SystemPreferences.CLUSTER_SPOT);
```
mzueva: As far as I remember, we shall use preference value only if `instance.getSpot() == null`

SilinPavel: correct

Issue #321, opened 2019-05-22
Azure Low-priority limit instance types


mzueva:
# Background
Azure supports limited number of instance families/types for low-priority VMs. API shall not allow to launch unsupported instance types as `spot` (low-priority) run.

# Implementation
- All API requests regarding price types shall support additional parameter `spot` and filter available instance types with respect to this parameter.
- Corresponding changes shall be applied to the GUI

PR #347, opened 2019-05-27
Issue 321 filter instance type

SilinPavel:
Related to #321 

Since Azure cloud doesn't allow run any instance type as Low Priority instance we need to provide possibilities to filter it by `spot` type. 
This PR adds ability to filter types and prices of cloud instance by `spot` flag. 

**NOTE**: This PR depends on #311


mzueva (changes requested):
Missing spot argument of `InstanceOfferManager.isInstanceAllowed` & `isToolInstanceAllowed`. Please check that all methods for getting instance types accept `spot argument`

api/src/main/java/com/epam/pipeline/manager/cloud/CloudFacadeImpl.java
```diff
@@ -173,6 +174,20 @@ public boolean instanceExists(final Long regionId, final String instanceId) {
         return getInstanceService(region).getInstanceTerminationState(region, instanceId);
     }
 
+    @Override
+    public List<InstanceType> getAllInstanceTypes(final Long regionId, final Boolean spot) {
+        if (regionId == null) {
+            return (List<InstanceType>) instancePriceServices.values()
+                    .stream()
+                    .map(priceService -> priceService.getAllInstanceTypes(regionId, spot))
```
mzueva: I'd suggest to split this into two methods, since they implement different logic.

api/src/main/java/com/epam/pipeline/manager/cloud/CloudInstancePriceService.java
```diff
@@ -36,4 +37,17 @@
     List<InstanceOffer> refreshPriceListForRegion(T region);
     double getSpotPrice(String instanceType, T region);
     double getPriceForDisk(List<InstanceOffer> offers, int instanceDisk, String instanceType, T region);
+    List<InstanceType> getAllInstanceTypes(Long id, Boolean spot);
```
mzueva: How does this method selects instance types by provider, if `regionId` is not specified?
Please, call argument `regionId`
If `spot` argument is required, make it `boolean` to prevent NPE

api/src/main/java/com/epam/pipeline/manager/cloud/azure/AzureInstancePriceService.java
```diff
@@ -87,7 +88,7 @@ public double getSpotPrice(final String instanceType, final AzureRegion region)
     public double getPriceForDisk(final List<InstanceOffer> diskOffers, final int instanceDisk,
                                  final String instanceType, final AzureRegion region) {
         final InstanceOfferRequestVO requestVO = new InstanceOfferRequestVO();
-        requestVO.setTermType(ON_DEMAND_TERM_TYPE);
+        requestVO.setTermType(PriceType.ON_DEMAND.getName());
```
mzueva: Does Azure have the same prices for disks attached to  `on_demand` and `low_priority` VMs?

api/src/main/java/com/epam/pipeline/controller/cluster/ClusterController.java
```diff
@@ -152,8 +153,9 @@
             })
     public Result<AllowedInstanceAndPriceTypes> loadAllowedInstanceAndPriceTypes(
             @RequestParam(required = false) final Long toolId,
-            @RequestParam(required = false) final Long regionId) {
-        return Result.success(clusterApiService.getAllowedInstanceAndPriceTypes(toolId, regionId));
+            @RequestParam(required = false) final Long regionId,
+            @RequestParam(required = false, defaultValue = "false") final boolean spot) {
```
mzueva: I'd suggest to take default value from SystemPreferences `cluster.spot`

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/cloud/azure/AzureInstancePriceService.java
SilinPavel: Can't find any information about difference in this case. Assume that, yes.

api/src/main/resources/db/migration/v2019.06.05_18.00__cloud_provider_in_instance_offer.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM pipeline.instance_offer;
```
mzueva: I think we can get `cloud_provider` from join with `region` table.

mzueva (changes requested):

mzueva (approved):

PR #358, opened 2019-05-29
Update parallel environment slots while GE autoscaling

### LINKED ISSUE #357: GE Autoscaler doesn't update parallel environment slots
Currently, GE autoscaler increases and decreases number of slots of the `main.q` according to the current cluster capacity. It allows to schedule jobs without specified number of slots up to the full `main.q` capacity. 

At the same time, if job is scheduled with specified number of slots then parallel environment capacity is used as well. And as long as GE autoscaler doesn't update default parallel environment capacity then running jobs total slots on the entire cluster cannot be higher then the original parallel environment capacity. For a single master run it is number of master cores and for a autoscaled run it is number of master cores multiplied by the number of default workers.

To enable proper multislots jobs scheduling GE autoscaler shall update not only the `main.q` capacity but default parallel environment capacity as well. Default parallel environment is `local`.

tcibinan:
Resolves #357.

The pull requests brings support for multislots jobs in GE autoscaling. See more information in the original issue.

### DIFF
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -20,6 +20,7 @@ from datetime import datetime, timedelta
 from pipeline import PipelineAPI, Logger as CloudPipelineLogger
 import subprocess
 import time
+import multiprocessing
 
 
 class ExecutionError(RuntimeError):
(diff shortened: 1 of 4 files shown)


workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -889,6 +936,7 @@ def _retrieve_preference(pipe, preference, default_value):
     instance_type = os.environ['instance_size']
     instance_image = os.environ['docker_image']
     price_type = os.environ['price_type']
+    instance_cores = int(os.environ['CLOUD_PIPELINE_NODE_CORES'])
```
sidoruka: If env not set - use cores count from OS

sidoruka (changes requested):

workflows/pipe-common/scripts/autoscale_sge.py
tcibinan: Done in d64e1e5.

sidoruka (approved):

Issue #223, opened 2019-04-29
Cloud Pipeline shall allow to expose compute node's FS to upload/download files


sidoruka:
**BACKGROUND**

For the `interactive runs` (e.g. Web GUI/Desktop/SSH) users are processing data in `ad-hoc` manner. Which requires data upload (from the local storage to the cloud) and results download (from the cloud to the local storage).

At the moment `Cloud Pipeline` supports a number of options for the user to perform that data transfers for the `interactive runs`:
- via the Object Storage (using Web GUI or CLI)
- via the File Storage (using Web GUI/CLI/WebDav-mounted-drive)

But those options require an intermediate location (bucket/fs) to be used. This often confuses users, when a small dataset shall be loaded to the specific location within a run's filesystem.

We can simplify this procedure using a direct exposure of the run's filesystem to the users. E.g.:
- User runs a job and it is initialized
-  `Cloud Pipeline` GUI shows a hyperlink `Browse filesystem` (next to the existing `SSH` button in the log form and in the `Active Runs` dashboard)
- Clicking a link - loads a `Storage browser` Web GUI (same as used for the Object and FS storages)
- User is able to view/download/upload/delete files/directories

**Approach**

To implement the direct FS exposure, the following approach is proposed:
* **(FS API)** A separate (python/flask) service shall be implemented. Service shall expose RESTful API, that shall allow to view/download/upload/delete files/directories. API shall be compatible (to a reasonable extent) with the current Object/FS storages API
  * Download operation shall support both files and directories
  * If a directory is requested for download - it shall result in a gzipped tarball
  * Data transfer operations may be quite heavy (e.g. hundreds of GB). As the traffic will flow through the EDGE service, we may experience performance degradation. To workaround this issue - intermediate object storage shall be used for that:
    * A separate object storage shall be used (identified by `CP_CAP_EXPOSE_FS_STORAGE` environment variable)
    * Storage shall have an auto-cleanup policy (to remove the temp files)
    * If user requests `upload` operation: API shall provide URL for the upload (pointing to the object storage), client will use that URL to upload the data and notify API that the upload is complete. As a result - API will move the file(s) from the object storage to the compute node
    * If user requests `download` operation: API shall copy the requested file to the object storage and provide URL to the client. Client shall download that file from the object storage
[...]

PR #359, opened 2019-05-29
Cloud Pipeline shall allow to expose compute node's FS to upload/download files

ekazachkova:
Implements issue #223 

The current PR provides `python/flask` service to expose compute node's file system.

## RESTful API methods
This service provides RESTful API with the following methods:
-  `/view/<path>` - lists files via specified path. `path` - the path to file/folder relative to `working_directory`. The `working_directory` is the path to directory that will be shared for user. This is the required parameter that should be specified when application is started. 
- `/download/<path>` - transfers file/folder via specified path on common object storage. If the user initiate to download the folder the `.tar.gz`  file will be provided for user. This action is long running so launches asynchronously.  The result of this request is the `task_id` that should be used for task status monitoring. When the download operation is completed a new folder with unique name `task_id` will be created on common object storage. The result of the whole download operation is the sign url to the file in `task_id` bucket's folder (e.g. sing url will be generated for object that located with s3://CP_CAP_EXPOSE_FS_STORAGE/3fcecb8986a34c54939dbaf8c4a2238b/data.tar.gz).
- `/status/<task_id>` - returns specified task status. Allowed states:  'pending', 'success', 'running', 'failure' or 'canceled'. If the task is finished with status 'failure' an error message will be returned. If the task is completed successfully the task result will be returned.
- `/cancel/<task_id>` - cancels task computation. Cleanups data if needed. 
- `/uploadUrl/<path>` - returns presigned url to upload file to storage by path cp://CP_CAP_EXPOSE_FS_STORAGE/<task_id>/<path> where `task_id` - random generated string, `path` - file location on compute node.
- `/upload/<task_id>` - transfers uploaded file from bucket to compute node. If the current task is still in progress but was canceled the created file will be removed.   
- `/delete/<path>` - removes file/folder from compute node.

## Launching
To run app use the following command:
```
python <path to project>/app.py --working_directory=/root/ --host=0.0.0.0 --port=8080 --log_dir=/root/logs
```
## Requirements

- Python 2.7
- pipe cli
- flask


sidoruka (changes requested):
1. Move the `fs-browser` directory to `workflows/pipe-common/`
2. Add `setup.py` with the dependencies/versions
3. If I run app.py with `--working_directory=/` - I expect that `curl http://127.0.0.1:8080/view/` will return the listing of the filesystem root. But I get `The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.`
4. Do not use `CP_CAP_EXPOSE_FS_STORAGE` environment variable in this module. Instead, expose an additional (mandatory) command line parameter to the `app.py`. It shall be an external responsibility to set the transfer storage
5. Allow to set suffix for the transfer storage location. At the moment all the files are placed in to the `<bucket root>/<task id>`. But in certain cases I'd like to isolate the `task id` folders into the some other directory. E.g.: `<bucket root>/<suffix>/<task id>`. Keeping in mind **(4)**, one shall be able to specify `--transfer-storage mybucket/fs_transfers`
6. `/view` shall return the size of files (ONLY files, NOT directories)

ekazachkova:
@sidoruka all requested changes was done. Could you review them?

rodichenko:
@sidoruka @mzueva as for GUI, source code is located at `/workflows/pipe-common/fs-browser/fs-browser-client`.

To build an application:
- Ensure **nodejs** and **npm** are installed
- Run the following command:
```
npm install
npm run build
```
All build assets will be located at `build` folder.

By default, **FS Browser Client** uses API methods exposed at the same url, e.g.:
```
API: https://server.com/fsbrowser/
Client: https://server.com/fsbrowser/#/
```
You can override API endpoint by specifying **FS_BROWSER_API** env var at the build stage (both relative and absolute).

sidoruka (approved):

Issue #363, opened 2019-05-29
Can't delete user if the notification queue has referenced entries


sidoruka:
If a notification queue references a user - such a user can't be deleted, as it violates the FK.

1. We shall cleanup the notification queue once a user is being deleted
2. We shall double check, whether there any similar case, that may prevent the correct user deletion

mzueva:
Fixed

PR #365, opened 2019-05-30
Issue 363 delete all notifications for user while delete the user

SilinPavel:
This PR provides fix for issue #363.
Now all notification for specific user will be deleted if request for user deleting is sent.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/user/UserAspect.java
+++ b/api/src/main/java/com/epam/pipeline/manager/user/UserAspect.java
@@ -1,0 +1,48 @@
+/*
+ *
+ *  * Copyright 2017-2019 EPAM Systems, Inc. (https://www.epam.com/)
+ *  *
+ *  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/epam/pipeline/manager/user/UserAspect.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Pointcut("execution(* com.epam.pipeline.manager.user.UserManager.deleteUser(..))")
+    public void onDeleteUser() {}
+
+    /**
+     * Deletes all notification to specific user. This Aspect is used when user is being deleted
+     * @param joinPoint
+     * @param userId
+     */
+    @Around(value = "onDeleteUser() && args(userId)")
```
mzueva: Maybe we can use one `AfterReturning` annotation instead of two methods? Or you need such two method approach to get `userId`?

SilinPavel: yes, indeed here we can use just `@Around` with direct pointcat expression in it

Issue #336, opened 2019-05-24
API shall check the existence of the kube's docker pull secrets at a startup time


sidoruka:
1. When API is launched - it shall check if all the kube secrets exist for all registered docker registries
2. If not - they shall be recreated (if registry configuration is available for the API)

mzueva:
Verified

PR #367, opened 2019-05-30
API shall check the existence of the kube's docker pull secrets at a startup time

ekazachkova:
Implementation for issue #336

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/docker/DockerRegistryManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/docker/DockerRegistryManager.java
@@ -410,6 +410,7 @@ public class DockerRegistryManager implements SecuredEntityManager {
         return new DockerRegistryList(registries);
     }
 
+    @Transactional(propagation = Propagation.REQUIRED)
     public void checkDockerSecrets() {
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/docker/DockerRegistryManager.java
```diff
@@ -409,6 +410,31 @@ public DockerRegistryList loadAllRegistriesContent() {
...
+                .forEach(registry -> {
+                    final String secretName = kubernetesManager.getValidSecretName(registry.getPath());
+                    if (!secrets.contains(secretName)) {
+                        final DockerRegistrySecret secret = DockerRegistrySecret
+                                .builder()
+                                .userName(registry.getUserName())
+                                .password(registry.getPassword())
+                                .registryUrl(registry.getPath())
+                                .build();
+                        registry.setSecretName(kubernetesManager.createDockerRegistrySecret(secret));
```
mzueva, line 434: We also shall persist this to DB

ekazachkova: done

mzueva (approved):

Issue #291, opened 2019-05-15
Support Nextflow pipelines


sidoruka:
**BACKGROUND**

At the moment `Cloud Pipeline` support running `WDL` pipeline using `Cromwell` executor:
- `WDL` scripts can be visualized and modified using the [Pipeline Builder](https://epam.github.io/cloud-pipeline/manual/06_Manage_Pipeline/6.1.1_Building_WDL_pipeline_with_graphical_PipelineBuilder/#example-pipeline)
- `WDL` pipelines can be run using `local` (all tasks run in the same container), `docker` (tasks can be run using different docker images) and `Cloud Pipeline` (tasks can be run across different nodes/docker images) backends

**[Nextflow](https://github.com/nextflow-io/nextflow)** is another pipelines building/execution engine, which looks quite similar to `WDL/Cromwell` in terms of concepts.

It would be great to introduce a support for running `Nextflow` pipelines within `Cloud Pipeline` using the same capabilities as for `WDL`

**Action items**

* Research possible integration approaches:
  * Using one of the existing `Nextflow` backends for the integration (like we are doing with `Cromwell`'s local backend). This shall be considered as a priority approach and shall not require much of the development. E.g. [Nextflow SGE executor](https://www.nextflow.io/docs/latest/executor.html#sge) looks promising and may be a good pair for the [Grid Engine autoscaling](https://epam.github.io/cloud-pipeline/release_notes/v.0.14/v.0.14_-_Release_notes/#auto-scaled-cluster). Another option is a [Nextflow Kubernetes executor](https://www.nextflow.io/docs/latest/executor.html#kubernetes), which may be a way to go as well.
  * Implementing own executor backend, that will integrate with `Cloud Pipeline`'s API. This approach shall be considered only if there is no way to implement the integration using existing techiques.
* We shall also research whether it is possible/feasible to introduce [Netflow scripting language](https://www.nextflow.io/docs/latest/script.html) support to the [Pipeline Builder](https://github.com/epam/pipeline-builder) and thus to the `Cloud Pipeline`
* Setup a pipeline template for the `nextflow` execution environment with a dummy script

TimSPb89:
> We shall also research whether it is possible/feasible to introduce Netflow scripting language support to the `Pipeline Builder` and thus to the `Cloud Pipeline`

[...]

PR #411, opened 2019-06-19
Initialize RAM consumable resource in GE

tcibinan:
Relates to #291.

The pull request brings support for RAM consumable resource in grid engine. In some rare cases RAM should be utilized as a consumable resource while scheduling tasks in grid engine. Usage of RAM as a consumable resource reduces the probability of out of memory errors but at the same time decreases an overall grid engine cluster performance.

RAM consumable resource name can be configured using `CP_CAP_GE_CONSUMABLE_RESOURCE_NAME_RAM` parameter. Default name: `ram`. Notice that all nodes of the grid engine cluster should use the same name for RAM consumable resource.

Both CPU/GPU configurations were tested with base ubuntu 16/18 and centos 7 images.

### DIFF
--- a/workflows/pipe-common/shell/sge_setup_resources
+++ b/workflows/pipe-common/shell/sge_setup_resources
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Copyright 2017-2019 EPAM Systems, Inc. (https://www.epam.com/)
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


mzueva (approved):

sidoruka:
@tcibinan did you test the GPU resources configuration after the RAM changes made?
If not - please test that the following configurations work:
1. GPU: Ubuntu 16/18, Centos 7
2. RAM: Ubuntu 16/18, Centos 7

If it was tested - please add this information to the PR description

tcibinan:
@sidoruka Information on the testing were added to the pull request description

sidoruka (approved):

Issue #436, opened 2019-06-24
pipe command line shall warn user on the JWT expiration


sidoruka:
`pipe` CLI shall provide information on the JWT token expiration:
1. When `configure` command is run - pipe shall print the expiration date of the provided token If it is less than 7 days left
2. When `--version` options are specified - pipe shall print the expiration date of the currently used token
3. When running any other command - pipe shall check the expiration of the JWT token. If it is less than 7 days left - warn user that a JWT token is going to expire in N days.

sidoruka:
Backport to release/v0.15

PR #453, opened 2019-06-25
Pipe command line shall warn user on the JWT expiration

rodichenko:
Pipe command line shall warn user on the JWT expiration - done (#436)

### DIFF
--- a/pipe-cli/src/config.py
+++ b/pipe-cli/src/config.py
@@ -41,7 +41,7 @@ def check_token(token, timezone, print_info=False):
         click.echo(click.style('No access token is provided', fg='red'), err=True)
         return
     try:
-        (header_decoded, payload_decoded, signature) = token.split('.')
+        payload_decoded = token.split('.')[1]
         # payload_decoded_correct_padding - is the same string with additional '=' symbols to make string length
         # be dividable by 4:
         payload_decoded_correct_padding = payload_decoded + b'=' * (4 - len(payload_decoded) % 4)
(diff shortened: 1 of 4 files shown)


sidoruka (commented):
Does this functionality also considers `$API_TOKEN` value if `pipe` was configured via the environment variables?

rodichenko:
> Does this functionality also considers `$API_TOKEN` value if `pipe` was configured via the environment variables?

Yes, jwt-token validation is implemented separately from the configuration itself; `pipe-cli` first reads `$API` & `$API_TOKEN` from the environment variables, then processes `configuration` file, and after that the validation occurs.

sidoruka (approved):

pipe-cli/src/config.py
```diff
@@ -12,33 +12,133 @@
...
+def check_token(token, timezone, print_info=False):
+    if not token:
+        click.echo(click.style('No access token is provided', fg='red'), err=True)
+        return
+    try:
+        payload_decoded = token.split('.')[1]
+        # payload_decoded_correct_padding - is the same string with additional '=' symbols to make string length
+        # be dividable by 4:
+        payload_decoded_correct_padding = payload_decoded + b'=' * (4 - len(payload_decoded) % 4)
+        payload = json.loads(base64.b64decode(payload_decoded_correct_padding))
```
mzueva: `jwt` library is already included into the project (take a look `src/utilitities/storage/common.py` -> `StorageOperations.get_user()`

mzueva (changes requested):

pipe-cli/src/config.py
```diff
@@ -12,33 +12,133 @@
...
+ALL_ERRORS = Exception
+
 
 class ConfigNotFoundError(Exception):
     def __init__(self):
         super(ConfigNotFoundError, self).__init__('Unable to locate configuration or it is incomplete. '
                                                   'You can configure pipe by running "pipe configure"')
 
 
+def check_token(token, timezone, print_info=False):
```
ekazachkova: I would prefer not to make this file more complicated and move some methods into utilities as far as it possible

pipe-cli/src/config.py
rodichenko: @mzueva fixed via e74e38ae60622e3f862f2084f8e6a49145263748

pipe-cli/src/config.py
rodichenko: @ekazachkova I agree, I moved the validation logic to the `access_token_validation.py` utility (e74e38ae60622e3f862f2084f8e6a49145263748)

mzueva (approved):

ekazachkova (approved):

PR #454, opened 2019-06-25
Use google cloud operation API to get termination state of an instance

SilinPavel:
This PR brings new usage of google cloud operation API to get termination reason for GCP instances.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPVMService.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPVMService.java
@@ -22,6 +22,8 @@ import com.epam.pipeline.entity.cloud.InstanceTerminationState;
 import com.epam.pipeline.entity.region.GCPRegion;
 import com.epam.pipeline.exception.cloud.gcp.GCPException;
 import com.google.api.services.compute.model.Instance;
+import com.google.api.services.compute.model.Operation;
+import com.google.api.services.compute.model.OperationList;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.ListUtils;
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPVMService.java
```diff
@@ -82,22 +87,40 @@ public Instance getRunningInstanceByRunId(final GCPRegion region, final String r
         }
     }
 
-    public Optional<InstanceTerminationState> getTerminationState(final GCPRegion region,
-                                                                  final String instanceId) {
+    public Optional<InstanceTerminationState> getTerminationState(final GCPRegion region, final String instanceId) {
+        Optional<InstanceTerminationState> result = Optional.empty();
```
mzueva: I'd suggest not to introduce a mutable variable. And just return result instead of assigning a new value.

api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPVMService.java
```diff
@@ -82,22 +87,40 @@ public Instance getRunningInstanceByRunId(final GCPRegion region, final String r
...
-                                .instanceId(instanceId)
-                                .stateCode(GCPInstanceStatus.TERMINATED.name())
-                                .stateMessage(instance.getStatusMessage()).build()
-                );
+            final OperationList operationList = gcpClient.buildComputeClient(region)
+                    .zoneOperations()
+                    .list(region.getProject(), region.getRegionCode())
+                    .setFilter(String.format(COMPUTE_OPERATIONS_FILTER, instanceId)).execute();
+
+            if (!operationList.isEmpty()) {
```
mzueva: To reduce code indentation: `ListUtils.emptyIfNull(operationList.getItems()).fillter(...).findFirst(...).map(...)`

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/cloud/gcp/GCPVMService.java
SilinPavel: Agree, looks much better :)

mzueva (approved):

PR #462, opened 2019-06-27
Issue 459 Fix NPE in ResourceMonitorManager

SilinPavel:
There was a case when `ResourceMonitorManager` can fail with NPE. 
This PR fixes this problem by setting field `prolongedAtTime` to current time for restarted runs (as it currently perform for newly created runs). In `ResourceMonitorManager` class we also assume that runs without  `prolongedAtTime` field isn't prolonged at all.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
@@ -199,8 +199,11 @@ public class ResourceMonitoringManager extends AbstractSchedulingManager {
         final int idleTimeout = preferenceManager.getPreference(SystemPreferences.SYSTEM_MAX_IDLE_TIMEOUT_MINUTES);
 
         final Map<String, PipelineRun> notProlongedRuns = running.entrySet().stream()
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -199,8 +199,11 @@ private void processIdleRuns(final List<PipelineRun> runs) {
         final int idleTimeout = preferenceManager.getPreference(SystemPreferences.SYSTEM_MAX_IDLE_TIMEOUT_MINUTES);
 
         final Map<String, PipelineRun> notProlongedRuns = running.entrySet().stream()
-                .filter(e -> DateUtils.nowUTC().isAfter(e.getValue().getProlongedAtTime()
-                        .plus(idleTimeout, ChronoUnit.MINUTES)))
+                .filter(e -> {
+                    LocalDateTime prolongedAtTime = e.getValue().getProlongedAtTime();
```
mzueva: I think we can use `Optional` here

mzueva (changes requested):

mzueva (approved):

Issue #468, opened 2019-07-01
Getting logs from k8s may cause OutOfMemory error


mzueva:
# Background
For some workloads, container logs may become very large: up to several gigabytes. When we try to fetch such logs it is likely to cause OutOfMemory error, since k8s library will try to load it into a single String object. We need to limit log's size to prevent such problems.
# Action items

- Introduce a new `SystemPreference` with pod log size limit in lines
- Method for getting pod logs shall fetch tail log lines with respect to this limit
- If actual pod logs exceeds limit add a prefix to log text to show that log is truncated

sidoruka:
Backport to release/v0.15

SilinPavel:
should be resolved by  88c025d

mzueva:
@sidoruka cherry-picked to release/0.15

PR #480, opened 2019-07-02
Issue 468 load pod logs partially

SilinPavel:
This PR brings ability to control loading of pod logs by specifying a size of logs to be loaded. 
New system parameter `system.logs.line.limit` is introduced, this parameter represents amount of logs that will be loaded,
Now, if logs don't fill in specified size only tail of this logs will be loaded


mzueva:
@SilinPavel I've updated issue #468, please review the changes.
Also do not forget to add a reference to the initial issue to PR and commits as it makes issue tracking easier.

api/src/main/java/com/epam/pipeline/manager/cluster/KubernetesManager.java
```diff
@@ -75,6 +75,8 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(KubernetesManager.class);
     private static final String DEFAULT_SVC_SCHEME = "http";
     private static final int NODE_PULL_TIMEOUT = 200;
+    private static final String NEW_LINE = "\n";
+    private static final String TRUNCATED_SIGN = "< ... Log was truncated, it is too big ...>";
```
mzueva: Please, change the message text to `< ... Some of the preceding lines were skipped due to log size limit ... >`.
And it is better to move it to `MessageConstants`

mzueva (approved):

Issue #418, opened 2019-06-20
Incorrect displaying of the "Start idle" checkbox


NShaforostov:
1. Create detach configuration object
2. Specify mandatory environment parameters, click the "Save" button
3. Add the second configuration
4. For the second configuration specify any Cmd template, click the "Save" button
5. Return to the first configuration tab, set "Start idle" checkbox
_("cmd template" field disappears)_
6. Return to the second configuration tab
**_Exp._**: "Start idle" checkbox is unset, "cmd template" field with the saved command is displaying
**_Act._**: "Start idle" checkbox is set, "cmd template" field with the saved command is displaying
7. Return to the first configuration tab
**_Exp._**: "Start idle" checkbox is set, "cmd template" field doesn't display
**_Act._**: "Start idle" checkbox is set, "cmd template" field is displaying

For pipeline configurations, the situation is the same.

NShaforostov:
Checked

PR #482, opened 2019-07-02
Gui issue #418 incorrect behavior of "start idle" checkbox

AleksandrGorodetskii:
### There was case with incorrect behavior of ```"Start idle"``` checkbox and ```Cmd template``` input form field.

When you activate 'Start idle' checkbox - form field have to dissapear and save that presence. 
But when you change tab to another configuration - checkbox is unset and form field appears, even if its value is 'sleep infinity' (this value is condition to dissapearing form field and checkbox activation). And vise versa.

The reason of that behavior is: 
1. ```"Checked"``` condition for checkbox not defined.
2. ```startIdle``` state of component is forced to ```false``` every time component did update, regardless of form input value.

#### This Pull Request fixes this issue (#418) via:
1.  Added checked condition to checkbox value from state.
2. Added proper state condition logic, which depends of ```Cmd template``` value (matches value every time component re-render)
3. Trim cmd value before match with default string, so ``` sleep infinity   ``` with spaces before or after the end of a string will force to update ```startIdle``` to the proper state.

### DIFF
--- a/client/src/components/pipelines/launch/form/LaunchPipelineForm.js
+++ b/client/src/components/pipelines/launch/form/LaunchPipelineForm.js
@@ -1080,7 +1080,8 @@ export default class LaunchPipelineForm extends localization.LocalizedReactCompo
         fireCloudInputs: {},
         fireCloudOutputs: {},
         fireCloudDefaultInputs: this.props.fireCloudMethod ? this.props.fireCloudMethod.methodInputs : null,
-        fireCloudDefaultOutputs: this.props.fireCloudMethod ? this.props.fireCloudMethod.methodOutputs : null
+        fireCloudDefaultOutputs: this.props.fireCloudMethod ? this.props.fireCloudMethod.methodOutputs : null,
+        startIdle: this.props.parameters.cmd_template.trim() === 'sleep infinity'
       });
     }
     this.setState(state);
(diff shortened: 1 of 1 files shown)


client/src/components/pipelines/launch/form/LaunchPipelineForm.js
```diff
@@ -1080,7 +1080,8 @@ export default class LaunchPipelineForm extends localization.LocalizedReactCompo
         fireCloudInputs: {},
         fireCloudOutputs: {},
         fireCloudDefaultInputs: this.props.fireCloudMethod ? this.props.fireCloudMethod.methodInputs : null,
-        fireCloudDefaultOutputs: this.props.fireCloudMethod ? this.props.fireCloudMethod.methodOutputs : null
+        fireCloudDefaultOutputs: this.props.fireCloudMethod ? this.props.fireCloudMethod.methodOutputs : null,
+        startIdle: this.props.parameters.cmd_template.trim() === 'sleep infinity'
```
rodichenko: `startIdle` should be updated within `resetState` method, not `prepare` one.
Also consider if `cmd_template` parameter is not initialized (`undefined` or `null`)

client/src/components/pipelines/launch/form/LaunchPipelineForm.js
```diff
@@ -2961,7 +2962,8 @@ export default class LaunchPipelineForm extends localization.LocalizedReactCompo
                 (this.state.pipeline && this.props.detached)
               }
               onChange={(e) => this.setState({startIdle: e.target.checked})}
-              value={this.state.startIdle}>
+              value={this.state.startIdle}
```
rodichenko: I'm not sure we should keep `value={this.state.startIdle}`

rodichenko (changes requested):

rodichenko (approved):

Issue #485, opened 2019-07-03
Worker nodes shall be created in the same regions as parent run


mzueva:
# Background
Currently Cloud Pipeline supports a number of options to run a scalable cluster: autoscaled SGE cluster and WDL pipeline with `Use another compute node` option. When launching a new `child` run in such cluster currently cloud region is not specified directly and it may be created in a region other then parent run region, this operation will fail due to various reasons. It is possible to fix the problem in the scaling/cromwell scripts or in the CP API for all cases. 

# Action items
1. API shall checked for all scheduled runs:
- if region is specified - use it
- if parent is specified - use it's region
- otherwise use default region
2. SGE Autoscaling scripts shall run worker runs with parent region specified
3. WDL pipelines shall run tasks with parent region specified

sidoruka:
@mzueva please backport this to release/v0.15

mzueva:
Verified. Cherry-picked to release/0.15: 04a98e1f1ac72412e2b89b000053eb900eb6ff79 52d2bab18ae45004c5c48421baeb92f6c23d2dc7 6472eeb2b435cdc578644d9145afe7695ad38d0d

PR #490, opened 2019-07-03
Inherit cloud region from parent run

tcibinan:
Resolves first item of #485.

The pull request makes worker runs inherit parent's run cloud region.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
@@ -321,7 +321,9 @@ public class PipelineRunManager {
 
         List<String> endpoints = configuration.isEraseRunEndpoints() ? Collections.emptyList() : tool.getEndpoints();
         configuration.setSecretName(tool.getSecretName());
-        PipelineRun run = createPipelineRun(version, configuration, pipeline, parentRunId, entityIds, configurationId);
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -1004,11 +1022,8 @@ private RunInstance configureRunInstance(PipelineConfiguration configuration) {
                 .orElse(instance.getNodeDisk()));
         instance.setNodeType(configuration.getInstanceType());
         instance.setNodeImage(configuration.getInstanceImage());
-        AbstractCloudRegion cloudRegion = Optional.ofNullable(configuration.getCloudRegionId())
-                .map(id -> cloudRegionManager.load(id))
-                .orElse(cloudRegionManager.loadDefaultRegion());
-        instance.setCloudRegionId(cloudRegion.getId());
-        instance.setCloudProvider(cloudRegion.getProvider());
+        Optional.ofNullable(region).map(AbstractCloudRegion::getId).ifPresent(instance::setCloudRegionId);
```
mzueva, line 1025: `region` must be not null at this point

mzueva (approved):

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
tcibinan: It is true for common runs but for DTS runs it is fine to have no region specified.

PR #503, opened 2019-07-08
Issue 334 add new flyway placeholder default.admin.id to migration

SilinPavel:
Tris PR adds new flyway placeholder to db migration scripts, and adds opportunity to change default admin id.
In order to do that user should add `flyway.placeholders.default.admin.id` prop to property file while deploy server from scratch.


mzueva (changes requested):
@SilinPavel 
Could you please fix two things:
1. Move migration to the current date
2. Replace `default.admin.id` with new property in `AuthManager` class

SilinPavel:
@mzueva Done

mzueva (approved):

Issue #502, opened 2019-07-08
DTS recursive symlinks


sidoruka:
There were a number of cases when DTS was trying to upload data from the on-prem storage, where "recursive" symlinks were present.
This caused DTS to follow symlinks infinitely.

DTS shall try to detect such issues and skip the upload for files/folders, that cause infinite loop over symlinks.

mzueva:
@ekazachkova 

1. Add an option to `pipe storage cp/mv` operations to handle symlinks (for local source) :

- `follow` - default
- `skip` - do not follow symlinks 
- `filter` - follow symlinks but check for cyclic links (keep track of visited symlinks)

2. Add options to `dts` service to set symlink policy for transfer operations

NShaforostov:
Verified.
Works fine

PR #509, opened 2019-07-09
DTS recursive symlinks

ekazachkova:
Implementation for issue #502 

Added a new option `--symlinks (-sl)` that supports three possible value:
- follow - follow symlinks (default)
- skip - do not follow symlinks
- filter - follow symlinks but check for cyclic links

Also, dts application properties names was changed

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -856,6 +856,7 @@ def storage_move_item(source, destination, recursive, force, exclude, include, q
                                                         "should be tub delimited and consist of two columns: "
                                                         "relative path to file and size.")
 @click.option('-sl', '--symlinks', required=False, default="follow",
+              type=click.Choice(['follow', 'filter', 'skip']),
               help="Describe symlinks processing strategy for local sources. Possible values: "
               "follow - follow symlinks (default); "
               "skip - do not follow symlinks; "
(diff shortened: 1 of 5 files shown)


pipe-cli/src/model/data_storage_wrapper.py
```diff
@@ -407,21 +418,34 @@ def leaf_path(source_path):
...
-            def list_items(path, parent, root=False):
+            def list_items(path, parent, symlinks, visited_symlinks, root=False):
                 for item in os.listdir(path):
                     absolute_path = os.path.join(path, item)
+                    if os.path.islink(absolute_path) and symlinks != AllowedSymlinkValues.FOLLOW:
+                        if symlinks == AllowedSymlinkValues.SKIP:
+                            continue
+                        if symlinks == AllowedSymlinkValues.FILTER:
+                            symlink_target = os.readlink(absolute_path)
+                            if absolute_path.startswith(symlink_target):
```
mzueva: Could you explain, why do we need this condition?

pipe-cli/pipe.py
```diff
@@ -855,13 +855,19 @@ def storage_move_item(source, destination, recursive, force, exclude, include, q
 @click.option('-l', '--file-list', required=False, help="Path to file with file paths that should be copied. This file "
                                                         "should be tub delimited and consist of two columns: "
                                                         "relative path to file and size.")
+@click.option('-sl', '--symlinks', required=False, default="follow",
```
mzueva, line 858: Can't we use [Choice](https://click.palletsprojects.com/en/7.x/options/#choice-options) to simplify allowed values handling?

mzueva (changes requested):

pipe-cli/pipe.py
ekazachkova: thanks, added

pipe-cli/src/model/data_storage_wrapper.py
ekazachkova: fixed

mzueva (approved):

Issue #500, opened 2019-07-08
Warn user when a large file is copied/moved to/from the fuse-mounted storage


sidoruka:
# Background

At the moment - use may perform `cp` or `mv` operations of the large files (50+ Gb) to and from the fuse-mounted buckets.
It is uncovered that such operations are not handled properly within the fuse implementations. Commands may hang for hours/timeout/produce zero-sized result.

The suggested approach for the copies of the large files is to use `pipe cp` command, which is behaving correctly for the huge volumes.

# Approach

To make users aware of this - we shall warn them:
1. Implement wrappers for the `cp` and `mv` commands (as it was done for the package managers in #158)
2. If a `cp/mv` command is called with the `source/dest` pointing to the bucket (e.g. `/cloud-data/<bucket>/...`) - we shall check the overall size of the data being transferred
3. If the size is greater than allowed (shall be configurable via an environment variable in Gb, if not set - treat this as "unlimited") - a warning message shall be shown (stderr):
```
You are going to transfer data using FUSE-mounted object storage. Data transfer exceeds ${MAX_SIZE_GB} Gb.
Such operation may not work properly. For such data volumes please use the `pipe` command interface: <cp/mv command translated into pipe cp/mv here>
```
4. Note: we shall not abort the user's command - only warning shall be shown

# Things to consider

It is not clear yet how to distinguish object storage paths and others, e.g.:
- NFS/SMB mounts
- Cluster wide LFS
- Different mount path and symlinks (e.g. `/cloud-data/ -> ~/cloud-data`)

Any proposals on these points are welcomed.

sidoruka:
@tcibinan while #511 is merged, would you like to layout some items for the further work on this (some scenarios/options handling/...)?

tcibinan:
Yes, of course. There are several `cp` and `mv` commands use cases that the merged pull request doesn't cover:
1. The usage of `--target-directory` parameter in `equals sign` notation avoids the restriction from being applied.
```
cp --target-directory=/destination/path /cloud-data/storage/folder
```
2. The usage of non-typical named arguments produces misleading offered `pipe cli` replacement.
```
cp --backup -S _suffix /cloud-data/storage/folder /destination/path
    ...
    pipe storage cp -rf /destination/path/_suffix /destination/path/
    pipe storage cp -rf cp://storage/folder /destination/path/
    ...
```

PR #511, opened 2019-07-10
Introduce storage mounts data transfer restrictor

tcibinan:
Resolves issue #500.

Pull request adds storage mounts data transfer restrictor for `cp` and `mv` commands. The restrictor warns user every time `cp` or `mv` command is used to download to storage mount, upload from storage mount or transfer between storage mount large enough data.

There are several environment variables that can be used to configure transfer manager behaviour:
- `CP_ALLOWED_MOUNT_TRANSFER_SIZE` - number of *gigabytes* that is allowed to be transferred without warning. By default is not specified and any amount of data is allowed.
- `CP_ALLOWED_MOUNT_TRANSFER_SIZE_TIMEOUT` - number of *seconds* that the transfer size retrieving operation can take. By default *5 seconds*.
- `CP_ALLOWED_MOUNT_TRANSFER_FILES` - number of files that is allowed to be transferred without warning. *Supported only for Azure cloud provider*. By default *100 files*.

Azure transfer getting size implementation differs from other cloud providers due to an existing issue https://github.com/Azure/azure-storage-fuse/issues/119.

### DIFF
--- a/deploy/contents/install/preferences/launch.env.properties.json
+++ b/deploy/contents/install/preferences/launch.env.properties.json
@@ -2,6 +2,7 @@
     "variables": {
         "CP_CLUSTER_NAME": "CLOUD_PIPELINE",
         "CP_RESTRICTED_PACKAGES": "nvidia-",
-        "CP_RESTRICTING_PACKAGE_MANAGERS": "apt,apt-get,aptitude,yum,pip,pip2,pip3"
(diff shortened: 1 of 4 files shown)


tcibinan:
[Set of commands](https://github.com/epam/cloud-pipeline/files/3381306/transfer_restrictor_validation_tests.txt) to roughly test out transfer restrictions is attached.

Issue #406, opened 2019-06-19
Elasticsearch Agent shall use Spring Data to fetch event data from DB


mzueva:
# Background
Current implementation of `elasticsearch-agent` uses Spring JDBC to fetch Events data from PostgreSQL. It would be great to replace it with Spring Data to eliminate boilerplate code.

# Action items
- Add `spring-data` dependency to `elasticsearch-agent` service
- Provide entity mapping for `PipelineEvent` class
- Provide `repository` for `PipelineEvent` class
- Change usage of `PipelineEventDao` with `repository`
- Provide tests for the new approach

PR #513, opened 2019-07-11
Issue #406: added Spring Data JPA implementation to fetch event data from DB

DmitryPuhovsky:
Implementation for issue #406

This PR provides proof of concept for using `Spring Data JPA` with `elasticsearch-agent` service.
`Spring Data JPA` support was added to fetch event data from DB instead of `JDBC`.

The following changes was added to `pipeline_event` table:
-  a new column `id` was added
-  type for columns `operation` and `object_type` was changed from `text` to `integer`


mzueva:
Closing this PR for now, maybe we'll return to this idea later.

Issue #415, opened 2019-06-20
File preview doesn't update without page refreshing


NShaforostov:
1. Create any file with the content in a storage
2. Click it to open the attributes panel
3. Click the "full view" button for the file content
4. Edit the file, save changes

**_Exp._**: file preview is changed
**_Act._**: file preview is still previous even you reopen the attributes panel or navigate to another  object in the library tree and then return to the created file

NShaforostov:
Verified. Fixed

PR #519, opened 2019-07-12
Gui issue #415 file preview doesnt update without page refreshing

AleksandrGorodetskii:
#### There was case with incorrect file preview data, after changes to file was apply.

If you change file, file preview was not updated due to cached content of file (always apply previous file content from cache, Until you send new get request from server)

#### This Pull Request fixes this issue (#415 ) via:
1. Injected ```DataStorageCache``` to ```DataStorage``` component.
2. After changed file was saved - get current file Id, path, version
3. Fetch that file from server.


client/src/components/pipelines/browser/DataStorage.js
```diff
@@ -35,6 +35,7 @@ import {
   Spin,
   Table
 } from 'antd';
+import dataStorageCache from '../../../models/dataStorage/DataStorageCache';
```
rodichenko: This import is redundant - `dataStorageCache` should be injected (it is initialized in `Root.js`)

client/src/components/pipelines/browser/DataStorage.js
```diff
@@ -481,14 +483,21 @@ export default class DataStorage extends React.Component {
   };
 
   saveEditableFile = async(path, content) => {
+    const currentItemContent = dataStorageCache.getContent(
```
rodichenko: We should use injected `dataStorageCache` (via component's `props`)

rodichenko (changes requested):

rodichenko (approved):

Issue #493, opened 2019-07-04
Setting of the tool icon size


NShaforostov:
### Case 1
1. Open system-level settings, select **Preferences** tab
2. Set a new value into the `misc.max.tool.icon.size.kb` preference that is bigger than default 50 Kb, e.g. 200
3. Save and close pop-up
4. Open any tool, try to upload an icon with the size in the range from 50 to 200 Kb

**_Exp._**: a new icon will be uploaded
**_Act._**: the follow-up error appears: ![2](https://user-images.githubusercontent.com/45459424/60665995-9f6fb480-9e6e-11e9-9e59-dc934f6a952b.png), a new icon isn't uploaded

Is it a normal behavior? If yes - probably, we should implement a warning or something else to restrict the input into the `misc.max.tool.icon.size.kb` preference?

### Case 2
1. Open system-level settings, select **Preferences** tab
2. Set a new value into the `misc.max.tool.icon.size.kb` preference that is smaller than default 50 Kb, e.g. 30
3. Save and close pop-up
4. Open any tool, try to upload an icon with the size in the range from 30 to 50 Kb
5. An error appears, e.g.:  
![1](https://user-images.githubusercontent.com/45459424/60665073-253e3080-9e6c-11e9-94b7-9a3dc4439884.png)

Can we make it a bit more graceful? For example: **`Uploaded icon is too large: 43,388 bytes. Only 30,000 bytes allowed`** or just in a style described above: **`Maximum file size is 30kb`**

### Case 3
1. Repeat steps 1-3 from **Case 2**
2. Open any tool, try to upload an icon with the size more than 50 Kb

**_Exp._**: an error in a style like in **Case 2** appears
**_Act._**: the follow-up error appears: ![2](https://user-images.githubusercontent.com/45459424/60665995-9f6fb480-9e6e-11e9-9e59-dc934f6a952b.png)

Probably, we should unify error messages for **Case 2** and **Case 3**?

NShaforostov:
Verified. Fixed

PR #529, opened 2019-07-16
Gui issue #493 setting of the tool icon size

AleksandrGorodetskii:
### There was case with incorrect error message, after you try to upload new icon, that exceeded file size from preferences.

That was because <Tool>  component did not receive ```misc.max.tool.icon.size.kb``` and always try to compare file size with constant value (50kB) despite the actual preferences value. As a result - threw an incorrect error message if file was exceeded 50kB.

### This Pull Request fixes this issue (#493  ) via:
1. ```misc.max.tool.icon.size.kb``` value from preferences now passed to <Tool> component
2. Before uploading the new icon - size of file will match preferences max-icon-size value. If <Tool> doesnt get value from preferences, default_value uses instead.


client/src/components/tools/Tool.js
```diff
@@ -345,13 +346,14 @@ export default class Tool extends localization.LocalizedReactComponent {
         </Row>
       );
     }
+
     const doUpload = (file) => {
-      const MAX_FILE_SIZE_KB = 50;
+      const maximumFileSize = this.props.preferences.getPreferenceValue('misc.max.tool.icon.size.kb') || DEFAULT_FILE_SIZE_KB;
```
rodichenko: The last thing: please define `getter` for this preference in `preferences` class (just like `useSpot`, `deploymentName` etc) with `undefined` value if it's not set **or not a number** and **converted to a number** value otherwise (we don't want to compare `string` and `number`).

rodichenko (changes requested):

rodichenko (approved):

Issue #501, opened 2019-07-08
Estimate run's disk size according to the input/common parameters


sidoruka:
# Background

If a job is run with the disk size, which is not enough to handle the job's inputs - it will fail. (e.g. 10Gb disk is set for a run, which will process data using `STAR` aligner, where the genome index file is 20Gb)

# Approach

As an attempt to handle some of such cases - let's try to estimate the required disk size using the input parameters and warn the user if the requested disk is not enough.

1. When launching a job - try to get the size of all input/common parameters
2. We shall limit the time required to traverse all files, as this may take too much for lots of small files. Default: 3 sec
2. When comparing the data and disk sizes - we shall consider cluster configuration:
- For static cluster - sum of all node's and master disks shall be used for comparison
- For autoscaled cluster - sum of all initial node's and master disks shall be used for comparison (autoscaled nodes shall not be considered)
4. If the resulting size is greater than requested - warn the user (WebGUI and CLI):
```
The requested disk size for this run is N Gb, but the data that is going to be processed exceeds it (estimated volume: M Gb). This may cause run to fail with "Out Of Disk" error.
Do you want to use the suggested disk size K Gb (which also considers interim and resulting datasets)?
```
5. Where `K` is calculated as:
- Sum of input parameters (limited by the files number)
- Multiplied by `3` (to keep in mind the interim and resulting files)
- Shall not exceed 16Tb

6. If calculated `K` exceeds 16Tb we shall show a different warning message:
```
The requested disk size for this run is N Gb, but the data that is going to be processed exceeds 16 Tb (which is a hard limit). Please use the cluster run configuration to scale the disks horizontally or reduce the input data volume.
Do you want to use the maximum disk size 16 Tb anyway?
```

mzueva:
@rodichenko new API method is deployed to dev `POST /datastorage/path/size`

rodichenko:
@sidoruka @mzueva GUI part implemented (via 977e09a6f07d44efb59c390737db156441e9d998)

sidoruka:
@mzueva @rodichenko please backport this feature to `release/0.15`

rodichenko:
@sidoruka GUI backported to `release/0.15` via 3853cac1c9c1c7a5bde362028551e4c449585f4e

mzueva:
@sidoruka cherry-picked a5d5751038affd0864a6610d38fc07e2ef8bfc0b

NShaforostov:
Is a user warning implemented via CLI?

PR #544, opened 2019-07-23
Issue #501: Estimate run's disk size according to the input/common parameters

ekazachkova:
This PR provides implementation for issue #501 
 
A new API method `datastorage/path/size` was added. This method takes  an array with paths (from input parameters) and returns their sizes in bytes with the following format:
```
[
   {
      "path": "s3://my-bucket/folder/",
      "size": "1984",
      "dataStorageId": "1",
      "completed": "true"
   } 
]
```
Also, preference `storage.listing.time.limit` with default value 3000 ms was added.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/controller/datastorage/DataStorageController.java
+++ b/api/src/main/java/com/epam/pipeline/controller/datastorage/DataStorageController.java
@@ -25,7 +25,6 @@ import com.epam.pipeline.controller.vo.data.storage.UpdateDataStorageItemVO;
 import com.epam.pipeline.controller.vo.security.EntityWithPermissionVO;
 import com.epam.pipeline.entity.datastorage.AbstractDataStorage;
 import com.epam.pipeline.entity.datastorage.AbstractDataStorageItem;
-import com.epam.pipeline.entity.datastorage.TemporaryCredentials;
 import com.epam.pipeline.entity.datastorage.DataStorageAction;
 import com.epam.pipeline.entity.datastorage.DataStorageDownloadFileUrl;
 import com.epam.pipeline.entity.datastorage.DataStorageException;
@@ -34,6 +33,8 @@ import com.epam.pipeline.entity.datastorage.DataStorageItemContent;
 import com.epam.pipeline.entity.datastorage.DataStorageListing;
 import com.epam.pipeline.entity.datastorage.DataStorageStreamingContent;
 import com.epam.pipeline.entity.datastorage.DataStorageWithShareMount;
+import com.epam.pipeline.entity.datastorage.TemporaryCredentials;
+import com.epam.pipeline.entity.datastorage.PathDescription;
 import com.epam.pipeline.entity.datastorage.rules.DataStorageRule;
 import com.epam.pipeline.manager.datastorage.DataStorageApiService;
 import io.swagger.annotations.Api;
@@ -46,7 +47,6 @@ import org.apache.commons.fileupload.FileItemStream;
 import org.apache.commons.fileupload.FileUploadException;
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
 import org.apache.commons.io.FilenameUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Controller;
@@ -643,7 +643,7 @@ public class DataStorageController extends AbstractRestController {
         return Result.success(dataStorageApiService.getStoragePermission(page, pageSize, filterMask));
     }
 
-    @PostMapping(value = "/datastorage/sizes")
+    @PostMapping(value = "/datastorage/path/size")
     @ResponseBody
     @ApiOperation(
             value = "Returns full size specified by path.",
@@ -652,7 +652,7 @@ public class DataStorageController extends AbstractRestController {
     @ApiResponses(
             value = {@ApiResponse(code = HTTP_STATUS_OK, message = API_STATUS_DESCRIPTION)
             })
-    public Result<List<ImmutablePair<String, Long>>> getDataSizes(@RequestBody final List<String> paths) {
+    public Result<List<PathDescription>> getDataSizes(@RequestBody final List<String> paths) {
         return Result.success(dataStorageApiService.getDataSizes(paths));
     }
 }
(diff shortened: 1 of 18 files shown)


api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -575,6 +576,34 @@ public String adjustStoragePath(final String name, final DataStorageType storage
...
+                .map(path -> new ImmutablePair<>(path, getDataSize(path)))
+                .collect(Collectors.toList());
+    }
+
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private Long getDataSize(final String path) {
+        try {
+            Assert.state(StringUtils.isNotBlank(path), messageHelper
+                    .getMessage(MessageConstants.ERROR_DATASTORAGE_PATH_IS_EMPTY));
+            final String[] pathParts = path.split(ProviderUtils.DELIMITER);
```
mzueva: I think is more simple to parse path as URI.

api/src/main/java/com/epam/pipeline/controller/datastorage/DataStorageController.java
```diff
@@ -642,4 +643,16 @@ public Result restoreFileVersion(
...
+    @PostMapping(value = "/datastorage/sizes")
+    @ResponseBody
+    @ApiOperation(
+            value = "Returns full size specified by path.",
+            notes = "Returns full size specified by path.",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ApiResponses(
+            value = {@ApiResponse(code = HTTP_STATUS_OK, message = API_STATUS_DESCRIPTION)
+            })
+    public Result<List<ImmutablePair<String, Long>>> getDataSizes(@RequestBody final List<String> paths) {
```
mzueva: I'd prefer to add a simple class `PathDescription` or smth like that rather than use `Pair`

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageApiService.java
```diff
@@ -316,4 +317,8 @@ public EntityWithPermissionVO getStoragePermission(Integer page, Integer pageSiz
         return grantPermissionManager
                 .loadAllEntitiesPermissions(AclClass.DATA_STORAGE, page, pageSize, true, mask);
     }
+
+    public List<ImmutablePair<String, Long>> getDataSizes(final List<String> paths) {
```
mzueva: We need to check write permissions for paths (introduce a custom method for this check)

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/aws/s3/S3Helper.java
```diff
@@ -603,6 +604,31 @@ public boolean checkBucket(String bucket) {
...
+        final String requestPath = Optional.ofNullable(path).orElse("");
+        final AmazonS3 client = getDefaultS3Client();
+        final boolean rootOrFolder = ProviderUtils.isRootOrFolder(requestPath);
+
+        ObjectListing listing = client.listObjects(dataStorage.getPath(), requestPath);
+        long folderSize = 0L;
+        boolean hasNextPageMarker = true;
+        while (hasNextPageMarker) {
+            for (final S3ObjectSummary summary : listing.getObjectSummaries()) {
+                if (rootOrFolder) {
```
mzueva: Code duplication in `ProviderUtils` ?

api/src/main/java/com/epam/pipeline/controller/datastorage/DataStorageController.java
```diff
@@ -642,4 +643,16 @@ public Result restoreFileVersion(
         return Result.success(dataStorageApiService.getStoragePermission(page, pageSize, filterMask));
     }
 
+    @PostMapping(value = "/datastorage/sizes")
```
mzueva: `/datastorage/path/size`

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/ProviderUtils.java
```diff
@@ -45,4 +47,26 @@ public static String normalizeBucketName(String name) {
...
     }
+
+    public static <T> Long getSizeByPath(final Iterable<T> items, final String requestPath,
+                                         final Function<T, Long> getSize, final Function<T, String> getName) {
+        final boolean rootOrFolder = isRootOrFolder(requestPath);
+
+        Long folderSize = 0L;
+        for (final T item : items) {
+            if (rootOrFolder) {
+                folderSize += getSize.apply(item);
```
mzueva: I'm not sure but how `size` may be returned for folder? I thought that approach was to list objects resursively, in this case we should have only objects (files) not folders.

mzueva (changes requested):
Execution of method for getting total object size shall be limited by time in milliseconds. If computation doesn't end in this timeout we shall return accumulated size as is. If object size wasn't checked for any of paths due to timeout return some predefined value (e.g. -1).

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
ekazachkova: thanks, done

api/src/main/java/com/epam/pipeline/controller/datastorage/DataStorageController.java
ekazachkova: done

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageApiService.java
ekazachkova: done

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/aws/s3/S3Helper.java
ekazachkova: now yes, done

api/src/main/java/com/epam/pipeline/controller/datastorage/DataStorageController.java
ekazachkova: done

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/ProviderUtils.java
ekazachkova: comment added

mzueva (approved):

Issue #538, opened 2019-07-19
Elasticsearch agent: application properties for ES shall resolve environments variables


ekazachkova:
Currently, some settings  for `elasticsearch-agent` should be written as is in the `application.properties` file. It may lead to leakage of secret data such as tokens, passwords and etc.  To resolve this problem let's use environments variables instead of the real ones.

The following values should be added to the `application.properties` file:
```
elasticsearch.client.url=${ES_CLIENT_URL}
elasticsearch.client.port=${ES_CLIENT_PORT}
database.url=${PGS_URL}
database.username=${PGS_USER}
database.password=${PGS_PASS}
cloud.pipeline.host=${API}
cloud.pipeline.token=${API_TOKEN}
```
Also, make sure that all properties are resolved correctly.

PR #545, opened 2019-07-23
Issue #538: Some properties in ES Application.properties use environment variables

DmitryPuhovsky:
Resolves issue #538

The following values in the application.properties file were changed for use evironment variables:

- elasticsearch.client.url=${ES_CLIENT_URL}
- elasticsearch.client.port=${ES_CLIENT_PORT}
- database.url=${PGS_URL}
- database.username=${PGS_USER}
- database.password=${PGS_PASS}
- cloud.pipeline.host=${API}
- cloud.pipeline.token=${API_TOKEN}

### DIFF
--- a/elasticsearch-agent/src/main/resources/application.properties
+++ b/elasticsearch-agent/src/main/resources/application.properties
@@ -3,21 +3,21 @@ server.port=8080
 server.servlet.context-path=/elastic-agent
 
 #ElasticSearch Client Configuration
-elasticsearch.client.url=localhost
-elasticsearch.client.port=9200
+elasticsearch.client.url=${ES_CLIENT_URL:localhost}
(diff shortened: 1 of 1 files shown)


elasticsearch-agent/src/main/resources/application.properties
```diff
@@ -3,21 +3,21 @@ server.port=8080
 server.servlet.context-path=/elastic-agent
 
 #ElasticSearch Client Configuration
-elasticsearch.client.url=localhost
-elasticsearch.client.port=9200
+elasticsearch.client.url=${ES_CLIENT_URL}
```
ekazachkova: may be we should add default values for all properties where it is possible?

DmitryPuhovsky: Done.

Issue #557, opened 2019-07-26
Cluster runs are paused by "Autopause" but can't be resumed


NShaforostov:
Currently, cluster runs couldn't be paused manually even if price type is "**On-demand**".
But if cluster run was launched with enabled "**Auto pause**" option, parent-run or its child-runs will be paused (when autopause conditions are satisfied, of course). In that case, user couldn't resume such paused runs and go on his work because only "**Terminate**" buttons will be available:
![1](https://user-images.githubusercontent.com/45459424/61949435-9082aa80-afb3-11e9-9cf9-ec12c744330e.png)

sidoruka:
Autopause daemon shall not handle clusters

sidoruka:
Backport to `release/0.15`

NShaforostov:
Verified. Fixed.
@SilinPavel, please backport to `release/0.15`

SilinPavel:
@NShaforostov Done.

PR #562, opened 2019-07-30
Support to not handle clusters in autopause daemon

kamyshova:
Resolves issue #557 
The pull request added a support to not handle clusters in autopause daemon

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
@@ -275,7 +275,7 @@ public class ResourceMonitoringManager extends AbstractSchedulingManager {
                                                  action));
             switch (action) {
                 case PAUSE:
-                    if (run.getInstance().getSpot()) {
+                    if (run.getInstance().getSpot() || isClusterRun(run)) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -275,15 +275,15 @@ private void performActionOnIdleRun(PipelineRun run, IdleRunAction action, doubl
...
+                    if (run.getInstance().getSpot() || isClusterRun(run)) {
                         performNotify(run, cpuUsageRate, pipelinesToNotify);
                     } else {
                         performPause(run, cpuUsageRate);
                     }
 
                     break;
                 case PAUSE_OR_STOP:
-                    if (run.getInstance().getSpot()) {
+                    if (run.getInstance().getSpot() || isClusterRun(run)) {
```
SilinPavel: should we stop cluster here, or we shouldn't touch cluster's node at all?
f.e.
If our goal do not deal with cluster node at all we can simply do this on top of the method `processIdleRun`:
```
// filter cluster's runs
if (isClusterRun(run)) {
    return;
} 
...
...
```

SilinPavel (changes requested):

SilinPavel (approved):

Issue #549, opened 2019-07-24
GCP: Mounting bucket is failed to launched tool


kamyshova:
**Environment**: `GCP`
**Version**: `0.16.0.1916.46d7e7d065cf7db5eab314572f25e2631d124721`

A bucket with access is not mounted to launched tool and the following error appears in `MountDataStorages` task logs:

![image](https://user-images.githubusercontent.com/23356251/61791123-c55ef800-ae21-11e9-9085-c5966775bc5e.png)

sidoruka:
@kamyshova this shall work fine now, could you please recheck?

kamyshova:
@sidoruka Verified. Fixed, thanks

PR #569, opened 2019-07-31
Issue #549 install gcsfuse from public bucket

SilinPavel:
In order to solve #549 this PR introduce new way to install gcsfuse, firstly it will try to install it from public s3 bucket `s3://cloud-pipeline-oss-builds/` and only if this way failed, will try to install it from repo.

### DIFF
--- a/workflows/pipe-common/shell/install_gcsfuse
+++ b/workflows/pipe-common/shell/install_gcsfuse
@@ -77,10 +77,10 @@ if [ $? -ne 0 ]; then
     _IS_INSTALLED_FROM_PACKAGE_REPOSITORY=$?
     if [ $_IS_INSTALLED_FROM_PACKAGE_REPOSITORY -eq 0 ]; then
         echo "${_FUSE_TYPE} installed from package repository"
-        exit 0
     else
         echo "${_FUSE_TYPE} installation from package repository failed"
         exit 1
     fi
+else
+    echo "${_FUSE_TYPE} installed from local repository"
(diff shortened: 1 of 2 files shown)


workflows/pipe-common/shell/install_gcsfuse
```diff
@@ -57,6 +57,57 @@ EOF
     fi
 }
 
+function check_installed {
+      local _COMMAND_TO_CHECK=$1
+      command -v "$_COMMAND_TO_CHECK" >/dev/null 2>&1
+      return $?
+}
+
+function local_package_install {
```
sidoruka: It may make sense to move the `local_package_install` logic (and other corresponding functions) to a separate script, that can be further reused? We'll definitely perform such operations for other components

sidoruka (changes requested):
@SilinPavel can you please confirm that this functionality was tested for all the supported distros:
* Ubuntu 16
* Ubuntu 18
* Debian 9
* Centos 7

workflows/pipe-common/shell/install_gcsfuse
SilinPavel: good point

SilinPavel:
@sidoruka I checked that script works for all this distro:
-  Ubuntu 16
-  Ubuntu 18
-  Debian 9
-  Centos 7

mzueva (approved):

sidoruka (approved):

Issue #554, opened 2019-07-25
Handle resource limit errors during run resuming


mzueva:
# Background
We may hit resource limits while trying to resume previously paused run. E.g. instance type was available when run was initially launched, but at the moment of resume operation provider has no sufficient capacity for this type. In this case AWS will through an error when we try to start a stopped instance:

`We currently do not have sufficient p2.8xlarge capacity in the Availability Zone you requested
 (us-east-1a). Our system will be working on provisioning additional capacity. You can currently get p2.8xlarge capacity by not specifying an Availability Zone in your request or choosing us-east-1b, us-east-1c, us-east-1d, us-east-1e, us-east-1f. (Service: AmazonEC2; Status Code: 500; Error Code: InsufficientInstanceCapacity; Request ID: XXX)`

# Action items
1. Check how Azure and GCP work in such case
2. Check whether the same problem exists with account/subscription limits
3. Implement `InsufficientInstanceCapacity` and similar errors handling: if this error occurs during run resume, do not fail run, instead:
 - add log message for `ResumeRun` task
 - return run to `Paused` state 
 - API Client shall be notified about such an event. So that GUI shall be able to display a corresponding warning message

sidoruka:
If feasible - backport to `release/0.15`

mzueva:
@rodichenko @SilinPavel I'd suggest to add field `reason` to `RunStatus` and fill it in case of resume operation failure.

mzueva:
@rodichenko In the latest version API will return for `PipelineRun` an optional `reason` text field in list of `runStatuses`. In case of specific errors during resume operation it will contain reason for resume failure and returning back to paused state.

mzueva:
Backend part cherry-picked to release 3d64599

rodichenko:
@sidoruka @mzueva GUI part implemented. This feature backported to release/0.15 via fe6c4b5f5d59b6434098b0b9ff0d2f97e43ccf6e

rodichenko:
Warning messages are displayed at:
- **Run log** form:
![warning2](https://user-images.githubusercontent.com/5070496/62785738-ec782380-bac9-11e9-8326-8a36e4ad3d26.png)
- **Dashboard**:
![warning3](https://user-images.githubusercontent.com/5070496/62785752-f39f3180-bac9-11e9-92a5-6fc66d81ab62.png)
- **Active runs table**:
[...]

PR #584, opened 2019-08-07
issue #554 add new field reason to RunStatus

SilinPavel:
This PR adds new field to RunStatus class.
Now when resume can't be done by some reason we will fill it with massage


api/src/main/java/com/epam/pipeline/manager/notification/NotificationAspect.java
```diff
@@ -89,5 +96,13 @@ public void notifyRunStatusChanged(JoinPoint joinPoint, PipelineRun run) {
                      run.getPodId(), run.getPipelineName(), run.getVersion(), run.getStatus());
         notificationManager.notifyRunStatusChanged(run);
     }
+
+    private String getStatusReason(final PipelineRun run) {
```
mzueva: @SilinPavel Maybe we could just pass the status in `PipelineRun.stateReasonMessage`? It is also not the best solution.

mzueva (changes requested):

mzueva (approved):

Issue #397, opened 2019-06-17
Azure: Handle low-priority termination


mzueva:
Currently Azure low-priority VMs termination is not handled correctly - we just fail run with `NodeLost` error, which makes automatic restart impossible.

## Solution
Azure implementation for `CloudInstanceService.getInstanceTerminationState` shall handle `VMScaleSets` and determine whether instance is down due to low-priority termination. We also need to verify that setting appropriate status code to `instance.restart.state.reasons` will enable restart of low-priority Azure VMs.

mzueva:
Problem still remains open

mzueva:
@SilinPavel Let's try to use [VM scheduled Events](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/scheduled-events) to track low priority VM eviction. Introduce a script that will run on VM, track events and in case of preemtion label corresponding node in k8s with "preemted=true" label.

PR #590, opened 2019-08-12
issue #397 New approach for Low Priority Instances with azure events service

SilinPavel:
This PR solves #397

### DIFF
--- a/scripts/autoscaling/init_multicloud.sh
+++ b/scripts/autoscaling/init_multicloud.sh
@@ -250,6 +250,10 @@ elif [[ $cloud == *"Microsoft"* ]]; then
         _CLOUD_INSTANCE_IMAGE_ID=""
     fi
     _CLOUD_PROVIDER=AZURE
+
+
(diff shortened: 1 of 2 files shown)


scripts/autoscaling/init_multicloud.sh
```diff
@@ -250,6 +250,12 @@ elif [[ $cloud == *"Microsoft"* ]]; then
         _CLOUD_INSTANCE_IMAGE_ID=""
     fi
     _CLOUD_PROVIDER=AZURE
+
+    CHECK_AZURE_EVENTS_COMMAND="curl -k -H Metadata:true http://169.254.169.254/metadata/scheduledevents?api-version=2017-11-01 2> /dev/null | grep Preempt && kubectl label node $(hostname) preempted=true --kubeconfig='/etc/kubernetes/kubelet.conf'"
```
sidoruka: Maybe add `-q` for the grep?

sidoruka: Maybe use something more unique for the label name, e.g. `cloud-pipeline/preempted` or similar? As the `preempted` is becoming a quite common wording.

sidoruka (changes requested):

scripts/autoscaling/init_multicloud.sh
```diff
@@ -250,6 +250,12 @@ elif [[ $cloud == *"Microsoft"* ]]; then
...
+
+    # run 3 cron jobs to be able to check events each 20 seconds
+    crontab -l | { cat ; echo -e "* * * * * $CHECK_AZURE_EVENTS_COMMAND \n* * * * * sleep 20 && $CHECK_AZURE_EVENTS_COMMAND \n* * * * * sleep 40 && $CHECK_AZURE_EVENTS_COMMAND" ; } | crontab -
```
sidoruka, line 257: I was thinking about this a bit more - why not to use a simple shell script with an infinite loop and a `sleep` instead of the complex `cron` configuration?
@SilinPavel do you foresee any issues with that?

SilinPavel: @sidoruka I just think that it would 100% avoid any problem between iteration run. And each time cron run will be independent of previous one.

scripts/autoscaling/init_multicloud.sh
SilinPavel: agree

SilinPavel: fixed

Issue #427, opened 2019-06-21
Add block device swap support


sidoruka:
Swap via files, introduced by #199, may take too much time to be created.
We shall implement an optional way to use a dedicated block device for the swap

mzueva:
Ported to release 0.15 for AWS via b7966a0c1715807642f196a83a607441625af186

Issue #203, opened 2019-04-23
GCP: Incorporate swap configuration from the AWS/Azure


sidoruka:
Once #199 is resolved and tested - please incorporate those change into the GCP implementation:
- Support for GCP itself
- Port to the refactored version of the `nodeup.py`
- Maybe replace current `init.sh` templating approach with something more common/smart

PR #593, opened 2019-08-12
Swap device support

mzueva:
This PR enables support of swap configuration using a separate block device and provides solution for issues #427 and #203.

# Implementation details
1. Swap is configured using `swap` section in `cluster.networks.config` preference. The only supported/required configuration is `swap_ratio` with sets swap/ram ration. For example, with the following configuration for each run an additional swap block device will be configured with `device size = instance_ram_gb * 0.1`:

`"swap": [
    {
     "name": "swap_ratio",
     "path": "0.1"
    }]`

2. Swap device is configured in `init_multicloud.sh` script. Block device is selected by size. Block device is configured only if a disk with matching size exists and it is not the only one disk.

### DIFF
--- a/scripts/autoscaling/azure/nodeup.py
+++ b/scripts/autoscaling/azure/nodeup.py
@@ -15,6 +15,7 @@
 import argparse
 import functools
 import os
+import math
 import json
 import sys
 import re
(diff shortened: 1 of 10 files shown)


sidoruka (changes requested):
@mzueva, could you please also change the default settings, which are configured via `pipectl`?
E.g.: https://github.com/epam/cloud-pipeline/blob/fc4244e8159135840ab64877052b8040a01807c2/deploy/contents/install/cloud-configs/aws/cluster.networks.config.json#L22-L31

scripts/autoscaling/azure/nodeup.py
```diff
@@ -358,33 +360,46 @@ def get_os_profile(instance_name, ssh_pub_key, user, user_data_script, computer_
...
 
 
-def get_storage_profile(disk, image, instance_type):
+def get_data_disk(size, type, lun, disk_name=None):
+    disk = {
+                "diskSizeGB": size,
+                "lun": lun,
+                "createOption": "Empty",
+                "managedDisk": {
+                    "storageAccountType": type
```
SilinPavel: maybe it would be better to call this parameter like `disk_type` (and the others if it exists) to avoid using reserved python words as parameter name

scripts/autoscaling/azure/nodeup.py
```diff
@@ -851,11 +867,76 @@ def replace_proxies(cloud_region, init_script):
...
+        ram = get_ram_from_group(instance_types, 'cluster.allowed.instance.types.docker', ins_type)
+    return ram
+
+
+def get_ram_from_group(instance_types, group, instance_type):
+    if group in instance_types:
+        for type in instance_types[group]:
+            if type['name'] == instance_type:
+                return type['memory']
+
```
SilinPavel, line 913: maybe `return None` is missing

mzueva: Yes, fixed.

scripts/autoscaling/azure/nodeup.py
mzueva: Yes, fixed.

Issue #589, opened 2019-08-12
pipe CLI shall allow to update the executable automatically


sidoruka:
1. `pipe` CLI shall expose `update` command, that will grab the latest version of the CLI from the API instance and install it, instead of the existing one
2. This command shall get the API URL from:
- Command-line parameter (e.g. `pipe update https://cloud/pipeline/`)
- If this parameter is not set:
  - Environment variable (`$API`)
  - Configuration JSON file
- If none of the above is set - an error shall be thrown

sidoruka:
@maryvictol please verify this feature using:
* Linux
* Windows

maryvictol:
`pipe update` command has been verified on Version: 0.17.0.6278 using:
- Windows: checked `pipe update` and `pipe update <API URL>` variants -  it works well
- Linux-Binary: checked `pipe update` variant -  it works well 
- Linux-Tarball: checked `pipe update` variant -  it works well

PR #615, opened 2019-08-16
pipe CLI shall allow to update the executable automatically

ekazachkova:
This PR provides implementation for issue #589 

a new CLI option `update` is introduced with optional argument (API url). Usage examples:
```
pipe update <api url>
```
If API url is not specified the `pipe` will try to find this url in environment variables (`$API`) or configuration JSON.

The behavior of the update operation depends on the OS.

#### For Linux binary:
- deletes current pipe executable file
- downloads the latest version from API

#### For Windows binary:
- download the latest version of pipe CLI to temporary directory
- removes all files under the directory that contains `pipe.exe`
- copies files from tmp directory to src directory described at the previous  point
- removes temporary directory

#### For other OSes (i.e. installed from the sources dist tarball):
- an error message will be shown and no update will happen

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 import click
 import requests
 import sys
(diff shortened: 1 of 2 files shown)


sidoruka:
I'm ok with this PR.
Can we please merge it and add version check features (#697)?

Issue #623, opened 2019-08-23
Cloud Pipeline shall allow to "block" the user


sidoruka:
User blocking/unblocking shall be available in a manual and an automated manners:
1. Manual:
- Administrator shall be able to set the "blocked" flag for the specific user or a whole group
- This flag will prevent user(s) to access Cloud Pipeline platform by all the means (GUI/CLI/API)
- This flag can be set either from the Web GUI (User management console) and Cloud Pipeline CLI (for automation reasons)
2. Automated:
- If a user is blocked in the IdP 
- API shall be configurable to set the "blocked" attribute/claim name (which will be analyzed in the SAMLResponse object)
- In this case - "blocked" flag will be set for the user automatically, without admin's manual actions

For both options:
- Any content generated by the blocked users shall be kept in the Cloud Pipeline
- All the log trails on the users activities shall be kept as well
- According to other policies, e.g. retention, some of the user owned assets may be deleted

mzueva:
#### First part of implementation
1. Add `blocked` field to user table in db adding a new flyway migration
2. Add `blocked` field to PipelineUser model
3. Add handling of this new field to `UserDao` and related xml file
4. Add `blockUser` field API method `updateUser` in `UserController`
5. Add handling of `blocked` field to `SAMLUserDetailsServiceImpl` class
6. Extend `UserManagerTest` class for all CRUD operations

mzueva:
Second part of implementation: block groups
- Create a table `group_status` for storing blocked groups (group name TEXT PK, blocked boolean deafult false)
- Add method `updateBlockStatusGroup` to `UserController` and underlying layers
- Add check to `SAMLUserDetailsServiceImpl ` that user doesn't belong to any of blocked groups

mzueva:
We need to discuss approach to JWT validation: currently we fully trust token and do not load user from DB during validation. Shall we change this approach?

mzueva:
@Wedds Third part of implementation:

- Introduce a new application property `saml.user.blocked.attribute` and add it as `@Value` to `SAMLUserDetailsServiceImpl` class
- If a user exists during SAML authentication and `saml.user.blocked.attribute`defined, exists and is set to `true`, update user status to blocked and deny access

mzueva:
@rodichenko now method **_GET role/loadAll_** returns **_blocked_** flag. If we need to add this field to any other method, please let me know.
Setting block status is available via **_POST /group/{role_name}/block_** method.

[...]

PR #633, opened 2019-09-04
Issue 623_3 Cloud Pipeline shall allow to "block" the user via properties

Wedds:
PR solves the third part of issue #623
It brings handling of specific `block` property. If a user has this property initialized with true value, his access would be denied, even if he isn't blocked in databases directly or via group status.

- New application property `saml.user.blocked.attribute` added to SAMLUserDetailsServiceImpl class
- If a user exists during SAML authentication and property is defined and is set to true, user status is updating to `blocked` and access would be denied


mzueva (changes requested):
Please, rebase onto develop, it will be easier to review

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
```diff
@@ -220,6 +222,36 @@ public void shouldThrowAuthorizationExceptionForBlockedUser() {
...
+
+    @Transactional(propagation = Propagation.REQUIRES_NEW, rollbackFor = Exception.class)
+    public void shouldThrowAuthorizationExceptionForUserWithBlockedApplicationProperty(final String blockedValue) {
+        blockUserUsingCredentials(blockedValue);
+        userDetailsService.loadUserBySAML(credential);
+    }
+
+    private void blockUserUsingCredentials(final String blockedValue) {
+        user.setUserName(USER_NAME);
+        final String SAML_ATTRIBUTES_STRING_BLOCKED_USER =
```
ekazachkova: upper case letters used for final static fields

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
```diff
@@ -71,6 +71,8 @@
     private Map<String, String> expectedAttributes = new HashMap<>();
     private List<String> groups = Stream.of(SAML_ATTRIBUTE_1, SAML_ATTRIBUTE_2).collect(Collectors.toList());
     private UserContext expectedUserContext = new UserContext(1L, USER_NAME.toUpperCase());
+    private final String SAML_ATTRIBUTE_BLOCKED_USER_KEY = "blocked";
```
ekazachkova: could we make this fields static?

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -87,25 +91,39 @@ public UserContext loadUserBySAML(SAMLCredential credential) {
...
+        return userContext;
+    }
+
+    private void throwUserIsBlocked(final String userName) {
+        LOGGER.debug("User {} is blocked!", userName);
+        throw new LockedException("User is blocked!");
+    }
+
+    private boolean hasBlockedStatusAttribute(final Map<String, String> attributes) {
+        final String blockingStatus = attributes.get(blockedKey);
```
ekazachkova: Use `SAMLCredential#getAttributeAsString` method here

ekazachkova (changes requested):

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
```diff
@@ -220,6 +220,32 @@ public void shouldThrowAuthorizationExceptionForBlockedUser() {
         userDetailsService.loadUserBySAML(credential);
     }
 
+    @Test
+    public void testBlockedValueCaseNonSensitiveness() {
```
ekazachkova: Could you please add a three separate negative test instead off one?

ekazachkova: Also, could you please add a positive tests? (if attribute does not specified/ if attribute is equal to false e.g.)

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
```diff
@@ -220,6 +220,32 @@ public void shouldThrowAuthorizationExceptionForBlockedUser() {
...
+    @Transactional(propagation = Propagation.REQUIRES_NEW, rollbackFor = Exception.class)
+    public void shouldThrowAuthorizationExceptionForUserWithBlockedApplicationProperty(final String blockedValue) {
+        blockUserUsingCredentials(blockedValue);
+        userDetailsService.loadUserBySAML(credential);
+    }
+
+    private void blockUserUsingCredentials(final String blockedValue) {
+        user.setUserName(USER_NAME);
+        Mockito.when(userManager.loadUserByName(Matchers.anyString())).thenReturn(user);
+        ReflectionTestUtils.setField(userDetailsService, "blockedAttribute", blockedValue);
```
ekazachkova: I would prefer to do not use `blockedValue` variable at this line (to my mind it is not clear). Could you use a new one "test_value" e.g?

api/src/test/resources/test-application.properties
```diff
@@ -72,6 +72,7 @@ saml.authorities.attribute.names=test_name
 saml.authn.request.binding=
 saml.sign.key=
 saml.user.attributes=email=test_email,user=test_user
+saml.user.blocked.attribute=false
```
ekazachkova: Do we need to specify this value here if we still use `ReflectionTestUtils.setField`?

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -61,6 +62,9 @@
     @Value("${saml.user.auto.create: EXPLICIT}")
     private SamlUserRegisterStrategy autoCreateUsers;
 
+    @Value("${saml.user.blocked.attribute: false}")
```
ekazachkova: Could you add this property to dev `application.properties` file and leave it empty? Also, please check that application still starts correctly.

ekazachkova (changes requested):

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
Wedds: Done

Wedds: Done

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
Wedds: Done

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
Wedds: Done

api/src/test/resources/test-application.properties
Wedds: Done

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -89,35 +93,47 @@ public UserContext loadUserBySAML(SAMLCredential credential) {
...
             throw new LockedException("User is blocked!");
         }
+        return userContext;
+    }
+
+    private boolean isValidGroupsBlockingStatus(final List<String> groups) {
+        return userManager.loadGroupBlockingStatus(groups).stream().noneMatch(GroupStatus::isBlocked);
+    }
+
+    private void throwUserIsBlocked(String userName) {
```
ekazachkova: add final please

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -62,6 +63,9 @@
     @Value("${saml.user.auto.create: EXPLICIT}")
     private SamlUserRegisterStrategy autoCreateUsers;
 
+    @Value("${saml.user.blocked.attribute: false}")
```
ekazachkova: Could you please remove `false` from defaults, just leave it empty

ekazachkova (changes requested):

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
Wedds: Done

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
Wedds: Done

ekazachkova (approved):

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -79,7 +83,7 @@ public UserContext loadUserBySAML(SAMLCredential credential) {
         String userName = credential.getNameID().getValue().toUpperCase();
         List<String> groups = readAuthorities(credential);
         Map<String, String> attributes = readAttributes(credential);
-
+        final UserContext userContext;
```
mzueva: It is a good time to clean up this code a bit.
I suggest to use `Optional` and replace `if` in the following manner:
```
Optional.ofNullable(userManager.loadUserByName(userName))
.map(user -> processRegisteredUser(user, ...))
.orElseGet(() -> processNewUser(...))
```

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -89,35 +93,47 @@ public UserContext loadUserBySAML(SAMLCredential credential) {
             PipelineUser createdUser = userManager.createUser(userName,
                     roles, groups, attributes, null);
             LOGGER.debug("Created user {} with groups {}", userName, groups);
-            validateGroupsBlockingStatus(groups, userName);
-            UserContext userContext = new UserContext(createdUser.getId(), userName);
+            userContext = new UserContext(createdUser.getId(), userName);
```
mzueva: We need to apply blocked group and attribute check for a new user as well.

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -52,6 +52,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SAMLUserDetailsServiceImpl.class);
     private static final String ATTRIBUTES_DELIMITER = "=";
+    private static final String ATTR_USER_BLOCKED_VALUE = "true";
```
mzueva: move this value to application.properties

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
```diff
@@ -241,13 +245,58 @@ public void shouldThrowAuthorizationExceptionForBlockedUser() {
...
+    @Test
+    @Transactional(propagation = Propagation.REQUIRES_NEW, rollbackFor = Exception.class)
+    public void shouldAcceptUserWithValidNotBlockedApplicationProperty() {
+        blockUserUsingCredentials(SAML_ATTRIBUTE_NOT_BLOCKED_USER_VALUE);
+        final UserContext actualUserContext = userDetailsService.loadUserBySAML(credential);
+        Assert.assertEquals(expectedUserContext.getUsername(), actualUserContext.getUsername());
+    }
+
+    @Test
+    @Transactional(propagation = Propagation.REQUIRES_NEW, rollbackFor = Exception.class)
```
mzueva: You can add `@Transactional` to test class instead of annotation each test individally

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
```diff
@@ -241,13 +245,58 @@ public void shouldThrowAuthorizationExceptionForBlockedUser() {
...
     @Test(expected = LockedException.class)
     @Transactional(propagation = Propagation.REQUIRES_NEW, rollbackFor = Exception.class)
     public void shouldThrowAuthorizationExceptionForUserFromBlockedGroup() {
         blockOneGroupForCurrentUser();
         userDetailsService.loadUserBySAML(credential);
     }
 
+    private void blockUserUsingCredentials(final String blockedValue) {
+        user.setUserName(USER_NAME);
+        Mockito.when(userManager.loadUserByName(Matchers.anyString())).thenReturn(user);
```
mzueva: maybe use `eq` matcher instead of `anyString`

mzueva (changes requested):

api/src/test/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImplTest.java
mzueva: @ekazachkova I'm not sure what change was required here?

ekazachkova: Sorry, my comment is really confusing.
The idea was just to make test code a little close to real life. It is possible to set `SAMLUserDetailsServiceImpl#blockedAttribute` as "TRUE", but in fact SAML attribute name will be more specific (attribute name and attribute value would not be the same). So it would be more clearly to create a new test constant (but mocked `credential#getAttributeAsString` method should still return boolean value).

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -77,39 +84,54 @@
...
+            throwUserIsBlocked(userName);
+        }
+        return userContext;
+    }
+
+    private UserContext processNewUser(final String userName, final List<String> groups,
+                                       final Map<String, String> attributes) {
+        final String message = messageHelper.getMessage(MessageConstants.ERROR_USER_NAME_NOT_FOUND, userName);
+        checkAbilityToCreate(userName, groups);
+        LOGGER.debug(message);
```
mzueva: I think we can remove local variable `message` and use
```
LOGGER.debug(messageHelper.getMessage(MessageConstants.ERROR_USER_NAME_NOT_FOUND, userName));
```

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -77,39 +84,54 @@
...
+        final List<Long> roles = roleManager.getDefaultRolesIds();
+        final PipelineUser createdUser = userManager.createUser(userName,
+                                                                roles, groups, attributes, null);
+        LOGGER.debug("Created user {} with groups {}", userName, groups);
+        final UserContext userContext = new UserContext(createdUser.getId(), userName);
+        userContext.setGroups(createdUser.getGroups());
+        userContext.setRoles(createdUser.getRoles());
+        validateGroupsBlockingStatus(createdUser.getAuthorities(), userName);
```
mzueva: Since this method is executed for both: new and existing users, we can move it common code in `loadUserBySAML` method. Use `UserContext.getAuthorities` method for filtering. Good point - in this class authorities doesn't contain username, so we don't need to filter it out in `validateGroupsBlockingStatus` method.

api/src/main/java/com/epam/pipeline/security/saml/SAMLUserDetailsServiceImpl.java
```diff
@@ -77,39 +84,54 @@
...
+        return userContext;
+    }
+
+    private UserContext processRegisteredUser(final String userName, final List<String> groups,
+                                              final Map<String, String> attributes, final PipelineUser loadedUser) {
+        final UserContext userContext;
```
mzueva: After moving `validateGroupsBlockingStatus` to common method, remove declaration of `userContext ` and just return value from each of `if` branches. It is better not to split variable declaration and assignment.

mzueva (changes requested):
@Wedds Please, check `SystemPreferencesValidationTest` it seems to be broken after rebase

mzueva (approved):

Issue #607, opened 2019-08-15
`pipe`: excess output for the `run` command execution


NShaforostov:
### 1. Excess hint after each `run` command execution

See the hint started from "Docker image, instance type, instance disk..."

Example 1:
``` bash
$ pipe run -p -n simplepipeline
Fetching pipeline info...done.
"simplepipeline@draft-877ec67e" pipeline arguments:
No parameters are configured
Docker image, instance type, instance disk and cmd template are required parameters if pipeline was not provided.
$
```

Example 2:
``` bash
$ pipe run -pt spot -n simplepipeline
Fetching pipeline info...done.
Evaluating estimated price...done.

Price per hour (Standard_D2s_v3, hdd 20)        0.03 $

Are you sure you want to schedule a run of simplepipeline? [y/N]: y
"simplepipeline@draft-877ec67e" pipeline run scheduled with RunId: 6993
Docker image, instance type, instance disk and cmd template are required parameters if pipeline was not provided.
$
```

In "quiet mode" excess hint also appears.
Example 3:
``` bash
$ pipe run -n simplepipeline -q
Are you sure you want to schedule a run of simplepipeline? [y/N]: y
3857
docker_image, instance_type, instance_disk, cmd_template
$
```

Could we output such message only in cases when a pipeline name/ID wasn't specified?

### 2. Empty error message in case of the launch cancel

When you try to launch a pipeline without the `-y` option and on the confirm question you answer `N` to cancel the launch, an empty error message appears:

``` bash
$ pipe run -n 6383 -pt spot
Fetching pipeline info...done.
Evaluating estimated price...done.

Price per hour (Standard_D2s_v3, hdd 20)        0.03 $

Are you sure you want to schedule a run of 6383? [y/N]: N
Error:
$
```

In "quiet mode" that empty error message doesn't display.

sidoruka:
Moved to `release/0.15` via 83e6b3e85dc677b8cff52f299191e20d5479fbbf

NShaforostov:
Verified

PR #662, opened 2019-09-16
`pipe`: excess output for the `run` command execution

ekazachkova:
Fix for issue #607 
There were a 2 problems:
1. incorrect error message
2. empty error message after abort operation


NShaforostov:
Verified.
Works fine

SilinPavel (approved):

Issue #609, opened 2019-08-15
`pipe`: unclear option `-nc` for the `run` command


NShaforostov:
When trying to launch a pipeline with single `-nc` option an error throws:
``` bash
$ pipe run -pt spot -n simplepipeline -nc 6
Traceback (most recent call last):
  File "pipe-cli/pipe.py", line 1030, in <module>
  File "site-packages/click/core.py", line 722, in __call__
  File "site-packages/click/core.py", line 697, in main
  File "site-packages/click/core.py", line 1066, in invoke
  File "site-packages/click/core.py", line 895, in invoke
  File "site-packages/click/core.py", line 535, in invoke
  File "site-packages/click/decorators.py", line 17, in new_func
  File "pipe-cli/src/config.py", line 119, in validate_access_token_wrapper
  File "site-packages/click/core.py", line 535, in invoke
  File "pipe-cli/pipe.py", line 639, in run
  File "pipe-cli/src/utilities/pipeline_run_operations.py", line 74, in run
  File "pipe-cli/src/utilities/cluster_manager.py", line 30, in calculate_cluster_from_cores
  File "pipe-cli/src/utilities/cluster_manager.py", line 64, in get_instance_type
RuntimeError: No instances found for type c4
[3089] Failed to execute script pipe
$
```

Could we make an error message more informative and understandable for users?

When trying to launch a pipeline with the both `-nc` and `-ic` options a cluster starts normally but a specified number of cores doesn't affect anything.

And what could be the right usage of the `-nc` option?

sidoruka:
Moved to `release/0.15` via 5f23ffcf18c017f6fe3cfba79a34856994314822

NShaforostov:
Verified

PR #664, opened 2019-09-17
pipe: unclear option -nc for the run command

ekazachkova:
This PR provides fix for issue #609 

This PR contains the following changes:
- help message updated for using `-ic` and `-nc` commands simultaneously
- attempt to load default instance type from `cluster.instance.type` preference added
- default core type changed from `c4` to `m5`


NShaforostov:
Verified.
Works fine

sidoruka (approved):

Issue #666, opened 2019-09-17
Introduce tags for PipelineRun


SilinPavel:
This issue propose to add tags for PipelineRun to be able to solve #624 

- Tag should have nave and value
- Tags should be returned along with pipeline run

SilinPavel:
@Wedds To solve this issue let's implement the next steps:

1.  Introduce new **jsonb** column for tags in pipeline run table
2.  Introduce new field Map<String, String> tags for PipelineRun
3.  Using Postgre jsonb type functionality implement CRUD methods for tags (change methods fro loading run, add method for updating tags for specific run)
4. Implement parsing of tags to Map<String, String>, for examples see work with **metadata** in MetadataDao

SilinPavel:
checked, works fine.

PR #669, opened 2019-09-17
Issue 666 Implementation of tags for PipelineRun

Wedds:
PR solves issue #666 
It brings handling of tags for PipelineRun entity.

- Add column `tags` (jsonb) to `pipeline_run` table
- Add tags field (Map<String, String>) for PipelineRun
- Add functionality for proceeding tags to PipelineRunDao and PipelineRunManager

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
@@ -22,6 +22,7 @@ import com.epam.pipeline.controller.PagedResult;
 import com.epam.pipeline.controller.vo.PagingRunFilterVO;
 import com.epam.pipeline.controller.vo.PipelineRunFilterVO;
 import com.epam.pipeline.controller.vo.PipelineRunServiceUrlVO;
+import com.epam.pipeline.controller.vo.TagsVO;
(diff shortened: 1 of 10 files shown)


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -840,6 +840,20 @@ public PipelineRun updateRunSids(Long runId, List<RunSid> runSids) {
         return pipelineRun;
     }
 
+    /**
+     * Updates run's tags
+     * @param run {@link PipelineRun} for pipeline run which tagsshould be updated
+     * @return
+     */
+    @Transactional(propagation = Propagation.REQUIRED)
+    public PipelineRun updateTags(final PipelineRun run) {
```
SilinPavel: lets also create api endpoint for `POST /run/{runId}/tag` to be able to update run tags by api

SilinPavel: So we will need to change signature of the updateTags method, it should get run id and some TagsVO with maps of tags inside

SilinPavel (changes requested):

SilinPavel (approved):

Issue #624, opened 2019-08-23
Highlight "Idle" jobs in the Cloud Pipeline API/GUI/CLI


sidoruka:
1. Cloud Pipeline shall allow to view high-level metrics information for the Active Runs:
- Idle (CPU/Memory consumption is lower than a certain level, defined by the admin) - such runs produce extra costs
- Pressure (CPU/Memory/Disk consumption is higher than a certain level, defined by the admin) - such runs may accidentally fail
2. We shall reuse existing resources monitoring indices and the thresholds that are used for auto-pausing and email notifications
3. Users shall be able to drill-down into the resource consumption metrics (e.g. click "Idle" run and review the resource consumption for the whole job lifetime)
4. Users will be able to sort/filter for idle or pressured runs within a list of active/completed runs
5. The "Idle" or "Pressure" info shall be available for each of the jobs running via:
* GUI
* CLI
* API

SilinPavel:
@Wedds 
First step:
1. Introduce new field Map<String, String> tags for PipelineRun
2. Using Postgre jsonb type functionality change SQL queries to be able to join table metadata for corresponding entity_id and entity_class and select tags (you should select only "tags" field from data column)
3. Change RowMapper from PipelineRunDao to be able to parse and set tags from pipelineRun object

SilinPavel:
@Wedds 
Lets implement the next steps:
1) In the ResourceMonitoringManager change `processRuns` method to set new tag `IDLED=true` or delete previous value if needed to pipelineRun
2) In the ResourceMonitoringManager change `processOverloadedRuns` method to set to runs new tag `PRESSURED=true` or delete previous if needed and collect this runs to list
3) Implement method `updateTags(List<PipelineRun> runs)` for `PipelineRunManager` that will be update tags in DB by batch query (see as example method `updatePipelineRunsLastNotification`)
4) with this method update tags for collected runs in 1) and 2)

NShaforostov:
I've tried to verify this feature.
The following preferences were set:
- `system.max.idle.timeout.minutes` `1`
- `system.idle.cpu.threshold` `50`

In total, six runs were launched at each stand:
- empty DEFAULT pipeline with enabled the `Start idle` checkbox (**On-Demand** with enabled the `Auto pause` checkbox, **On-Demand** with disabled the `Auto pause` checkbox, **Spot**)
[...]

PR #674, opened 2019-09-20
Issue 624 Highlight "Idle" and "Pressured" jobs in the Cloud Pipeline API/GUI/CLI

Wedds:
PR solves the issue #624.

It brings tagging of PipelineRuns with low- or high-resource consumption.
ResourceMonitorManager add "IDLED" or "PRESSURED" tag for a run entities, when it's CPU/Memory/Disk consumption is lower/higher than a certain level, defined by the admin, and removes such tags, when resource consumption going back to normal level.
- Added changes to inner ResourceMonitorManager logic;
- Added method `updateRunsTags` to PipelineRunManager to use batch updating of runs tags;
- Tests for new functionality were provided as well.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.PostConstruct;
@@ -72,6 +73,9 @@ public class ResourceMonitoringManager extends AbstractSchedulingManager {
     private static final int MILLIS = 1000;
     private static final double PERCENT = 100.0;
     private static final double ONE_THOUSANDTH = 0.001;
+    private static final String UTILIZATION_LEVEL_LOW = "IDLED";
+    private static final String UTILIZATION_LEVEL_HIGH = "PRESSURED";
+    private static final String TRUE_VALUE_STRING = "true";
 
     private final PipelineRunManager pipelineRunManager;
     private final NotificationManager notificationManager;
@@ -118,11 +122,12 @@ public class ResourceMonitoringManager extends AbstractSchedulingManager {
     public void monitorResourceUsage() {
         List<PipelineRun> runs = pipelineRunManager.loadRunningPipelineRuns();
 
-        processIdleRuns(runs);
-        processOverloadedRuns(runs);
+        final List<PipelineRun> runsToUpdateTags = processIdleRuns(runs);
+        runsToUpdateTags.addAll(processOverloadedRuns(runs));
+        pipelineRunManager.updateRunsTags(runsToUpdateTags);
     }
 
-    private void processOverloadedRuns(final List<PipelineRun> runs) {
+    private List<PipelineRun> processOverloadedRuns(final List<PipelineRun> runs) {
         final Map<String, PipelineRun> running = runs.stream()
                 .filter(r -> {
                     final boolean hasNodeName = Objects.nonNull(r.getInstance())
@@ -156,7 +161,48 @@ public class ResourceMonitoringManager extends AbstractSchedulingManager {
                 .filter(pod -> isPodUnderPressure(pod.getValue(), thresholds))
                 .collect(Collectors.toList());
 
+        final List<PipelineRun> runsToUpdateTags = getRunsToUpdatePressuredTags(running, runsToNotify);
         notificationManager.notifyHighResourceConsumingRuns(runsToNotify, NotificationType.HIGH_CONSUMED_RESOURCES);
+        return runsToUpdateTags;
+    }
+
+    private List<PipelineRun> getRunsToUpdatePressuredTags(final Map<String, PipelineRun> running,
+                              final List<Pair<PipelineRun, Map<ELKUsageMetric, Double>>> runsToNotify) {
+        final List<PipelineRun> listOfRunsToNotify = runsToNotify
+                .parallelStream()
+                .map(Pair::getLeft)
+                .collect(Collectors.toList());
+        final Predicate<PipelineRun> runToBeNotified = listOfRunsToNotify::contains;
+        final Stream<PipelineRun> runsToAddTag = listOfRunsToNotify
+                .parallelStream()
+                .filter(this::hasNoPressuredTag)
+                .map(this::addPressuredTagToRun);
+        final Stream<PipelineRun> runsToRemoveTag = running.values()
+                .parallelStream()
+                .filter(runToBeNotified.negate())
+                .filter(this::hasPressuredTag)
+                .map(this::removePressuredTagFromRun);
+        return Stream.concat(runsToAddTag, runsToRemoveTag).collect(Collectors.toList());
+    }
+
+    private boolean hasPressuredTag(final PipelineRun run) {
+        final String tagToBeRemoved = run.getTags().get(UTILIZATION_LEVEL_HIGH);
+        return tagToBeRemoved != null
+                && tagToBeRemoved.equals(TRUE_VALUE_STRING);
+    }
+
+    private boolean hasNoPressuredTag(final PipelineRun run) {
+        return !hasPressuredTag(run);
+    }
+
+    private PipelineRun removePressuredTagFromRun(final PipelineRun run) {
+        run.getTags().remove(UTILIZATION_LEVEL_HIGH);
+        return run;
+    }
+
+    private PipelineRun addPressuredTagToRun(final PipelineRun run) {
+        run.getTags().put(UTILIZATION_LEVEL_HIGH, TRUE_VALUE_STRING);
+        return run;
     }
 
     private Pair<PipelineRun, Map<ELKUsageMetric, Double>> matchRunAndMetrics(
@@ -193,7 +239,7 @@ public class ResourceMonitoringManager extends AbstractSchedulingManager {
         return result;
     }
 
-    private void processIdleRuns(final List<PipelineRun> runs) {
+    private List<PipelineRun> processIdleRuns(final List<PipelineRun> runs) {
         final Map<String, PipelineRun> running = runs.stream()
                 .collect(Collectors.toMap(PipelineRun::getPodId, r -> r));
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -243,13 +291,20 @@ private void processIdleRuns(final List<PipelineRun> runs) {
                 double cpuUsageRate = metric / MILLIS / type.getVCPU();
                 if (Precision.compareTo(cpuUsageRate, idleCpuLevel, ONE_THOUSANDTH) < 0) {
                     processIdleRun(run, actionTimeout, action, runsToNotify, runsToUpdate, cpuUsageRate);
+                    if (run.getTags().put(UTILIZATION_LEVEL_LOW, TRUE_VALUE_STRING) == null) {
```
SilinPavel: you can use computeIfAbsent here

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -243,13 +291,20 @@ private void processIdleRuns(final List<PipelineRun> runs) {
...
+                        runsToUpdateTags.add(run);
+                    }
                 } else if (run.getLastIdleNotificationTime() != null) { // No action is longer needed, clear timeout
                     run.setLastIdleNotificationTime(null);
+                    final String tagToBeRemoved = run.getTags().remove(UTILIZATION_LEVEL_LOW);
+                    if (tagToBeRemoved != null
```
SilinPavel: take a look at method remove(key, value). I believe it could be suitable here as well

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -118,11 +122,12 @@ public void removeOldIndices() {
     public void monitorResourceUsage() {
         List<PipelineRun> runs = pipelineRunManager.loadRunningPipelineRuns();
 
-        processIdleRuns(runs);
-        processOverloadedRuns(runs);
+        final List<PipelineRun> runsToUpdateTags = processIdleRuns(runs);
```
SilinPavel: lets updates tags for overloaded and idled runs separately, to increase code readability

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +161,48 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
+                .map(this::addPressuredTagToRun);
+        final Stream<PipelineRun> runsToRemoveTag = running.values()
+                .parallelStream()
+                .filter(runToBeNotified.negate())
+                .filter(this::hasPressuredTag)
+                .map(this::removePressuredTagFromRun);
+        return Stream.concat(runsToAddTag, runsToRemoveTag).collect(Collectors.toList());
+    }
+
+    private boolean hasPressuredTag(final PipelineRun run) {
```
SilinPavel: try to get rid of such small methods, I believe that in stream context it will obvious without methods. f.e.:

final Stream<PipelineRun> runsToRemoveTag = running.values()
          .parallelStream()
          **.filter(run -> listOfRunsToNotify.contains(run))
          .filter(run -> run.getTags().contains(UTILIZATION_LEVEL_HIGH))**

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +161,48 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
                 .collect(Collectors.toList());
 
+        final List<PipelineRun> runsToUpdateTags = getRunsToUpdatePressuredTags(running, runsToNotify);
         notificationManager.notifyHighResourceConsumingRuns(runsToNotify, NotificationType.HIGH_CONSUMED_RESOURCES);
+        return runsToUpdateTags;
+    }
+
+    private List<PipelineRun> getRunsToUpdatePressuredTags(final Map<String, PipelineRun> running,
+                              final List<Pair<PipelineRun, Map<ELKUsageMetric, Double>>> runsToNotify) {
+        final List<PipelineRun> listOfRunsToNotify = runsToNotify
```
SilinPavel: think about changing this to `Set<Long> of runIds` it would be much faster

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -221,16 +267,18 @@ private void processIdleRuns(final List<PipelineRun> runs) {
         final IdleRunAction action = IdleRunAction.valueOf(preferenceManager
                 .getPreference(SystemPreferences.SYSTEM_IDLE_ACTION));
 
+        final List<PipelineRun> runsToUpdateTags = new ArrayList<>(running.size());
```
SilinPavel: move updating tags and sending notifications to `processRuns` method, it would be more readable and also logical

SilinPavel (changes requested):

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
SilinPavel: also lets think that if `run.getTags().contains(UTILIZATION_LEVEL_HIGH)` the value is `true` because we don't have other option

core/src/main/java/com/epam/pipeline/entity/pipeline/PipelineRun.java
```diff
@@ -173,4 +173,49 @@ public void parseParameters() {
...
+    }
+
+    /**
+     * If tag isn't set for run, add new one, or remove it otherwise.
+     * @param key tag key to be proceeded
+     * @param value tag value to be proceeded
+     * @return run with updated tags
+     */
+    public PipelineRun inverseTagForRun(final String key, final String value) {
+        if (hasTag(key)) {
```
SilinPavel: should it be inverted?

SilinPavel: if so, and tests are pass it, please correct a mistake and add test case for that it ResourceMonitoringManagerTest

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +160,26 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
                 .collect(Collectors.toList());
 
+        final List<PipelineRun> runsToUpdateTags = getRunsToUpdatePressuredTags(running, runsToNotify);
         notificationManager.notifyHighResourceConsumingRuns(runsToNotify, NotificationType.HIGH_CONSUMED_RESOURCES);
+        pipelineRunManager.updateRunsTags(runsToUpdateTags);
+    }
+
+    private List<PipelineRun> getRunsToUpdatePressuredTags(final Map<String, PipelineRun> running,
+                              final List<Pair<PipelineRun, Map<ELKUsageMetric, Double>>> runsToNotify) {
+        final Set<Long> listOfRunsIdToNotify = runsToNotify
```
SilinPavel: lets call variable reasonable short
I mean that in this case we can't get rid of `runsId` or `Notify` in a naming, because it's important, but `listOf` can be deleted, it would let to code be more readable and short

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +160,26 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
+                .parallelStream()
```
SilinPavel: lets do it without parallel streams, since this code already starts in separate executor

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +160,26 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
+                .map(p -> p.getLeft().getId())
+                .collect(Collectors.toSet());
+        final Stream<PipelineRun> runsStream = running.values().parallelStream();
+        final Predicate<PipelineRun> isToBeNotified = r -> listOfRunsIdToNotify.contains(r.getId());
+        final Stream<PipelineRun> runsToAddTag = runsStream
```
SilinPavel: actually you can't do mapping twice, because you can't go through stream twice, so get rid of this local variable. And it certainly should be covered by tests

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +160,26 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
+                .filter(isToBeNotified)
+                .map(r -> r.inverseTagForRun(UTILIZATION_LEVEL_HIGH, TRUE_VALUE_STRING));
```
SilinPavel: there is something wrong with logic.
we should check that if pipeline will be notified and doesn't have tag we should add it
and if it will not be notified and tag is present - we should delete it

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -243,15 +264,21 @@ private void processIdleRuns(final List<PipelineRun> runs) {
                 double cpuUsageRate = metric / MILLIS / type.getVCPU();
                 if (Precision.compareTo(cpuUsageRate, idleCpuLevel, ONE_THOUSANDTH) < 0) {
                     processIdleRun(run, actionTimeout, action, runsToNotify, runsToUpdate, cpuUsageRate);
+                    if (run.addTag(UTILIZATION_LEVEL_LOW, TRUE_VALUE_STRING)) {
+                        runsToUpdateTags.add(run);
```
SilinPavel: actually can we use runsToUpdate to update tags as well? Am I right that it will contain all runs with updated tags? and we can just do:
```
if (Precision.compareTo(cpuUsageRate, idleCpuLevel, ONE_THOUSANDTH) < 0) {
    processIdleRun(run, actionTimeout, action, runsToNotify, runsToUpdate, cpuUsageRate);
    run.addTag(UTILIZATION_LEVEL_LOW, TRUE_VALUE_STRING)
} else if (run.getLastIdleNotificationTime() != null) {
    run.setLastIdleNotificationTime(null);
    run.removeTag(UTILIZATION_LEVEL_LOW, TRUE_VALUE_STRING)
    runsToUpdate.add(run);
}
...
pipelineRunManager.updatePipelineRunsLastNotification(runsToUpdate);
pipelineRunManager.updateRunsTags(runsToUpdate);
```

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -156,7 +160,29 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
+        pipelineRunManager.updateRunsTags(runsToUpdateTags);
+    }
+
+    private List<PipelineRun> getRunsToUpdatePressuredTags(final Map<String, PipelineRun> running,
+                              final List<Pair<PipelineRun, Map<ELKUsageMetric, Double>>> runsToNotify) {
+        final Set<Long> runsIdToNotify = runsToNotify
+                .stream()
+                .map(p -> p.getLeft().getId())
+                .collect(Collectors.toSet());
+        final Predicate<PipelineRun> isToBeNotified = r -> runsIdToNotify.contains(r.getId());
```
SilinPavel: also do we need this? I believe that `runsIdToNotify.contains` also understandable

SilinPavel (approved):

Issue #645, opened 2019-09-09
Rename GitLab's repository if the name of the Pipeline is changed


sidoruka:
At the moment, if the `Pipeline` is renamed - the underlying gitlab repository keeps the previous name.

We shall:
* Research is it feasible to rename the gitlab repo (without the side-effects)
* Research if the current mapping of the Cloud Pipeline permissions to the GitLab roles - allows renaming the repositories
* If so - implement this feature: 
  * Rename the GitLab repo if the Cloud Pipeline `pipeline` is renamed
  * GUI shall display a notification, when such operation is performed: `This operation will change the clone/pull/push URL. If this pipeline was used somewhere - make sure to change the remote address`

mzueva:
@Wedds Please, check whether Gitlab CE has API to rename repository/project and if it works correctly on our version.

mzueva:
@Wedds `PipelineManager.update` method shall check if name of pipeline has actually changed and rename underlying gitlab repository as well. Notes:

- Method to rename gitlab repository shall be added to `GitManager` and `GitClient` classes.
- Do not forget to update `Pipeline` fields `repository` and `repositorySsh` fields after renaming of repository
- If process of repository renaming fails, all db changes shall be reverted
- Use `issue_312_gitlab_changes_on_behalf_of_user` as a base branch as it introduces significant changes in gitlab integration code

mzueva:
@sidoruka deployed to dev

sidoruka:
Works great
@NShaforostov please proceed with the docs update

NShaforostov:
Verified.

> Works great
> @NShaforostov please proceed with the docs update

Docs updated via #807

PR #692, opened 2019-09-27
Issue 645 Rename GitLab's repository if the name of the Pipeline is changed.

Wedds:
PR solves the issue #645 
It performs pipeline repository (GitLab project) rename if pipeline name is changed.

- Method to rename GitLab repository is added to `GitManager` and `GitClient` classes.
- Pipeline fields `repository` and `repositorySsh` are updating with the renaming of a repository.
- If the process of repository renaming fails, all DB changes are reverted.
- Tests are added and common methods are extracted to util class.

image
image

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/git/GitlabClient.java
+++ b/api/src/main/java/com/epam/pipeline/manager/git/GitlabClient.java
@@ -68,6 +68,8 @@ import java.util.Optional;
 import java.util.TimeZone;
 import java.util.stream.Stream;
 
+import static com.epam.pipeline.utils.GitUtils.convertPipeNameToProject;
+
 @Wither
 @AllArgsConstructor
 @NoArgsConstructor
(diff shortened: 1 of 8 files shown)


api/src/main/java/com/epam/pipeline/manager/git/GitlabClient.java
```diff
@@ -331,6 +326,15 @@ public GitRepositoryEntry createProjectHook(String hookUrl) throws GitClientExce
         return addProjectHook(projectId, hookUrl);
     }
 
+    public GitProject updateProjectName(final String currentName, final String newName) throws GitClientException{
+        final String normalizedNewName = convertPipeNameToProject(newName);
```
SilinPavel: use explicit call here `GitUtils.convertPipeNameToProject` instead of static import

api/src/main/java/com/epam/pipeline/manager/git/GitlabClient.java
```diff
@@ -331,6 +326,15 @@ public GitRepositoryEntry createProjectHook(String hookUrl) throws GitClientExce
         return addProjectHook(projectId, hookUrl);
     }
 
+    public GitProject updateProjectName(final String currentName, final String newName) throws GitClientException{
```
SilinPavel: miss a space `GitClientException{`

api/src/main/java/com/epam/pipeline/utils/GitUtils.java
```diff
@@ -23,6 +23,7 @@
 
     //regex for alphanumeric characters, underscore, dots and dash, allowing dash only in the middle
     private static final Pattern NAME_PATTERN = Pattern.compile("^[a-zA-Z0-9_.]+([-.][a-zA-Z0-9_]+)*$");
+    private static final String PROJECT_NAME_IN_URL_PATTERN = "(?<=/)[\\w]+(?=\\.git$)";
```
SilinPavel: since we use this pattern for replacing a name, we won't we be able to use `.` in name of repo. So I believe that we should also leave a comment here that we don't count on dots here because of our rule of naming from method `convertPipeNameToProject` or change a pattern to  `(?<=/)[\\w\\.]+(?=\\.git$)`

SilinPavel (changes requested):

SilinPavel (approved):

PR #693, opened 2019-09-30
Issue #623: Cloud Pipeline shall allow to "block" the user

ekazachkova:
Implements issue #623 for blocking user roles

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/user/RoleDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/user/RoleDao.java
@@ -176,9 +176,7 @@ public class RoleDao extends NamedParameterJdbcDaoSupport {
             }
             if (getGroupStatus) {
                 final boolean groupStatus = rs.getBoolean(GROUP_BLOCKED.name());
-                if (!rs.wasNull()) {
-                    role.setBlocked(groupStatus);
-                }
+                role.setBlocked(!rs.wasNull() && groupStatus);
             }
             return role;
         }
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/epam/pipeline/dao/user/RoleDao.java
```diff
@@ -173,18 +174,24 @@ static Role parseRole(ResultSet rs, Role role) throws SQLException {
             if (!rs.wasNull()) {
                 role.setDefaultStorageId(defaultStorageId);
             }
+            if (getGroupStatus) {
+                final boolean groupStatus = rs.getBoolean(GROUP_BLOCKED.name());
+                if (!rs.wasNull()) {
```
mzueva: Maybe set `blocked` status to `false` if it is not present?

mzueva (changes requested):

core/src/main/java/com/epam/pipeline/entity/user/Role.java
```diff
@@ -34,6 +34,7 @@
     private boolean predefined;
     private boolean userDefault;
     private Long defaultStorageId;
+    private Boolean blocked;
```
mzueva, line 37: please add default `false` parameter value to default constructor

api/src/main/java/com/epam/pipeline/entity/user/DefaultRoles.java
```diff
@@ -26,8 +26,8 @@
 @AllArgsConstructor
 public enum DefaultRoles {
 
-    ROLE_ADMIN(new Role(1L, "ROLE_ADMIN", true, false, null)),
-    ROLE_USER(new Role(2L, "ROLE_USER", true, true, null));
+    ROLE_ADMIN(new Role(1L, "ROLE_ADMIN", true, false, null, null)),
```
mzueva, line 29: add `false` value for `blocked` param

mzueva (changes requested):

mzueva (approved):

Issue #697, opened 2019-10-01
"pipe" shall consider current API version


sidoruka:
**Background**
At the moment, Cloud Pipeline CLI does not consider the API server version. It would be nice to show the server's version to the user and consider it when performing the update

Previously, there was no option to get the API server version, but this is going to be introduced by #696

**Approach**
- Before performing the actual download/update of the `pipe cli` version (introduced by #589 #615) - we shall check if the update is required. If a current version is greater or equal to the server's - notify the user and do not perform anything. Version check shall NOT consider the commit hash, only `major/minor/build` parts
- `--version` shall display the server's version (if `pipe cli` is configured, otherwise - skip this step).

NShaforostov:
I've tried to verify this feature.
I got the "old" version 0.16.0.2938.b39e1318156277bf71fa30a84ca98506f20522ce and tried to update to version 0.16.0.3090.6ed66ee576c8fc0496b7b4f3a9c8f36dfb06606d:
- In Windows the `pipe update` command worked well.
- In Linux (WSL Ubuntu 18.04):
  - trying to perform the command `pipe update` (call by the command name) from the `home` directory threw the error `Error: [Errno 21] Is a directory: '/home/<USER_NAME>/pipe'`. **Update didn't perform**
  - trying to perform the command `pipe update` (call by the command name) from any other directory threw the error `Error: [Errno 2] No such file or directory: '<PATH>/pipe'`. **Update didn't perform**
  -  trying to perform the command `pipe update` from the `pipe` directory worked well. Update has performed

Also in both - Windows and Linux (WSL Ubuntu 18.04):
- displaying of the server's version via `--version` works fine
- checking of the update required works fine

sidoruka:
@mzueva @ekazachkova please take care of this, once you'll have time

mzueva:
@NShaforostov fixed via #776

NShaforostov:
Verified. Works fine

NShaforostov:
Added to docs via #807

PR #705, opened 2019-10-02
"pipe" shall consider current API version

ekazachkova:
Implementation for issue #697 
Provides the following changes:
- for `--version` operation: additional info message with current Cloud Pipeline API version was added 
- for `update` operation: do not update Cloud Pipeline CLI if it is already up-to-date

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -17,6 +17,7 @@ import requests
 import sys
 from prettytable import prettytable
 
+from src.api.app_info import ApplicationInfo
 from src.api.cluster import Cluster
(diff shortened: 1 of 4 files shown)


pipe-cli/pipe.py
```diff
@@ -38,9 +39,19 @@
 MAX_CORES_COUNT = 10000
 
 
+def silent_print_api_version():
+    try:
+        api_info = ApplicationInfo().info()
+    except ConfigNotFoundError:
+        return
+    if 'version' in api_info and api_info['version']:
```
mzueva: Let's move it into try block

mzueva (approved):

Issue #168, opened 2019-04-18
Switch cAdvisor endpoint to the elastic index for the node resources visualization


sidoruka:
As we now have the resource info in the elastic index (which is used for the auto-pause feature) - it can be used as an alternative to the direct cAdisor request to display resources utilization for a longer period of time.

The following behavior will be good to implement:
1. Introduce a new `application.properties` configuration item (enum: cadvisor/elastic)
2. If (1) is set to `elastic` and elastic search instance connectivity is configured for the API server - it shall be used as a source for the resource utilization info
3. Otherwise (either (1) is set to `cadvisor` or elastic is not configured) - existing approach shall be used

PR #712, opened 2019-10-02
Fix elastic monitoring backend datetime issues

tcibinan:
Relates to issue #168 and fixes changes merged in #306.

Pull request brings several fixes:
- Use node instance `creationTimestamp` rather then `createdDate` as a default usage monitoring period start.
- Configure elastic to not fail on unavailable indices.
- Trim monitoring stats to suit requested period.
- Unify `cadvisor` and `elastic` monitoring backends date time formatting.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
@@ -183,12 +183,8 @@ public class NodesManager {
                 return Optional.of(nodeInstance);
             }
         } catch (KubernetesClientException e) {
-            return missingNodeInstance(name, request);
+            log.warn("Node wasn't found in cluster.", e);
         }
-        return missingNodeInstance(name, request);
-    }
-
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
```diff
@@ -181,7 +182,13 @@ public NodeInstance getNode(String name, FilterPodsRequest request) {
                 );
                 return Optional.of(nodeInstance);
             }
+        } catch (KubernetesClientException e) {
+            return missingNodeInstance(name, request);
```
mzueva: Maybe just log exception and use `return` outside of catch clause.

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
tcibinan: Done.

mzueva (approved):

PR #748, opened 2019-10-17
Fix issue 738

SilinPavel:
This PR related to issue #738, when sometimes ESMonitoringManager can't load info about disk usage for the last minute

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/AbstractMetricRequester.java
+++ b/api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/AbstractMetricRequester.java
@@ -202,30 +202,16 @@ public abstract class AbstractMetricRequester implements MetricRequester, Monito
         return parseStatsResponse(executeRequest(request));
     }
 
-    protected SearchSourceBuilder nodeStatsQuery(final String nodeName, final LocalDateTime from,
-                                                 final LocalDateTime to) {
+    protected SearchSourceBuilder statsQuery(final String nodeName, final String type,
+                                             final LocalDateTime from, final LocalDateTime to) {
         return new SearchSourceBuilder()
                 .query(QueryBuilders.boolQuery()
                         .filter(QueryBuilders.termsQuery(path(FIELD_METRICS_TAGS, FIELD_NODENAME_RAW), nodeName))
-                        .filter(QueryBuilders.termQuery(path(FIELD_METRICS_TAGS, FIELD_TYPE), NODE))
+                        .filter(QueryBuilders.termQuery(path(FIELD_METRICS_TAGS, FIELD_TYPE), type))
                         .filter(QueryBuilders.termQuery(path(FIELD_DOCUMENT_TYPE), metric().getName()))
                         .filter(QueryBuilders.rangeQuery(metric().getTimestamp())
                                 .from(from.toInstant(ZoneOffset.UTC).toEpochMilli())
-                                .to(to.toInstant(ZoneOffset.UTC).toEpochMilli())))
-                .size(0);
-    }
-
-    protected SearchSourceBuilder podStatsQuery(final String nodeName, final LocalDateTime from,
-                                                 final LocalDateTime to) {
-        return new SearchSourceBuilder()
-                .query(QueryBuilders.boolQuery()
-                        .filter(QueryBuilders.termsQuery(path(FIELD_METRICS_TAGS, FIELD_NODENAME_RAW), nodeName))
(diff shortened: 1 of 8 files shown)


api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/PodFSRequester.java
```diff
@@ -53,12 +67,43 @@ public SearchRequest buildRequest(final Collection<String> resourceIds, final Lo
...
+        final HashMap<String, MonitoringStats.DisksUsage.DiskStats> diskUsageMap = new HashMap<>();
+        disksUsage.setStatsByDevices(diskUsageMap);
+        monitoringStats.setDisksUsage(disksUsage);
+
+        final MonitoringStats.DisksUsage.DiskStats diskStats = new MonitoringStats.DisksUsage.DiskStats();
+        diskStats.setCapacity(((Number)
+                MapUtils.emptyIfNull((Map) metrics.get(FILESYSTEM + "/" + LIMIT)).get(VALUE)).longValue()
+        );
+        diskStats.setUsableSpace(((Number)
+                MapUtils.emptyIfNull((Map) metrics.get(FILESYSTEM + "/" + USAGE)).get(VALUE)).longValue()
```
tcibinan: Calls to `.longValue()` will fail with NPE if `metrics` doesn't contain requesting fields. Probably we should leave both capacity and usable space unset in such case.

api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/PodFSRequester.java
```diff
@@ -53,12 +67,43 @@ public SearchRequest buildRequest(final Collection<String> resourceIds, final Lo
     protected SearchRequest buildStatsRequest(final String nodeName, final LocalDateTime from, final LocalDateTime to,
                                               final Duration interval) {
         return request(from, to,
-                podStatsQuery(nodeName, from, to)
+                containerStatsQuery(nodeName, POD_CONTAINER, from, to)
+                        .sort(FILE_SYSTEM_METRICS_TIMESTAMP)
```
tcibinan: I would recommend to reuse `ELKUsageMetric.POD_FS.name` here.

api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/AbstractMetricRequester.java
```diff
@@ -202,30 +202,16 @@ protected SearchRequest request(final LocalDateTime from, final LocalDateTime to
         return parseStatsResponse(executeRequest(request));
     }
 
-    protected SearchSourceBuilder nodeStatsQuery(final String nodeName, final LocalDateTime from,
-                                                 final LocalDateTime to) {
+    protected SearchSourceBuilder containerStatsQuery(final String nodeName, final String container,
```
tcibinan: Maybe it is better to call the method just `statsQuery` and the `container` parameter as `type` because it feels a bit misleading as long as it is not a container name.

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ESMonitoringManager.java
```diff
@@ -77,7 +78,7 @@ public long getPodDiskSpaceAvailable(final String nodeName, final String podId,
         final MonitoringStats.DisksUsage.DiskStats diskStats =
                 AbstractMetricRequester.getStatsRequester(ELKUsageMetric.POD_FS, client)
                         .requestStats(nodeName,
-                                DateUtils.nowUTC().minus(duration),
+                                DateUtils.nowUTC().minus(duration.multipliedBy(TWO)),
```
tcibinan: I suppose we can use more than just two durations because we take the last one anyway. Probably we can use something like `Math.max(numberOfIntervals(), 2)`.

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ESMonitoringManager.java
```diff
@@ -77,7 +78,7 @@ public long getPodDiskSpaceAvailable(final String nodeName, final String podId,
...
                                 DateUtils.nowUTC(),
                                 duration
                         )
```
tcibinan, line 84: As long as we can get more than one monitoring stats from such interval we should sort monitoring stats by `startTime` and get only the latest one. Use can reuse sorting from `getStats` method.

tcibinan (changes requested):

tcibinan (approved):

PR #751, opened 2019-10-18
"pipe update" for one-folder distribution

ekazachkova:
Implements issue #589  for Linux for folder distributive

Since `build_linux.sh` builds `one-file` and `one-folder` distributive at the same time we should consider this behavior during `update` operation. To determine the type of distribution we use `build.info` file which contains such information.
- if this file contains `one-file` string `Cloud Pipeline CLI` will update current executable file
- if this file contains `one-folder` string `Cloud Pipeline CLI` will update corresponding folder:
1. downloads `pipe.tar.gz` from Cloud Pipeline API into temporary folder `~.pipe/tmp`
2. extracts it into temporary folder `~.pipe/tmp`
3. removes all under folder with executable files and copies extracted files into it
- if `build.info` file does not exists  `Cloud Pipeline CLI` will update as in case with single executable file
- if `build.info` file exists but contains string that is different from `one-file` or `one-folder` values an error will be occurred.


ekazachkova:
Added support for el6 Linux distros:
- if `bundle.info` file does not exists `Cloud Pipeline CLI` will be updated as in case with not `centos6` dist
- if `bundle.info` contains `centos:6` as second line `Cloud Pipeline CLI` will be updated for `el6` linux distr
- in other cases `Cloud Pipeline CLI` works as before
@sidoruka could you check this changes?

PR #752, opened 2019-10-18
Extract functionality of pods releasing from PodMonitor to its own service

Wedds:
Some tests revealed, that `updateStatus` and `releaseUnusedPods` are not working correctly if rates are changed in preferences. We decided to extract the functionality of unused pod release to side service and schedule this task using Spring scheduling directly.


mzueva:
We need to rethink refactoring of PodMonitor, I'll close this PR for now.

Issue #744, opened 2019-10-16
AWS: Nodeup retry process may stuck when first attempt to create a spot instance failed


mzueva:
**Describe the bug**
We observer the following behaviour when running AWS spot instances:

1. First nodeup attempt creates SpotRequest, waits for its fulfilment, finds created instance and waits for instance running state and connection on 8888 port
2. Connection on 8888 port is not available in exepected amount of attempts and current 
 nodeup attempt is failed, instance is terminated.
3. Second nodeup attempt starts, it finds existing SpotRequest by run id, checks its status (fullfilled) and starts to wait for the same instance (associated with SpotRequest) to be up. This will never happen since instance is already in terminating state.

**Expected behaviour**
We shall check that instance associated with SpotRequest is in appropriate status, if not - create a new SpotRequest and start nodeup process from scratch.

**Environment:**
 - Cloud Provider: AWS
 - Version  0.15.0.2994

sidoruka:
Cancel spot request when terminating a node

Please backport to `release/0.15`

PR #756, opened 2019-10-21
Issue #744 Cancel spot request if terminate instance

SilinPavel:
This PR related to #744 
Now if spot instance is terminated by nodeup.py we also will cancel spot-request of this instance.
Also, with every next attempt of nodeup.py we will also check that not only spot-request is fullfilled but instance in `pending` or `running` state as well

### DIFF
--- a/scripts/autoscaling/aws/nodeup.py
+++ b/scripts/autoscaling/aws/nodeup.py
@@ -575,6 +575,11 @@ def label_node(nodename, run_id, api, cluster_name, cluster_role, aws_region):
     pipe_log('Instance {} is assigned to RunID: {}\n-'.format(nodename, run_id))
 
 
+def instance_is_active(ec2, instance_id):
+    status = get_current_status(ec2, instance_id)
+    return status == 16 or status == 0
+
+
 def verify_run_id(ec2, run_id):
     pipe_log('Checking if instance already exists for RunID {}'.format(run_id))
     response = ec2.describe_instances(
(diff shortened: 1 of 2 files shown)


scripts/autoscaling/aws/nodeup.py
```diff
@@ -543,8 +543,10 @@ def check_instance(ec2, ins_id, run_id, num_rep, time_rep):
     pipe_log('- Waiting for instance boot up...')
     result = poll_instance(sock, time_rep, ipaddr, port)
     rep = 0
-    while result != 0:
+    active = False
+    while result != 0 or not active:
```
mzueva, line 550: I think it shall be `and` instead of `or`

scripts/autoscaling/aws/nodeup.py
```diff
@@ -575,6 +577,11 @@ def label_node(nodename, run_id, api, cluster_name, cluster_role, aws_region):
     pipe_log('Instance {} is assigned to RunID: {}\n-'.format(nodename, run_id))
 
 
+def instance_is_active(ec2, instance_id):
+    status = get_current_status(ec2, instance_id)
+    return status == 16 or status == 0
```
mzueva: Could you please extract numbers `16` and `0` to constants?

mzueva (changes requested):

scripts/autoscaling/aws/nodeup.py
SilinPavel: No, `or` is relevant here, this statement check that instance return an answer on poll request and its state still active (Running or Pending), because some times we have situation when instance already terminating (according to status on a cloud), but instance its self still doing work and can answer on poll request

scripts/autoscaling/aws/nodeup.py
SilinPavel: done

scripts/autoscaling/aws/nodeup.py
SilinPavel: So for the initial issue I added here checks in the lines `926` and `601`, and here is check for another case, this case can be reproduced when you manually terminate instance on the aws console and at the same time nodeup already waits on polling

mzueva: OK, I got it

mzueva:
@SilinPavel I'm fine with these changes. Could you please also move it to common code in pipeline library and I'll merge it.

mzueva (approved):

PR #763, opened 2019-10-22
(issue #624) Fix bug when IDLE tag wasn't set after a long time

SilinPavel:
This PR related to #624 
It fixes bug when check nodes for CPU usage can be skipped when `system.max.idle.timeout.minutes` set to 1, it this case some of the data can be unalienable at the moment in elastic search. This PR move a window for monitoring request.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/AbstractMetricRequester.java
+++ b/api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/AbstractMetricRequester.java
@@ -19,6 +19,7 @@ package com.epam.pipeline.dao.monitoring.metricrequester;
 import com.epam.pipeline.entity.cluster.monitoring.ELKUsageMetric;
 import com.epam.pipeline.entity.cluster.monitoring.MonitoringStats;
 import com.epam.pipeline.exception.PipelineException;
+import org.apache.commons.lang3.tuple.Pair;
 import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.action.support.IndicesOptions;
@@ -29,6 +30,7 @@ import org.elasticsearch.search.aggregations.AggregationBuilders;
 import org.elasticsearch.search.aggregations.Aggregations;
 import org.elasticsearch.search.aggregations.bucket.MultiBucketsAggregation;
 import org.elasticsearch.search.aggregations.bucket.histogram.DateHistogramAggregationBuilder;
+import org.elasticsearch.search.aggregations.bucket.terms.Terms;
 import org.elasticsearch.search.aggregations.metrics.ParsedSingleValueNumericMetricsAggregation;
 import org.elasticsearch.search.aggregations.metrics.avg.AvgAggregationBuilder;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
@@ -45,6 +47,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public abstract class AbstractMetricRequester implements MetricRequester, MonitoringRequester {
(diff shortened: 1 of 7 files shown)


api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/CPURequester.java
```diff
@@ -75,6 +75,7 @@ public SearchRequest buildRequest(final Collection<String> resourceIds, final Lo
     @Override
     public Map<String, Double> parseResponse(final SearchResponse response) {
         return ((Terms)response.getAggregations().get(AGGREGATION_POD_NAME)).getBuckets().stream()
+                .filter(b -> Double.isFinite(((Avg) b.getAggregations().get(AVG_AGGREGATION + USAGE_RATE)).getValue()))
```
tcibinan: Probably we could remove duplicated code by using temporary `Pair` object. Something like the following. Moreover we can reuse such code in all requesters.
```java
buckets
.map(b -> Pair.of(b.getKey().toString(), ((Avg) b.getAggregations().get(AVG_AGGREGATION + NODE_UTILIZATION)).getValue()))
.filter(utilizationEntry -> Double.isFinite(utilizationEntry.getRight()))
.collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
```

tcibinan: Btw, there is a handy null-safe `AbstractMetricRequester.doubleValue` function which can be used to reduce boilerplate code. Moreover it already filters out infinite numbers.
```
doubleValue(aggregations(b), AVG_AGGREGATION + NODE_UTILIZATION))
```

api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/MemoryRequester.java
```diff
@@ -73,6 +73,8 @@ public SearchRequest buildRequest(final Collection<String> resourceIds, final Lo
     @Override
     public Map<String, Double> parseResponse(final SearchResponse response) {
         return ((Terms) response.getAggregations().get(AGGREGATION_NODE_NAME)).getBuckets().stream()
+                .filter(b -> Double.isFinite(((Avg) b.getAggregations()
```
tcibinan: Same here.

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -131,27 +132,28 @@ private void processOverloadedRuns(final List<PipelineRun> runs) {
...
+        log.debug(messageHelper.getMessage(MessageConstants.DEBUG_RUN_METRICS_REQUEST,
+                "MEMORY, DISK ", running.size(), String.join(", ", running.keySet())));
 
-        final LocalDateTime now = DateUtils.nowUTC();
+        final LocalDateTime previousMinute = previousMinuteTime();
         final Map<ELKUsageMetric, Map<String, Double>> metrics = Stream.of(ELKUsageMetric.MEM, ELKUsageMetric.FS)
                 .collect(Collectors.toMap(metric -> metric, metric ->
                         monitoringDao.loadMetrics(metric, running.keySet(),
-                            now.minusMinutes(timeRange), now)));
+                                previousMinute.minusMinutes(timeRange), previousMinute)));
```
tcibinan: Probably the right end of the monitoring period can be `now()`? In this case actual time range will be bigger then the one specified in the preference but at least we will receive relevant stats.

tcibinan (changes requested):

api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/AbstractMetricRequester.java
```diff
@@ -151,6 +154,17 @@ public static MonitoringRequester getStatsRequester(final ELKUsageMetric metric,
         );
     }
 
+    public Map<String, Double> collectAggregation(final SearchResponse response,
+                                                  final String aggName, final String subAggName) {
+        return ((Terms)response.getAggregations().get(aggName)).getBuckets()
+                .stream()
+                .map(b -> Pair.of(b.getKey().toString(),
+                        doubleValue(b.getAggregations().asList(), subAggName))
```
tcibinan: I would suggest to use `aggregations(b)` call here rather than `b.getAggregations().asList()`. It does just the same but is null-safe.

tcibinan (approved):

mzueva (approved):

Issue #733, opened 2019-10-14
Search: ES synchronization improvements


mzueva:
**Background**
Recent performance tests have shown that Elasticsearch synchronization is prone to OutOfMemory errors in case of large amount of synchronized data or large size of individual documents (run logs, files in Gitlab).

**Approach**
We shall introduce a number of improvements to have a better control over memory consumption and make synchronization more stable.
1. Process of events loading, merging and processing shall be splitted into chunks according to settings in application.properties.
2. Currently if syncronization fails at some point, e.g. when uploading one of the documents to Elasticsearch all synchonization step fails and events for processed entities are not cleared from DB. We shall handle this situation and clear events for successfully synchronized entities, even if some other entities were not synchronized.
3. Size of ES documents shall be limited according to settings in application.properties. For now we shall apply this to PipelineRun entities and content of Gitlab (Pipeline) files.

**Minor Fixes**

- Check what creates sync events for `pipeline_user` entity type

mzueva:
@Wedds We shall add request size limit control to `BulkRequestSender`. Let's additionally split provided `documentRequests` (`indexChunk` method)  into parts matching  `elastic.request.limit.size` from application.properties. 
If any single document exceeds this limit, skip it and log appropriate message.

PR #764, opened 2019-10-22
Issue 733 Bulk request size limitation

Wedds:
This PR partially solves issue #733 .
Request chunk splitting additionally covered by chunk memory size limitation besides elements count limitation. Elasticsearch can't index request with size over the value of its `http.max_content_length` property, so we are tracking bulk request size.

- introduce `elastic.request.limit.size.mb` in application.properties;
- monitor documents' size during chunk creation (execute bulk request if the next request added will cause limitation oversize; skip the files exceeding the limit).

### DIFF
--- a/elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/BulkRequestSender.java
+++ b/elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/BulkRequestSender.java
@@ -20,14 +20,16 @@ import com.epam.pipeline.elasticsearchagent.service.BulkResponsePostProcessor;
 import com.epam.pipeline.elasticsearchagent.service.ElasticsearchServiceClient;
 import com.epam.pipeline.elasticsearchagent.service.ResponseIdConverter;
 import lombok.AllArgsConstructor;
+import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.collections4.ListUtils;
 import org.elasticsearch.action.DocWriteRequest;
 import org.elasticsearch.action.bulk.BulkResponse;
+import org.elasticsearch.action.index.IndexRequest;
 import org.springframework.util.ObjectUtils;
 
 import java.time.LocalDateTime;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/BulkRequestSender.java
```diff
@@ -112,4 +140,56 @@ private void indexChunk(final String indexName,
...
     }
+
+    private boolean exceedsDocSizeLimit(final DocWriteRequest request) {
+        return RequestChunk.getRequestSizeMb(request) > requestLimitMb;
+    }
+
+    @Getter
+    private static class RequestChunk {
+
+        private double sizeMB;
```
mzueva: Can't we use integer or long for size? Double comparison maybe tricky, and I generally try to avoid it.

elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/BulkRequestSender.java
```diff
@@ -87,13 +92,36 @@ public void indexDocuments(final String indexName,
...
+            .forEach(request -> tryToProceedRequestsInChunk(indexName, objectTypes, syncStart, requestChunk, request));
+        if (!requestChunk.isEmpty()) {
+            tryToIndexChunk(indexName, objectTypes, syncStart, requestChunk);
+        }
+    }
+
+    private void tryToProceedRequestsInChunk(final String indexName, final List<PipelineEvent.ObjectType> objectTypes,
+                                             final LocalDateTime syncStart, final RequestChunk chunk,
+                                             final DocWriteRequest request) {
+        if (exceedsDocSizeLimit(request)) {
```
mzueva: I think we can reuse approach from `IndexRequestContainer`, as it provides a very similar logic but it checks only size, so we can add optional check of size

mzueva (changes requested):

elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/BulkRequestSender.java
Wedds: Comparing size in bytes from now

elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/BulkRequestSender.java
Wedds: Done

Issue #759, opened 2019-10-21
Allow to restrict the visibility of the jobs


sidoruka:
**Background**
At the moment, Cloud Pipeline inherits the jobs' permissions from the parent object (e.g. from a pipeline or a tool). E.g.:
* `user1` launches `tool1` with run id `123`
* `tool1` is readable by `user2`
* `user2` is able to see the the run with id `123`

In certain use cases, this is not desired and shall be limited.

**Approach**
The following behavior shall be implemented to restrict the visibility of the jobs:
* There shall be a global preference that shall switch behavior:
  * Inherit (current approach, when the visibility is controlled by the tool/pipeline permissions)
  * Only owner (when only the person who launch a job can see it). Shall be set as a default for the Cloud Pipeline environment
* This can be also controlled per user/group (we shall add the same preference to the user/group profile configuration)
* Admins shall see everything for both cases as now

sidoruka:
This shall be backported to `release/0.15`

mzueva:
@sidoruka Just a minor clarification: `Pipeline runs` inherit permissions only from `Pipelines`, tool runs are visible only to owner and admins.

mzueva:
Implemented via #769

rodichenko:
GUI part implemented via ef281ef6fc679d5a73c077f5c4b51bcd4238abdf

rodichenko:
@sidoruka backported to `release/0.15` via de47c0aa9a48151f49d5948b5f273de6a6b0e361

NShaforostov:
I tried to verify this feature. It works well for users.
But I can't verify it for groups or global for Platform. I suppose that **Jobs visibility** dropdown box works not fully correct (for user or group) - it should be possible to leave this field empty (without value - e.g. how it done for **Allowed price types**). Otherwise, due to priority rule (global - group - user) the visibility of any pipeline run is always set only by value of this setting for a user.

@rodichenko can you fix it?

NShaforostov:
Docs updated via #860

rodichenko:
@NShaforostov fixed, backported to `release/0.15` via 439238db7a7051dc4bcd620310b97fe54a20ae1b

NShaforostov:
@rodichenko thanks! Verified. Now, it works correctly.

PR #769, opened 2019-10-23
Issue #759: Allow to configure run visibility policy

mzueva:
Provides implementation for #759

## Implementation
New SystemPreference `launch.run.visibility` was added with two supported values: `OWNER` and `INHERIT` (default). 
- When `launch.run.visibility` is `INHERIT` non-admin users will see their own runs and run of pipelines they have access to.
- When `launch.run.visibility` is `OWNER` non-admin users will see only their own runs

Admin users will have access to all runs in both cases. 
Value of SystemPreference `launch.run.visibility` maybe overridden for a specific user or group (Cloud Pipeline managed role) using contextual preferences (`PUT /contextual/preferences`). The most specific preference (User -> Role -> System) will be applied during permission check.

**_Note!_** Run permissions are still not inherited from tools.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/security/PermissionsHelper.java
+++ b/api/src/main/java/com/epam/pipeline/manager/security/PermissionsHelper.java
@@ -59,12 +59,7 @@ public class PermissionsHelper {
         return isOwner(entity.getOwner());
     }
 
-    public boolean isAdmin() {
-        final GrantedAuthoritySid admin = new GrantedAuthoritySid(DefaultRoles.ROLE_ADMIN.getName());
-        return getSids().stream().anyMatch(sid -> sid.equals(admin));
-    }
-
-    private boolean isOwner(final String owner) {
+    public boolean isOwner(final String owner) {
         if (StringUtils.isBlank(owner)) {
             return false;
         }
@@ -73,6 +68,11 @@ public class PermissionsHelper {
             return false;
         }
         return owner.equalsIgnoreCase(currentUser);
+    }
+
+    public boolean isAdmin() {
+        final GrantedAuthoritySid admin = new GrantedAuthoritySid(DefaultRoles.ROLE_ADMIN.getName());
+        return getSids().stream().anyMatch(sid -> sid.equals(admin));
     }
 
     private List<Sid> getSids() {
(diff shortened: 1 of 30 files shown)


api/src/main/java/com/epam/pipeline/manager/security/PermissionsHelper.java
```diff
@@ -17,29 +17,59 @@
...
+        final GrantedAuthoritySid admin = new GrantedAuthoritySid(DefaultRoles.ROLE_ADMIN.getName());
+        return getSids().stream().anyMatch(sid -> sid.equals(admin));
+    }
+
+    private List<Sid> getSids() {
+        final Authentication authentication = authManager.getAuthentication();
+        return sidRetrievalStrategy.getSids(authentication);
+    }
+
+    private boolean isOwner(final String owner) {
```
tcibinan: Let's move overloaded methods closer to each over.

api/src/main/java/com/epam/pipeline/manager/security/run/RunVisibilityPolicy.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.manager.security.run;
+
+public enum RunVisibilityPolicy {
```
tcibinan, line 22: Just minor javadoc issues.

api/src/test/java/com/epam/pipeline/acl/run/PipelineAclFactory.java
```diff
@@ -0,0 +1,100 @@
...
+                .when(mockPipelineManager)
+                .load(eq(TEST_PIPELINE_ID));
+        doReturn(pipeline)
+                .when(mockPipelineManager)
+                .load(eq(TEST_PIPELINE_ID), anyBoolean());
+        aclService.getOrCreateObjectIdentity(pipeline);
+        // we need to grant permissions before changing the owner,
+        // because it may cause AccessDeniedException for non-owner, non-admin user
+        Optional.ofNullable(user)
+                .ifPresent(u -> grantPermission(pipeline, user, permissionMask));
```
tcibinan: Probably we should use `u` rather then `user` here.

Issue #757, opened 2019-10-21
Extend vm-monitor's supported events


sidoruka:
**Background**
At the moment, Cloud Pipeline can notify the platform's admins only about the "lost" compute nodes. 
There is also a number of the events, that shall be sent to the admins ASAP:
* If one of the Cloud Pipeline's services changes it's state (i.e. goes down or up)
* If one of the certificates is going to expire in N days (shall be configurable)

**Approach**

*Readiness*
* VM-Monitor shall check the `AVAILABLE` count the Kube deployments and compare to `DESIRED`
* If `AVAILABLE` is less than `DESIRED`, a notification shall be fired
* The list of the deployments shall be configurable, with the following defaults:
  * cp-api-db 
  * cp-api-srv
  * cp-notifier
  * cp-git
  * cp-git-sync
  * cp-edge
  * cp-docker-registry
  * cp-docker-comp
  * cp-clair
  * cp-search-elk
  * cp-search-srv
  * cp-heapster-elk
  * cp-heapster
  * cp-share-srv
  * cp-dav

*PKI expiration*
* VM-Monitor shall check all the PKI assets, available for the platform
* A list of the directories shall be configured as the list of `cp-config-global` configmap parameters (e.g. `CP_SHARE_SRV_CERT_DIR`)
* The following defaults shall be scanned:
  * `CP_SHARE_SRV_CERT_DIR`
  * `CP_EDGE_CERT_DIR`
  * `CP_DOCKER_CERT_DIR`
  * `CP_IDP_CERT_DIR`
  * `CP_GITLAB_CERT_DIR`
  * `CP_API_SRV_CERT_DIR`
  * `CP_COMMON_CERT_DIR`
* VM-Monitor shall traverse the specified directories and verify the `*.pem` files' expiration date (mask shall be configurable)
* If the expiration date is less than `now() - days_to_notify` - a notification shall be fired (notification shall include the path of the PKI file and it's CN + SAN)
* `days_to_notify` shall be configurable

sidoruka:
This shall be made available in `release/0.15` as well

mzueva:
@Wedds 
Please, implement a second part of this task:
Add a new scheduled task to `VMMonitor` service. This task shall list all folders provided in application properties, find files matching *.pem extension and validate/notify according to the description above.

NShaforostov:
Docs were updated via #860

PR #773, opened 2019-10-24
Issue #757 PKI certificates expiration monitoring

Wedds:
This PR solves the second part of issue #757.

CertificateMonitor service is introduced: it traverses specified directories, searches for certificate files and verifies their expiration date. If a certificate expires less than in a certain amount of days the notification is sent to the user.

- Implement certificates monitoring functionality.
- Modify `application.properties` to let the user configure: 
    1.  _days_ before expiration to notify 
    2.  certificates' file _extensions_
    3.  _directories_ to be scanned.

### DIFF
--- a/vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/model/cert/PkiCertificate.java
+++ b/vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/model/cert/PkiCertificate.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2017-2019 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.vmmonitor.model.cert;
+
+import lombok.Data;
+
+import java.nio.file.Path;
+import java.security.cert.X509Certificate;
+
+@Data
+public class PkiCertificate {
+
+    private final X509Certificate x509Certificate;
+    private final Path pathToCertFile;
+
+}
--- a/vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/model/cert/PkiCertificate.java
+++ b/vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/model/cert/PkiCertificate.java
@@ -16,12 +16,12 @@
 
 package com.epam.pipeline.vmmonitor.model.cert;
 
-import lombok.Data;
+import lombok.Value;
 
 import java.nio.file.Path;
 import java.security.cert.X509Certificate;
 
-@Data
+@Value
 public class PkiCertificate {
 
     private final X509Certificate x509Certificate;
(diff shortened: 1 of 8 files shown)


vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/CertificateMonitor.java
```diff
@@ -0,0 +1,105 @@
...
+                              @Value("${monitor.cert.file.masks:pem}") final String certificateMasks,
+                              @Value("${monitor.cert.expiration.notification.days:5}") final int daysToNotify) {
+        this.notificationService = notificationService;
+        this.directoriesToScan = Arrays.asList(directoriesToScan.split(DELIMITER));
+        this.certificateMasks = Arrays.asList(certificateMasks.split(DELIMITER));
+        this.daysToNotify = daysToNotify;
+    }
+
+    public void checkCertificates() {
+        directoriesToScan.forEach(this::scanDirForCertificates);
```
mzueva, line 64: Please add check for empty values (just skip it), for case, when some of folder env vars are not defined.

mzueva (changes requested):

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/model/cert/PkiCertificate.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.epam.pipeline.vmmonitor.model.cert;
+
+import lombok.Data;
+
+import java.nio.file.Path;
+import java.security.cert.X509Certificate;
+
+@Data
```
tcibinan: I suppose it will be better to mark it as `@Value` as long as the class is immutable.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/CertificateMonitor.java
Wedds: Done

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/impl/VMNotificationServiceImpl.java
```diff
@@ -89,6 +99,14 @@ public void notifyMissingLabels(final VirtualMachine vm,
         sendMessage(parameters, missingLabelsSubject, missingLabelsTemplatePath);
     }
 
+    @Override
+    public void notifyExpiringCertificate(final PkiCertificate certificate) {
+        final Map<String, Object> parameters = new HashMap<>();
```
tcibinan: It would be better to create an instance of `parameters` map right into `addCertParameters` method.

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/CertificateMonitor.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private void scanDirForCertificates(final String dirPath) {
+        if (dirPath.isEmpty()) {
```
mzueva: I meant something like StringUtils.isBlank(dirPath) - to prevent listing of empty or null string

mzueva (changes requested):

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/model/cert/PkiCertificate.java
Wedds: Done

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/impl/VMNotificationServiceImpl.java
Wedds: Done

mzueva (approved):

Issue #767, opened 2019-10-23
Azure: Server and GUI shall not allow Azure regions without storage account key


mzueva:
**Describe the bug**
Currently it is possible to save an Azure region without entering `storage account key` value. This leads to a broken configuration and breaks all operations with Azure datastorages.

**Expected behavior**
**Server:** `Storage account key` shall be required when a Azure new region is created. When an Azure region is updated and new value for `Storage account key` is not specified, we shall use previously stored value
**GUI**: `Storage account key` shall be required when a Azure new region is created.

**Environment:**
 - Cloud Provider: [Azure]
 - Version [0.16.0.3119.0795e94af6147dc71a0b2b354976398038714672]

PR #779, opened 2019-10-28
Issue 767 Check key for azure region

SilinPavel:
This PR related to #767 
If new storage key is not provided when region is updated - we will use old one
It also brings old changes for azure cloud region validation.


api/src/main/java/com/epam/pipeline/manager/region/AzureRegionHelper.java
```diff
@@ -74,4 +96,127 @@ public AzureRegion mergeRegions(final AzureRegion originalRegion, final AzureReg
...
+                    MessageConstants.ERROR_AZURE_RESOURCE_GROUP_NOT_FOUND, resourceGroup));
+        } catch (AuthenticationException e) {
+            throw new IllegalArgumentException(messageHelper.getMessage(
+                    MessageConstants.ERROR_AZURE_AUTH_FILE_IS_INVALID), e);
+        }
+    }
+
+    private Azure buildClient(final String authFile) {
+        try {
+            final Azure.Configurable builder = Azure.configure()
```
mzueva: I think this code is duplicated, could you please extract it to some common class.

api/src/main/java/com/epam/pipeline/manager/region/AzureRegionHelper.java
```diff
@@ -74,4 +96,127 @@ public AzureRegion mergeRegions(final AzureRegion originalRegion, final AzureReg
...
+    private Azure.Authenticated authenticate(final String authFile,
+                                             Azure.Configurable builder) throws IOException {
+        if (StringUtils.isBlank(authFile)) {
+            return builder.authenticate(AzureCliCredentials.create());
+        }
+        return builder.authenticate(new File(authFile));
+    }
+
+    private boolean isIpRangeValid(final AzurePolicy policy) {
+        if (!(StringUtils.isNotBlank(policy.getIpMax()) && StringUtils.isNotBlank(policy.getIpMin()))) {
```
mzueva: Maybe we should extract validation of two IPs range into some Util class?

mzueva (changes requested):

mzueva (approved):

PR #789, opened 2019-11-05
Restore folder for CLI (AWS support only)

ekazachkova:
This PR provides implementation for folder restore for Cloud Pipeline CLI (#792). For now this operation supports AWS provider only.


sidoruka:
@NShaforostov please take a look as well:
* Test it
* Update the CLI docs (shall be added into this PR)

pipe-cli/src/utilities/storage/s3.py
```diff
@@ -300,6 +303,30 @@ def load_delete_marker(self, bucket, path, client):
...
+        if path:
+            operation_parameters['Prefix'] = prefix
+        paginator = client.get_paginator('list_object_versions')
+        pages = paginator.paginate(**operation_parameters)
+        restore_us = dict(Objects=[])
+        for page in pages:
+            S3BucketOperations.process_listing(page, 'DeleteMarkers', restore_us, delimiter, exclude, include, prefix,
+                                               versions=True)
+            # flush once aws limit reached
+            if len(restore_us['Objects']) >= 1000:
```
mzueva: It seems, that this code will fail if `len(restore_us['Objects']) > 1000`

mzueva (changes requested):

pipe-cli/src/utilities/storage/s3.py
ekazachkova: fixed for deletion too

mzueva (approved):

mzueva:
@ekazachkova I've tested this branch and it seems to work fine, except for the following case: if I try to restore a file with `-r` option, nothing happens. @NShaforostov @sidoruka we need to decide, how to handle this case - throw an error if restored object is not a folder or restore it as file.

sidoruka:
> @ekazachkova I've tested this branch and it seems to work fine, except for the following case: if I try to restore a file with `-r` option, noting happens. @NShaforostov @sidoruka we need to decide, how to handle this case - throw an error if restored object is not a folder or restore it is file.

@mzueva how do we handle it for `pipe cp`?

NShaforostov:
Verified, works fine!

NShaforostov:
@sidoruka the CLI docs updated via [be7ad17](https://github.com/epam/cloud-pipeline/pull/789/commits/be7ad1772be461c2962ba843bf2376cd255c0e4c).

Issue #802, opened 2019-11-08
Allow to specify max number of autoscaled nodes


sidoruka:
**Background**
At the moment, Cloud Pipeline allows to spinup clusters, which are limited in size by `launch.max.scheduled.number`.
This preference is applied for both `static` and `autoscaled` clusters. E.g.:
* `launch.max.scheduled.number` is equal to 10
* If a user tries to create a static cluster with 11 nodes - an error message is thrown
* If a user tries to create an autoscaled cluster with 1 node at a startup and allow to scale up to 11 nodes - an error message is thrown as well

In certain cases, we shall allow distinguishing max size for `static` and `autoscaled` clusters. E.g. having `launch.max.scheduled.number = 10` (as in the case above), we shall be able to configure an autoscaled cluster up to (e.g.) 30 nodes. While having the `static` limit = `10`

**Approach**
* Introduce a new option: `ge.autoscaling.scale.up.to.max` (int)
  * No default value
  * Visible to the client
  * Value cannot be less than `launch.max.scheduled.number`
  * Group: `Grid engine autoscaling`
* GUI shall perform validation of the `Scale up to` value as:
  * If `ge.autoscaling.scale.up.to.max` is set - use it
  * If `ge.autoscaling.scale.up.to.max` is NOT set - use `launch.max.scheduled.number`

sidoruka:
@mzueva could you please verify, does the approach address the issue? May be some other changes may be required?

sidoruka:
@SilinPavel @rodichenko backport to `release/0.15` once done

SilinPavel:
@sidoruka @mzueva  I came across this idea, while I investigated this issue.
May be we don't need to introduce any new parameter here.
according to code `launch.max.scheduled.number` describes how much instances can be launching at the same time, and `cluster.max.size` describes how much instances can be active at the same time. So when we launch `autoscale` cluster we should worry only about minimal cluster size, since it will be run immediately in the same time, while other requested nodes (if they will be requested at all) will be launched in other run request, so I believe for `max size` of auto-scale cluster can be relevant only  `cluster.max.size`, just because we can't run instances more then `cluster.max.size` for the whole system.
So my approach would be:
 If we run autoscale cluster: 
- check on GUI that `Default child nodes` <= `launch.max.scheduled.number` (this already done on server because this value will be pass to `nodeCount` and we check this value)
- check on GUI that `Scale up to` <= `cluster.max.size` and if so warn user of this and ask to reconfigure

[...]

PR #804, opened 2019-11-08
(issue #802) New parameter for autoscale cluster

SilinPavel:
Introduce new system preference for maximum number of nodes for autoscale cluster

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
+++ b/api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
@@ -145,6 +145,14 @@ public final class PreferenceValidators {
         return (pref, dependencies) -> StringUtils.isNumeric(pref) && Long.parseLong(pref) > x;
     }
 
+    public static BiPredicate<String, Map<String, Preference>> isGreaterThanValueOrNull(String key) {
+        return (pref, dependencies) -> {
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
```diff
@@ -145,6 +145,14 @@
         return (pref, dependencies) -> StringUtils.isNumeric(pref) && Long.parseLong(pref) > x;
     }
 
+    public static BiPredicate<String, Map<String, Preference>> isGreaterThanValueOrNull(String key) {
+        return (pref, dependencies) -> {
+            Long valueToCompare = dependencies.containsKey(key) ? dependencies.get(key).get(Long::parseLong) : Long.MIN_VALUE;
+            return StringUtils.isBlank(pref) ||
+                    StringUtils.isNumeric(pref) && Long.parseLong(pref) > valueToCompare;
```
mzueva: `pref value` shall be >= `valueToCompare`

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
SilinPavel: @mzueva did you mean `>=`?

mzueva (approved):

Issue #298, opened 2019-05-16
GE Autoscaler shall respect CPU requirements of the job in the queue


sidoruka:
At the moment GE autoscaler treats each job in the queue as a `single-core` job.
But if one of the jobs was run using `-pe local 4` within a 2-cored machine - this will result in `unschedulable` jobs, that will wait in a queue forever (as autoscaler will setup a new 2-cored node).

We shall implement an approach, that allows processing the data even if the initial node type is not enough.

sidoruka:
**Proposal**
1. If the autoscaler decides to scale-up, it shall check the maximum core required among the queued jobs (e.g. we have 3 jobs in `qwait`: with 2/8/1 cores requested)
2. Autoscaler than shall contact `/price` API and request the appropriate node type of the family, defined by the master node (e.g. master is run in AWS using `m5.xlarge`, then autoscaler shall setup `m5.2xlarge` to fulfill `max(2,8,1)`)
3. This behavior shall be customizable via `System parameters`:
- `CP_CAP_AUTOSCALE_HYBRID` - (bool) enables this `hybrid` cluster configuration
- `CP_CAP_AUTOSCALE_HYBRID_FAMILY` - (string) overrides the instance family to be used for the node type selection (e.g. if master is run using `m5.xlarge` and `CP_CAP_AUTOSCALE_HYBRID_FAMILY` is equal to `r5` - autoscaler shall create `r5.2xlarge` for the (2) example)

@mzueva @tcibinan I'd like to hear your opinion regarding this

mzueva:
Some additions to the suggested approach

#### Algorithm of selecting a best matching node

- autoscaler detects that a new node shall be created, it checks the maximum cpu and ram requirements among all pending jobs
- list of all allowed instance types for current region is fetched from API `cluster/instance/allowed`
- all instance types not matching requirements are filtered out
- instances type are sorted by matching family, cpu, ram increasing and the first instance type is selected for run
- if no matching instance is present, autoscaler logs error message
- disk size for now is selected as master disk size

#### Scaling cluster down

- currently autoscaler scales down node only if job queue is empty
- autoscaler shall detect deadlock situation: a job is present in the queue with cpu requirement `8`, an idle node is present in the cluster with `4` cpu available, cluster is limited to size `1`
- if current cluster == maximum cluster size and idle nodes exist and it does not match any of pending jobs => such idle node shall be scaled down

#### TBD 

[...]

Issue #494, opened 2019-07-04
Dead worker jobs stuck in GE cluster


tcibinan:
There is an issue that prevents GE job from being marked as failed in GE in case of its host termination. In other words if some of the GE cluster workers is killed and it has running jobs then these jobs will stack in queue forever. Stuck jobs lead to whole clusters being stuck.

Fortunately GE detects the unavailable state of the terminated hosts and marks it as an `unheard` host. Moreover it can be configured to kill jobs on such unheard hosts using `reschedule_unknown` parameter in conjunction with `ENABLE_RESCHEDULE_KILL` setting. See more in the corresponding [man page](http://gridscheduler.sourceforge.net/htmlman/htmlman5/sge_conf.html).

```
qhost -h pipeline-1210 -q
    HOSTNAME                ARCH         NCPU NSOC NCOR NTHR  LOAD  MEMTOT  MEMUSE  SWAPTO  SWAPUS
    ----------------------------------------------------------------------------------------------
    global                  -               -    -    -    -     -       -       -       -       -
    pipeline-1210           lx-amd64        2    1    1    2     -    7.1G       -     0.0       -
       main.q               BIP   0/2/2         u
                                                ^
```

PR #817, opened 2019-12-02
Issue 298 respect sge job slots, autoscale depends on instance family

SilinPavel:
This PR related to #298.
The next changes were made:

- Reject job if it doesn't fit (f.e. we have job with local PE and 100 slots, currently we do not have instance on a cloud with such number of cores)
- `CP_CAP_AUTOSCALE_HYBRID` and `CP_CAP_AUTOSCALE_HYBRID_FAMILY` env vars introduced to be able to autoscale not only static type of instance, but also instances from some instance family
- For job with allocation rule like `round_robin` or `fill_up` script will count all available slots (f.e. we have 2 instance in the cluster with 4 cores, 2 cores already occupied by some job, we submit another job with 6 slots and allocation rule `round_robin`, instance with 4 cores will be run since we already have 2 available cores)


SilinPavel:
This PR was made according to approach to make changes locally in `GridEngine` class, however there are still exist some changes in API of such classes as `GridEngineAutoscaler` `GridEngineScaleUpHandler` `GridEngineScaleDownHandler` but it was made as abstract as possible in my point of view

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -138,14 +138,16 @@ def from_letter_code(code):
 
 class GridEngineJob:
 
-    def __init__(self, id, name, user, state, datetime, host=None, array=None):
+    def __init__(self, id, name, user, state, datetime, hosts=[], array=None, slots=0, pe="local"):
```
tcibinan: Using empty list as a default value is [pretty dangerous](https://nikos7am.com/posts/mutable-default-arguments/). We have used it before but it seems reasonable to avoid such kind of usage.

It will be safer to use `None` as a default value and then set it to an empty list in the body of `__init__` method.
```
self.hosts = hosts if hosts else []
```

tcibinan: Let's use single quotes as in the remaining script.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -272,6 +311,30 @@ def get_parallel_environment_slots(self, pe=_PARALLEL_ENVIRONMENT):
...
+        return str(self._or_default(self.cmd_executor.execute(GridEngine._SHOW_JOB_PARALLEL_ENVIRONMENT % job_id))).strip()
+
+    def get_job_slots(self, job_id):
+        """
+        Returns PE of the specific job.
+
+        :param job_id: id of a SGE job
+        """
+        return int(self._or_default(self.cmd_executor.execute(GridEngine._SHOW_JOB_PARALLEL_ENVIRONMENT_SLOTS % job_id), 0))
+
```
tcibinan: It seems that pythondoc for these three methods is kind of mixed up. Could you please verify or fix the docs?

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -155,59 +157,89 @@ class GridEngine:
...
-
-    def _parse_date(self, line, indentations):
-        return datetime.strptime(self._by_indent(line, indentations, 5), GridEngine._QSTAT_DATETIME_FORMAT)
-
-    def _parse_host(self, line, indentations):
-        queue_and_host = self._by_indent(line, indentations, 6)
+            tokens = line.strip().split()
+            # host line e.g. main.q@pipeline-18033          BIP   0/0/2          0.50     lx-amd64
+            if tokens[0].startswith(self._MAIN_Q):
+                current_host = tokens[0]
```
tcibinan: Currently, `current_host` variable contains not only the host name but also a query name. It is kind of misleading. Let's call `self._parse_host` right here. This way we don't need to call the method every time the host is required.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -155,59 +157,89 @@ class GridEngine:
...
+                continue
```
tcibinan: There is no need to call `continue` in here.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -155,59 +157,89 @@ class GridEngine:
...
+    _SHOW_PE_ALLOCATION_RULE = 'qconf -sp %s | grep "^allocation_rule" | awk \'{print $2}\''
     _REMOVE_HOST_FROM_HOST_GROUP = 'qconf -dattr hostgroup hostlist %s %s'
     _REMOVE_HOST_FROM_QUEUE_SETTINGS = 'qconf -purge queue slots %s@%s'
     _SHUTDOWN_HOST_EXECUTION_DAEMON = 'qconf -ke %s'
-    _QSTAT = 'qstat -u "*"'
+    _QSTAT = 'qstat -f -u "*"'
+    _SHOW_EXECUTION_HOSTS_SLOTS = 'qstat -f -u "*" | grep %s' % _MAIN_Q
     _QSTAT_DATETIME_FORMAT = '%m/%d/%Y %H:%M:%S'
-    _QSTAT_COLUMNS = ['job-ID', 'prior', 'name', 'user', 'state', 'submit/start at', 'queue', 'slots', 'ja-task-ID']
+    _QSTAT_COLUMNS = ['job-ID', 'prior', 'name', 'user', 'state', 'submit/start at', 'slots']
```
tcibinan: The field `_QSTAT_COLUMNS` is not used anywhere now. Let's remove it.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -155,59 +157,89 @@ class GridEngine:
...
+        return datetime.strptime(date, GridEngine._QSTAT_DATETIME_FORMAT)
+
+    def _parse_host(self, queue_and_host):
         return queue_and_host.split('@')[1] if queue_and_host else None
 
-    def _parse_array(self, line, indentations):
-        array_jobs = self._by_indent(line, indentations, 8)
+    def _parse_array(self, array_jobs):
         if not array_jobs:
             return None
```
tcibinan: Just to avoid some possible mistakes it will be better to return an empty list rather then `None` since we can merge job arrays from now on.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -155,59 +157,89 @@ class GridEngine:
...
+            if tokens[0].startswith(self._MAIN_Q):
+                current_host = tokens[0]
+                continue
+            # job line 15 0.50000 sleep.sh   root         r     11/27/2019 11:47:40     2
+            elif tokens[0].isdigit():
+                job_id = tokens[0]
+                if job_id in jobs:
+                    job = jobs[job_id]
+                    job.hosts.append(self._parse_host(current_host))
+                    job.array = sorted(job.array + self._parse_array(tokens[8] if len(tokens) >= 9 else None))
```
tcibinan: What will happen if we have an array job and a half of its subjobs is running and the other half is pending. It looks like we'll classify all the pending subjobs as running.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -272,6 +311,30 @@ def get_parallel_environment_slots(self, pe=_PARALLEL_ENVIRONMENT):
...
+        """
+        return self._or_default(self.cmd_executor.execute(GridEngine._SHOW_PE_ALLOCATION_RULE % pe), "$pe_slots").strip()
+
+    def get_job_parallel_environment(self, job_id):
+        """
+        Returns PE of the specific job.
+
+        :param job_id: id of a SGE job
+        """
+        return str(self._or_default(self.cmd_executor.execute(GridEngine._SHOW_JOB_PARALLEL_ENVIRONMENT % job_id))).strip()
```
tcibinan: There is no default value in here. Probably `local` can be used as a default here. What do you think?

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -272,6 +311,30 @@ def get_parallel_environment_slots(self, pe=_PARALLEL_ENVIRONMENT):
...
+
+    def get_job_slots(self, job_id):
+        """
+        Returns PE of the specific job.
+
+        :param job_id: id of a SGE job
+        """
+        return int(self._or_default(self.cmd_executor.execute(GridEngine._SHOW_JOB_PARALLEL_ENVIRONMENT_SLOTS % job_id), 0))
```
tcibinan: It seems that we should use `1` as a default number of slots. If you submit a job to a `pe_slots` query and doesn't specify slots explicitly it does mean that the job has a requirement of a single core.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -332,6 +417,9 @@ def _perform_command(self, action, msg, error_msg, skip_on_failure):
             if not skip_on_failure:
                 raise RuntimeError(error_msg, e)
 
+    def _or_default(self, value, default=None):
+        return value if value else default
```
tcibinan: I propose to use the following syntax. These way we can inline the method.
```python
def _or_default(self, value, default=None):
    return value or default
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -291,6 +354,28 @@ def delete_host(self, host, queue=_MAIN_Q, hostgroup=_ALL_HOSTS, skip_on_failure
         self._remove_host_from_host_group(host, hostgroup, skip_on_failure=skip_on_failure)
         self._remove_host_from_grid_engine(host, skip_on_failure=skip_on_failure)
 
+    def get_resource_demand(self, expired_jobs):
+        avaliable_slots = 0
+        demand_slots = 0
+        for line in self.cmd_executor.execute_to_lines(GridEngine._SHOW_EXECUTION_HOSTS_SLOTS):
```
tcibinan: Probably the number of available slots calculation can be extracted to a private method.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -272,6 +311,30 @@ def get_parallel_environment_slots(self, pe=_PARALLEL_ENVIRONMENT):
         """
         return int(self.cmd_executor.execute(GridEngine._SHOW_PARALLEL_ENVIRONMENT_SLOTS % pe).strip())
 
+    def get_pe_allocation_rule(self, pe):
+        """
+        Returns number of the parallel environment slots.
+
+        :param pe: Parallel environment to return number of slots for.
+        """
+        return self._or_default(self.cmd_executor.execute(GridEngine._SHOW_PE_ALLOCATION_RULE % pe), "$pe_slots").strip()
```
tcibinan: Let's extract a `GridEngineAllocationRule` enum.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -291,6 +354,28 @@ def delete_host(self, host, queue=_MAIN_Q, hostgroup=_ALL_HOSTS, skip_on_failure
...
         self._remove_host_from_grid_engine(host, skip_on_failure=skip_on_failure)
 
+    def get_resource_demand(self, expired_jobs):
+        avaliable_slots = 0
+        demand_slots = 0
+        for line in self.cmd_executor.execute_to_lines(GridEngine._SHOW_EXECUTION_HOSTS_SLOTS):
+            rsrv_used_total = line.strip().split()[2].split("/")
+            avaliable_slots += int(rsrv_used_total[2]) - int(rsrv_used_total[1]) - int(rsrv_used_total[0])
+        for job in expired_jobs:
+            if self.get_pe_allocation_rule(job.pe) == "$round_robin" or self.get_pe_allocation_rule(job.pe) == "$fill_up":
```
tcibinan: It will be better to use `in` syntax, such as the following
```python
if self.get_pe_allocation_rule(job.pe) in ["$round_robin", "$fill_up"]:
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -219,11 +251,18 @@ def _parse_array(self, line, indentations):
         else:
             return [int(array_jobs)]
 
-    def _by_indent(self, line, indentations, index):
-        if index >= len(indentations) - 1:
-            return line[indentations[index]:].strip()
-        else:
-            return line[indentations[index]:indentations[min(len(indentations) - 1, index + 1)]].strip()
+    def validate_job(self, job):
```
tcibinan: It sounds better to call this `is_job_valid` as long as it returns `boolean`.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -812,6 +918,14 @@ def scale_down(self, child_host):
         Logger.info('Start grid engine SCALING DOWN for %s host.' % child_host)
         return self.scale_down_handler.scale_down(child_host)
 
+    def is_job_fit_or_cancel(self, job):
```
tcibinan: Let's inline this method and kill the invalid jobs right in `scale` method.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -926,6 +1040,44 @@ def _retrieve_preference(pipe, preference, default_value):
         return default_value
 
 
+def _get_family_from_type(cloud_provider, instance_type):
+    if cloud_provider == "AWS":
```
tcibinan: Cloud providers enum will be of help here.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -926,6 +1040,44 @@ def _retrieve_preference(pipe, preference, default_value):
...
+        return [instance for instance in allowed if instance['name'] == default_instance]
+
+
+def _match_with_allowed(resource, allowed):
+    for instance in allowed:
+        if instance['vcpu'] >= resource.cpu:
+            return instance
+    return allowed.pop()
+
+
```
tcibinan: Let's encapsulate these methods in some `InstancePool` class.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -926,6 +1040,44 @@ def _retrieve_preference(pipe, preference, default_value):
         return default_value
 
 
+def _get_family_from_type(cloud_provider, instance_type):
```
tcibinan: Let's add tests for this method.

tcibinan (changes requested):

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: probably yes

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: yes, you right, it will be resolved as running, so this logic should be changed to be able to work with this case

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -314,27 +313,27 @@ def get_parallel_environment_slots(self, pe=_PARALLEL_ENVIRONMENT):
...
     def get_pe_allocation_rule(self, pe):
         """
-        Returns number of the parallel environment slots.
+        Returns allocation rule of the pe as string, e.g. '$round_robin'
 
-        :param pe: Parallel environment to return number of slots for.
+        :param pe: Parallel environment to return allocation rule.
         """
-        return self._or_default(self.cmd_executor.execute(GridEngine._SHOW_PE_ALLOCATION_RULE % pe), "$pe_slots").strip()
+        return AllocationRule(self.cmd_executor.execute(GridEngine._SHOW_PE_ALLOCATION_RULE % pe or "$pe_slots").strip())
```
tcibinan: Looks like a typo with braces, probably we can make things clearer
```python
stdout = self.cmd_executor.execute(GridEngine._SHOW_PE_ALLOCATION_RULE % pe)
return AllocationRule(stdout.strip()) if stdout else AllocationRule.PE_SLOTS
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -845,9 +840,19 @@ def scale(self):
         additional_hosts = self.host_storage.load_hosts()
         Logger.info('There are %s additional pipelines.' % len(additional_hosts))
         updated_jobs = self.grid_engine.get_jobs()
-        pending_jobs = [job for job in updated_jobs if job.state == GridEngineJobState.PENDING
-                        and self.is_job_fit_or_cancel(job)]
         running_jobs = [job for job in updated_jobs if job.state == GridEngineJobState.RUNNING]
+
+        # kill jobs that are pending and can't be satisfied with requested resource
+        # f.i. we have only 3 instance max and the biggest possible type has 10 cores but job requests 40 cores
+        pending_jobs = []
```
tcibinan: It looks pretty complicated. Maybe we can find all the invalid nodes first then find all the valid pending jobs.
```
pending_jobs = [job for job in updated_jobs if job.state == GridEngineJobState.PENDING]
invalid_jobs = [job for job in pending_jobs if not self.grid_engine.is_job_valid(job)]
# kill invalid jobs
valid_pending_jobs = [job for job in pending_jobs if job not in invalid_jobs]
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -991,6 +988,63 @@ def _remove_worker_from_hosts(self, host):
...
+
+    def __init__(self, cloud_provider, region_id, master_instance_type, instance_family, pipe):
+        self.region_id = region_id
+        self.pipe = pipe
+        self.instance_family = instance_family
+        self.master_instance_type = master_instance_type
+        self.cloud_provider = cloud_provider
+
+    @staticmethod
+    def get_family_from_type(cloud_provider, instance_type):
```
tcibinan, line 1141: Let's make it non-static. It can make things easier :)

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -991,6 +988,63 @@ def _remove_worker_from_hosts(self, host):
...
+                    [instance for instance in allowed if self._is_instance_from_family(instance["name"])],
+                    cmp=lambda i1, i2: i1['vcpu'] - i2['vcpu']
+            )
+        else:
+            Logger.info("Hybrid autoscaling is disabled, allowed list of instances will be trimmed to master instance type: {type}"
+                        .format(type=self.master_instance_type))
+            return [instance for instance in allowed if instance['name'] == self.master_instance_type]
+
+    @staticmethod
+    def match_with_allowed(resource, allowed):
```
tcibinan: It can be non-static as well.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -314,27 +313,27 @@ def get_parallel_environment_slots(self, pe=_PARALLEL_ENVIRONMENT):
 
     def get_pe_allocation_rule(self, pe):
         """
-        Returns number of the parallel environment slots.
+        Returns allocation rule of the pe as string, e.g. '$round_robin'
```
tcibinan: It is returned as an object rather than string. Let's just say it returns allocation rule.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -22,6 +22,7 @@
 import time
 import multiprocessing
 import re
+from enum import Enum
```
tcibinan: I'm not sure that `enum` package is available in python 2. We can just use the same `CloudProvider` and `AllocationRule` classes with an additional factory methods.

SilinPavel: It's here, so let it be

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: good point

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: Here I would stick with my solution, yes it has a little bit more nesting in the code, but if we move it to method let say 'filter_pending_jobs' it would be pretty accurate and self documented. Also here we have only 2 iteration through the array while filtering instead of 4

tcibinan: Sounds reasonable.

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: now work with array jobs is performed as it is just different jobs. e.g.:
we have submitted qsub -t 1-10 <script>
so there will be 10 jobs like 1.1 1.2 1.3

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: here i would leave it as is, I believe this method is really `static` by its function

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: agreed, now it's called as `select_instance` and it gets `resource`, `price_type` and `hybrid_autoscale`  to be able to call `get_allowed_instances`, so we don't need to call it in `scale_up` of `GridEngineScaleUpHandler`, we just call `self.instance_helper.select_instance(resource, self.price_type, self.hybrid_autoscale)` and we are good to go with selected `instance_type` to run an additional worker

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -291,6 +360,32 @@ def delete_host(self, host, queue=_MAIN_Q, hostgroup=_ALL_HOSTS, skip_on_failure
...
+                    demand_slots += job.slots - available_slots
+                    available_slots = 0
+            else:
+                demand_slots += job.slots
+        return ComputeResource(demand_slots)
+
+    def _get_available_slots(self):
+        available_slots = 0
+        for line in self.cmd_executor.execute_to_lines(GridEngine._SHOW_EXECUTION_HOSTS_SLOTS):
+            rsrv_used_total = line.strip().split()[2].split("/")
```
tcibinan, line 408: Please add a comment with an expected line format.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -389,11 +487,9 @@ def __init__(self, cmd_executor, pipe, grid_engine, host_storage, parent_run_id,
         :param parent_run_id: Additional nodes parent run id.
         :param default_hostfile: Default host file location.
         :param instance_disk: Additional nodes disk size.
-        :param instance_type: Additional nodes instance type.
         :param instance_image: Additional nodes docker image.
         :param price_type: Additional nodes price type.
         :param region_id: Additional nodes Cloud Region id.
-        :param instance_cores:  Additional nodes cores number.
         :param polling_timeout: Kubernetes and Pipeline APIs polling timeout - in seconds.
         :param polling_delay: Polling delay - in seconds.
```
tcibinan, line 557: Please add pythondoc about the new parameters.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -876,6 +995,105 @@ def _remove_worker_from_hosts(self, host):
...
+        self.master_instance_type = master_instance_type
+        self.cloud_provider = cloud_provider
+        self.max_core_per_instance = max_core_per_instance
+
+    def select_instance(self, resource, price_type, hybrid_autoscale):
+        allowed = self._get_allowed_instances(price_type, hybrid_autoscale)
+        for instance in allowed:
+            if instance['vcpu'] >= resource.cpu:
+                return instance
+        return allowed.pop()
```
tcibinan, line 1113: Probably we can return an object of some `Instance` type rather then a `dict` which is returned by `PipelineApi`. What do you think?

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -876,6 +995,105 @@ def _remove_worker_from_hosts(self, host):
...
+
+    def get_max_allowed(self, price_type, hybrid_autoscale):
+        return self._get_allowed_instances(price_type, hybrid_autoscale).pop()
+
+    def _get_allowed_instances(self, price_type, hybrid_autoscale):
+        is_spot = price_type == "on-demand" or price_type == "on_demand"
+        allowed = pipe.get_allowed_instance_types(self.region_id, is_spot)["cluster.allowed.instance.types"]
+        if hybrid_autoscale:
+            return sorted(
+                    [instance for instance in allowed if self._is_instance_from_family(instance["name"])
```
tcibinan: Probably we can extract a private method here.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -876,6 +995,105 @@ def _remove_worker_from_hosts(self, host):
...
+            if instance['vcpu'] >= resource.cpu:
+                return instance
+        return allowed.pop()
+
+    def get_max_allowed(self, price_type, hybrid_autoscale):
+        return self._get_allowed_instances(price_type, hybrid_autoscale).pop()
+
+    def _get_allowed_instances(self, price_type, hybrid_autoscale):
+        is_spot = price_type == "on-demand" or price_type == "on_demand"
+        allowed = pipe.get_allowed_instance_types(self.region_id, is_spot)["cluster.allowed.instance.types"]
```
tcibinan: Are you sure that `cluster.allowed.instance.types` is sufficient here? As far as I remember where are two groups of allowed instances. One for pipelines and the other one for tools. It seems that we have to differentiate them somehow.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -876,6 +995,105 @@ def _remove_worker_from_hosts(self, host):
...
+            return None
+
+    def _is_instance_from_family(self, instance_type):
+        return CloudPipelineInstanceHelper.get_family_from_type(self.cloud_provider, instance_type) == self.instance_family
+
+
+class CloudProvider:
+
+    def __init__(self, value):
+        self.value = value
```
tcibinan: Let's validate that cloud provider is one of the supported ones. Same for `AllocationRule` class.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -876,6 +995,105 @@ def _remove_worker_from_hosts(self, host):
...
+        return CloudProvider("AZURE")
+
+    def __eq__(self, other):
+        if not isinstance(other, CloudProvider):
+            # don't attempt to compare against unrelated types
+            return False
+        return other.value == self.value
+
+
+class AllocationRule:
```
tcibinan, line 1188: Just several copy-paste typos in the class :) Please replace `CloudProvider` with `AllocationRule`.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -876,6 +995,105 @@ def _remove_worker_from_hosts(self, host):
...
+        if cloud_provider == CloudProvider.aws():
+            return instance_type.rsplit('.', 1)[0]
+        elif cloud_provider == CloudProvider.gcp():
+            return instance_type.rsplit('-', 2)[1] if "custom" not in instance_type else None
+        elif cloud_provider == CloudProvider.azure():
+            # will return Bms for Standard_B1ms or Dsv3 for Standard_D2s_v3 instance types
+            search = re.search("([a-zA-Z]+)\d+(.*)", instance_type.split('_', 1)[1].replace("_", ""))
+            return search.group(1) + search.group(2) if search else None
+        else:
+            return None
```
tcibinan, line 1153: This method returns `None` in some cases. But it looks like such behavior isn't properly handled later. Let's say that if family type is `None` we should just use master machine instance type. Does it seem reasonable?

Current implementation will just fail every time trying to determine suitable instance type.

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: yes, you are right there is should be `cluster.allowed.instance.types.docker`

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: my bad :)

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: definitely should raise exception here

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -951,26 +1217,45 @@ def _retrieve_preference(pipe, preference, default_value):
...
+
+    pipe = PipelineAPI(api_url=pipeline_api, log_dir='/common/workdir/.pipe.log')
+
+    instance_helper = CloudPipelineInstanceHelper(cloud_provider=cloud_provider, region_id=region_id,
+                                                  instance_family=instance_family, master_instance_type=instance_type,
+                                                  pipe=pipe, pipeline_version=pipeline_version,
+                                                  max_core_per_instance=autoscale_max_core_per_instance)
+
+    max_instance_cores = instance_helper.get_max_allowed(price_type, hybrid_autoscale).cpu
+    max_cluster_cores = max_instance_cores * max_additional_hosts + instance_cores
```
tcibinan: We should count permanent worker cores here as well as additional ones. Could you please look through the pull request once again maybe there are similar cases where we doesn't handle permanent worker properly? I've looked through and haven't found anything. But maybe you will.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -755,7 +877,11 @@ def scale(self):
                 if len(additional_hosts) < self.max_additional_hosts:
                     Logger.info('There are %s/%s additional child pipelines. Scaling up will be performed.' %
                                 (len(additional_hosts), self.max_additional_hosts))
-                    self.scale_up()
+                    resource = self.grid_engine.get_resource_demand(expired_jobs)
```
tcibinan: I'm concerned that we use `expired_jobs` here and not `pending_jobs`. Just imagine that we have only 1 expired job (it was in queue for 30 seconds) and 1000 jobs that were submitted within last 30 seconds. We'll scale up small instance rather than a big one. Let's change this.

workflows/pipe-common/scripts/autoscale_sge.py
SilinPavel: I also checked and didn't find anything else

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -951,26 +1232,46 @@ def _retrieve_preference(pipe, preference, default_value):
...
+    pipe = PipelineAPI(api_url=pipeline_api, log_dir='/common/workdir/.pipe.log')
+
+    instance_helper = CloudPipelineInstanceHelper(cloud_provider=cloud_provider, region_id=region_id,
+                                                  instance_family=instance_family, master_instance_type=instance_type,
+                                                  pipe=pipe, pipeline_version=pipeline_version,
+                                                  max_core_per_instance=autoscale_max_core_per_instance)
+
+    max_instance_cores = instance_helper.get_max_allowed(price_type, hybrid_autoscale).cpu
+    # (1 + int(os.getenv('node_count', 0))) - master plus default workers
+    max_cluster_cores = max_instance_cores * max_additional_hosts + instance_cores * (1 + int(os.getenv('node_count', 0)))
```
tcibinan: Let's just extract the variable for `default_worker_nodes_number` or smth. This way the comment can be omitted.

tcibinan (approved):

NShaforostov:
@SilinPavel, I tried to verify this feature and accidentally encountered the following issue on **AWS**:
1. Autoscale-cluster was launched with 3 possible autoscale-child. Master-node was _spot_ instance with 2 CPU
2. During the testing to the cluster were attached:
   - 1 worker-node with 8 CPU
   - 2 worker-nodes - each with 16 CPU
3. At the one moment, all nodes (master and workers) were out of jobs.
4. Around this moment 1 worker-node was terminated by **AWS** (`Server failure reason: Server.SpotInstanceTermination: Spot instance termination`)
5. I did not notice it right away, and try to start following "light" array-jobs in sequence:
   - `qsub -pe round_robin 3 -t 1:3 -b y sleep 3m`
   - `qsub -pe fill_up 3 -t 1:3 -b y sleep 3m`
   - `qsub -pe local 3 -t 1:3 -b y sleep 3m`
6. The first job was distributed between master-node and both 16CPU-workers. The second and third jobs remained in the queue list and not even started after the finish of the first job.
7. Autoscaler stopped working:
image
This state didn't change through time.
8. According to logs, attempts to scale down cluster had been infinitely being performed:
image

tcibinan:
@NShaforostov Regarding the issue you have encountered. From my point of view the described bug is most likely not specific to only this pull request. Probably the problem affects all spot autoscaled clusters in the current develop version. It looks like a separate issue (critical bug) to me. What do you think @SilinPavel?

By the way there is a similar yet not the same issue #494.

NShaforostov:
So, I verified this feature. General totals:

1. Reject job if it doesn't "fit" - works fine
2. `CP_CAP_AUTOSCALE_HYBRID`,  `CP_CAP_AUTOSCALE_HYBRID_FAMILY`, `CP_CAP_AUTOSCALE_HYBRID_MAX_CORE_PER_NODE` env vars - works correctly
3. Jobs with allocation rule like `round_robin` or `fill_up` are processed correctly
4. The deadlock situation is processed (as it was described in issue)

Also, I have the following notes (suggestions) to autoscaler behavior with the "deadlock" situation (they may seem useful):
### Note 1
Imagine the case: autoscale cluster is launched (f.e. master node has 2 CPU and 2 autoscale workers).
1. The first worker was launched with 8CPU and performs something
2. The second worker was launched with 16CPU and performs something
3. To the queue, jobs were added that couldn't be assigned to the first worker
(f.e. `qsub -pe local 12 -b y <cmd>`)
4. The first worker performed his job and becomes "empty"

In the current implementation, the first worker will be run idle until the "light" jobs (that could be assigned to it) will be added to the queue or - the second worker will finish all his jobs and then they both will be scaled down.

**_Suggestion_**: might we should scale down such worker nodes (with less CPUs but empty) to avoid money wastes?

### Note 2
Imagine the case: autoscale cluster is launched (f.e. master node has 2 CPU and 2 autoscale workers).
1. The first worker was launched with 4CPU (_or any CPU count less than or equal to the second worker_) and performs something
2. The second worker was launched with 8CPU and performs something
3. To the queue, jobs were added that couldn't be assigned to any of the exist workers
(f.e. `qsub -pe local 12 -b y <cmd>`)
4. The second worker performed his job and becomes "empty"

In the current implementation, the second worker will be run idle and will not be scaled down. It will be "wait" - until the first (weaker) worker will finish his job and will be scaled down.

**_Suggestion_**: might we should scale down such worker (with more CPUs but empty) to avoid long expectations?

tcibinan:
@NShaforostov The issue that you described in [your first comment](https://github.com/epam/cloud-pipeline/pull/817#issuecomment-583356817) seems to be fixed after #947. The [second comment](https://github.com/epam/cloud-pipeline/pull/817#issuecomment-583525055)'s improvement is pretty good and it has to be a part of this pull request. I'll try to implement that. Thanks.

Issue #822, opened 2019-12-06
Support multithreading for the `pipe-fuse`


sidoruka:
**Background**
At the moment, `pipe-fuse` uses a single-threaded fuse model (e.g. `nothread=True`).
This limits the `pipe-fuse` usage to a very small subset of cases. E.g. we can't replace the `s3fs` or `goofys`, until several processes can interact with the mount point.

**Approach**
* Research the fuse/fusepy threading model and any issues that may be uncovered by that (e.g. broken caches)
* Implement the multithreaded fuse operations for both Object and File backends

sidoruka:
Cherrypick please to `release/0.15`

NShaforostov:
Tried to perform various operations on the Object storage mounted via the `pipe storage mount -t` command:
- `cp/mv` commands from the local fs into the mountpoint directories are being performed well (checked up to 10 threads)
- `cp/mv` commands from the mountpoint directories into the other mountpoint directories are being performed well (checked up to 10 threads)
- `cp/mv` commands from the mountpoint directories into the local fs are being performed well (checked up to 10 threads)

But once during performing `cp` command from the one mounted Object storage to another mounted Object storage the error appeared: `Transport endpoint is not connected`. Logging was disabled, so there are no traces of this behavior. I'll try to reproduce this with `-v` flag.
There is a chance, that at this moment API was redeployed (thus STS token was not retrieved and a pipe-fuse failed)

NShaforostov:
Docs were updated via #860

PR #847, opened 2019-12-16
Enable multithreading for Pipe FUSE

mzueva:
This PR provides implementation for issue #822 .
A new flag `-t/--threads` was added to `pipe storage mount` command.  When this flag is specified Pipe FUSE will start with enabled multithreading feature.

### DIFF
--- a/pipe-cli/mount/pipefuse/pipefs.py
+++ b/pipe-cli/mount/pipefuse/pipefs.py
@@ -22,6 +22,7 @@ import stat
 import time
 
 from fuse import FuseOSError, Operations
+from threading import RLock
 
 import fuseutils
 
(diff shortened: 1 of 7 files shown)


pipe-cli/mount/pipefuse/pipefs.py
```diff
@@ -150,12 +176,14 @@ def statfs(self, path):
...
 
+    @syncronized
     def unlink(self, path):
         self.client.delete(path)
 
     def symlink(self, name, target):
         raise UnsupportedOperationException("symlink")
 
+    @syncronized
     def rename(self, old, new):
```
tcibinan, line 187: During the `rename` operation we lock only on the old path. Most likely it is harmless but probably we can lock both paths. But... it can cause deadlocks if two processes rename same files. What do you think?

Issue #851, opened 2019-12-17
Log user creation and first/last login date and time


sidoruka:
**Background**
At the moment, Cloud Pipeline does not keep the date and time of the users' creation/login. But it's required for traceability

**Approach**
* Log the date/time of the user creation
* Log the date/time of the user's first login
* Expose this information via the corresponding API methods

sidoruka:
Cherry-pick to `release/0.15` once done

NShaforostov:
Verified jointly with #852 
Works well

PR #863, opened 2019-12-18
Save user registration and first login date

SilinPavel:
Related to #851 
Now we save registration date when create new user.
Also first login date will be stored in DB while user perform first login.


mzueva (changes requested):
Could you please review tests, it seems that `NotificationManagerTest` and some others are broken

mzueva (approved):

Issue #853, opened 2019-12-17
Cannot open any directory via FSBrowser in case when the "FSBrowser work directory" is set to the non-default value


NShaforostov:
**Describe the bug**
Cannot open any directory via FSBrowser in the case when the non-default value of the `storage.fsbrowser.wd` variable is set

**To Reproduce**
1. Open System Settings
2. Click the **Preferences** tab
3. Find the `storage.fsbrowser.wd` preference
4. Set any correct value into this preference, e.g. `/home`, save changes and close popup
5. Launch any tool, open the **Run logs** page
6. Wait until the **BROWSER** hyperlink will appear, click it
7. In an appeared directory tree click any folder
8. See an error `[Errno 2] No such file or directory: <path>`

**Expected behavior**
The folder selected at step 7 is opened.

**Screenshots**
![image](https://user-images.githubusercontent.com/45459424/70995492-b71bbc00-20e1-11ea-8f9f-a53fc631e0ad.png)

**Environment:**
 - Cloud Provider: N/A
 - Version 0.16.0.3746
 - Browser: N/A

sidoruka:
This shall be fixed in the `FSBrowser` server part. For the client - WD shall be presented as a filesystem root. Client shall not be aware of the `real` FS structure

NShaforostov:
Verified. Works fine!

PR #866, opened 2019-12-19
FSBrowser doesn't clean the temp files, when download is canceled

ekazachkova:
This PR provides fix for issue #818 

This PR contains the following changes:
- added new `fsbrowser` option `--tmp_directory` to specify temporary directory for downloading folders
- added new preference `storage.fsbrowser.tmp`
- added corresponding environment  variables
- `cancel` method fixed: remove tarball from temporary folder if exists

PR #867, opened 2019-12-19
Cannot open any directory via FSBrowser in case when the "FSBrowser work directory" is set to the non-default value

ekazachkova:
This PR provides fix for issue #853 

`view` method should return path that relative from the working directory


sidoruka:
@NShaforostov same as #866 - waiting for your testing results

NShaforostov:
Verified. Works fine!

Issue #761, opened 2019-10-21
Cloud Pipeline shall allow users to generate spendings reports


NewOlya:
### Background
It would be convenient to implement the feature of global spendings control and management that could include the following capabilities:
- generate and view spending reports (tables and charts)
   - for different time periods (e.g. month/quarter/year, etc.)
   - for different spending sources (e.g. storages, launched compute instances, launched tools, etc.)
   - for different levels (e.g. whole platform, by specific user(s), by specific group(s))
   - with differences depending on the access level (admin/user)
- download spending reports to the local workstation - in the table view (`*.csv` format)
- configure spending quotas
   - for different time periods (e.g. day/month/quarter/year)
   - for different levels (e.g. whole platform, specific user, specific group)
- configure platform actions when these quotas are exceeded (e.g. notify, stop runs, block user)
- configure automatic notifications sending about current spending and approaching to the quotas

### Approach
For these features, the new page shall be implemented, that would be available from the new item of the main Cloud Pipeline menu (e.g. named "**Reports and quotas**"). This page should contain 2 tabs (as it's implemented at the "Runs" page):
- "**Spending reports**" tab, where users can view/generate and download different spending reports according to their access level (_more details will be described in the following comments to this issue_)
- "**Quotas**" tab (invisible for general users), where admins can configure spending quotas, platform actions on quotas exceeding and sending notifications about current spending (_more details are described in the_ #762)

The "Spending reports" tab should be displayed by default.

NewOlya:
For the beginning, the general description of report pages and what they should contain:

#### Types and permissions
Users can view (generate) reports according to their permissions:
- admins - all report types (over the whole platform / specific billing center(s) / specific user group(s) / specific user(s))
- users with special permissions on billing center(s) - reports on available billing center(s) and on actions of users that are related to that billing center(s)
- other users - only reports on their actions

#### Report toolbar
Report toolbar shall contain the following controls:
1. The control that allows to select a current time period for the report (let's call it a "period selector"). Possible values should be:
   - **Month** (by default) - the current month to date. Data should be aggregated by days
   - **Quarter** - the current quarter of the year to date. Data should be aggregated by days
   - **Year** - the current year to date. Data should be aggregated by months
   - **All** - all time period of the platform usage (since the beginning of spending logs). Data should be aggregated by months
   - additional **Calendar** sub-control that allows to select another "identical" time period (not just the current one) according to the specified by the period selector:
[...]

PR #869, opened 2019-12-19
Issue #761 Create API method to pull run activity stats.

Wedds:
This PR related to issue #761. 
To build spending reports we need to pull pipeline run's activity statistic, which is described by run's statuses changes.
For this puprpose new method `loadRunsActivityStats` is introduced in `PipelineRunController`: it returns runs, that caused spending for requested period.
- necessary methods are implemented in `PipelineRunController` and underlaying services
- tests for new functionality are provided
- to reduce amount of transfered data runs' statuses are filtered, so we keep in response only ones, that intersects with requested period

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
@@ -891,7 +891,6 @@ public class PipelineRunManager {
 
         return runs.stream()
             .peek(run -> run.setRunStatuses(runStatuses.get(run.getId())))
-            .filter(run -> CollectionUtils.isNotEmpty(run.getRunStatuses()))
             .collect(Collectors.toList());
     }
 
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
@@ -25,6 +25,7 @@ import com.epam.pipeline.controller.vo.PipelineRunServiceUrlVO;
 import com.epam.pipeline.controller.vo.TagsVO;
 import com.epam.pipeline.dao.pipeline.PipelineRunDao;
 import com.epam.pipeline.entity.AbstractSecuredEntity;
+import com.epam.pipeline.entity.BaseEntity;
 import com.epam.pipeline.entity.cluster.InstancePrice;
 import com.epam.pipeline.entity.cluster.PriceType;
 import com.epam.pipeline.entity.configuration.ExecutionEnvironment;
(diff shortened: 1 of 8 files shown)


api/src/main/resources/dao/pipeline-run-dao.xml
```diff
@@ -1358,6 +1358,62 @@
...
+                        price_per_hour,
+                        state_reason,
+                        non_pause,
+                        node_real_disk,
+                        node_cloud_provider,
+                        tags
+                    FROM
+                        pipeline.pipeline_run r
+                    WHERE
+                        r.status = 2
```
mzueva: I think condition should be the following: latest status from run_statuses (filtered by date) shall be not final

mzueva (changes requested):

api/src/main/resources/dao/pipeline-run-dao.xml
Wedds: Done

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -869,6 +872,58 @@ public PipelineRun updateTags(final Long runId, final TagsVO newTags) {
...
+        final List<Long> runIds = runs.stream()
+            .map(BaseEntity::getId)
+            .collect(Collectors.toList());
+
+        final Map<Long, List<RunStatus>> runStatuses = runStatusManager.loadRunStatus(runIds).entrySet().stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, e -> adjustStatuses(e.getValue(), start, end)));
+
+        return runs.stream()
+            .peek(run -> run.setRunStatuses(runStatuses.get(run.getId())))
+            .filter(run -> CollectionUtils.isNotEmpty(run.getRunStatuses()))
```
mzueva: Do not filter by empty `runStatuses` - for old runs this table is empty.

api/src/test/java/com/epam/pipeline/dao/pipeline/PipelineRunDaoTest.java
```diff
@@ -111,13 +112,16 @@
...
     private static final String TAG_VALUE_1 = "value1";
     private static final String TAG_VALUE_2 = "value2";
+    private static final ZoneId ZONE_ID = ZoneId.systemDefault();
 
     @Autowired
     private PipelineRunDao pipelineRunDao;
 
     @Autowired
-    private FilterDao filterDao;
+    RunStatusDao runStatusDao;
```
mzueva: missing `private`

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
Wedds: Done

api/src/test/java/com/epam/pipeline/dao/pipeline/PipelineRunDaoTest.java
Wedds: Done

mzueva (approved):

Issue #854, opened 2019-12-17
Ability to exclude (hide) specific files/directories from displaying in FSBrowser


NShaforostov:
**Background**
At the moment, Cloud Pipeline displays all node files/directories (according to the specified work directory). But that "full" user access to all node content could be unsafety. In certain cases, it would be convenient for admins to have the ability to hide specific files or directories.

**Approach**
Admins should have the ability to specify which node files/directories shall be hidden for displaying via FSBrowser.
A new system preference should be added into the "Data storage" section to expand the FSBrowser behavior, e.g.: `storage.fsbrowser.black.list`. Which shall behave in the same manner as `storage.mount.black.list`:
- Type: string
- Default value: `/bin,/var,/home,/root,/sbin,/sys,/usr,/boot,/dev,/lib,/proc`
- Validation: can be empty
- Visibility: false

This preference shall control, which paths are "hidden" from the output to the client. Any request to the "hidden" path - shall be restricted.
File(directory) names shall be specified using a comma.

NShaforostov:
@ekazachkova, I tried to verify this, the following issues were found:
1. The feature doesn't work if the specified path is started from spaces. E.g. if the following: `/bin, /var` specify into `storage.fsbrowser.black.list` then `/bin` directory will be excluded but `/var` will be available
2. The feature doesn't work if the specified path is a relative path. E.g. if the following: `/bin,var` specify into `storage.fsbrowser.black.list` then `/bin` directory will be excluded but `/var` will be available. I suggest if the specified path is relative then it should be excluded relatively to the `storage.fsbrowser.wd` value
3. The feature didn't work when I specified the path like `/cloud-data/<storage_name>` (where storage_name is the existing storage name) into `storage.fsbrowser.black.list` (see more details in #882)

ekazachkova:
@NShaforostov  added fixes for points 1 and 2
Could you verify them again?

NShaforostov:
@ekazachkova 
Verified. Points 1 and 2 works fine now

PR #870, opened 2019-12-19
Ability to exclude (hide) specific files/directories from displaying in FSBrowser

ekazachkova:
This PR provides implementation for issue #854 

The following changes was added:
- introduced a new preference `storage.fsbrowser.black.list` with default value `/bin,/var,/home,/root,/sbin,/sys,/usr,/boot,/dev,/lib,/proc`
- introduced a new `fsbrowser` option `--exclude` that expects as input a string with comma separated absolute paths
- black list implemented for `view` operation: user cannot view files/folders from `--exclude` list
- black list implemented for `download` operation: user cannot put on bucket files/folders from `--exclude` list
- black implemented for `upload` operation: user cannot download file from bucket into files/folders from `--exclude` list


NShaforostov:
Tried to verify this, the following issue was found:
- if manually specify any path from `storage.fsbrowser.black.list` into the address bar - it will be "opened" as an empty directory, e.g.:
image
In this case the message `No such file or directory` should be displayed (as for any other nonexisting path)
- if manually specify a subfolder (existing or not) to any path from `storage.fsbrowser.black.list` into the address bar - it will be also "opened" as an empty directory, e.g.:
image
In this case the message `No such file or directory` should be also displayed

All paths from `storage.fsbrowser.black.list` should be fully invisible (as if they did not exist) for users

NShaforostov:
Verified. Works fine!

Issue #880, opened 2019-12-23
Autoscaler shall not kill the "Lost" nodes immediately


sidoruka:
# Background

From time to time we run into cases when a compute node gets unresponsive (kubelet can't send the heartbeats) due to the high CPU/Memory usage.
If such a node is detected - the run is immediately stopped as the node's status is "Unknown".

While such an oversubscription is not a good usage pattern - this may lead to the data/compute results loss. Thus, we shall try to handle such situations.

# Approach

1. If a node gets into "Unknown" state - we shall not kill the run right away
2. We shall introduce a preference, that shall control the timeout period before stopping a run

**Note:** *Before the implementation itself - please verify if the node/pod can survive the resources starvation (e.g. if the node's state is unknown - is the pod deleted or it can proceed running, once a node is restored)*

SilinPavel:
@sidoruka 
As I find out in process of investigation, a pod can be alive during resource starvation if we disable eviction strategy, actually I have had 3 different scenarios (resource starvation was modeled by artificial stress test):
1.  When Node starts to be in a "starvation" NodeStatus become  NotReady, but pod still will be in Running state, after resources clean up all statuses will go back to normal.
2. If we will simulate starvation a little longer (around 2 minutes in my case), NodeStatus will have value = unknown, pod still will be in Running state, after resources clean up all statuses will go back to normal.
3. If a starvation will be simulated much longer(5-7 minutes in my case) we will have pod in unknown state as well, in this case after resource clean up node still can go back to normal state, but pod will be terminated.

So I believe that first and second scenarios is that we are looking for, and in this case our pipeline can be back in a normal state after small resource starvation.

SilinPavel:
In addition I want to add some details related to this issue.
It was found out that we can change setting for node-controller-manager system pod, to be able to increase period of time before pipeline pod will be terminated in case (3). According to documentation (https://kubernetes.io/docs/concepts/architecture/nodes/#node-condition):
```
If the Status of the Ready condition remains Unknown or False for longer than the pod-eviction-timeout, 
an argument is passed to the kube-controller-manager and all the Pods on the node are scheduled for 
deletion by the Node Controller. The default eviction timeout duration is five minutes. In some cases 
when the node is unreachable, the apiserver is unable to communicate with the kubelet on the node. 
The decision to delete the pods cannot be communicated to the kubelet until communication with the 
apiserver is re-established. In the meantime, the pods that are scheduled for deletion may continue 
to run on the partitioned node.

[...]

PR #905, opened 2020-01-22
Issue 880 wait for lost node

SilinPavel:
This PR provides implementation of #880 
It makes some changes to kube configuration, according to discussion in #880 , it also provides implementation of approach of deferred failing of a pipeline that is run on a lost node.
Accordingly to this approach:
- New system preference was introduced: `cluster.node.lost.tolerant.seconds`
-  `AutoscaleManager` will store `timestamp` when node was lost for the first time
- Each next time it will check that 'timeout' (`cluster.node.lost.tolerant.seconds`) not exhausted 
- If 'timeout' will be exhausted pipeline of this node will be failed
- If lost node will have enough time to recover nothing will happen, and pipeline will continue to work, timestamp of last loss will be reset


api/src/main/java/com/epam/pipeline/manager/cluster/AutoscaleManager.java
```diff
@@ -515,20 +519,34 @@ private boolean isNodeAvailable(final KubernetesClient client, final String node
...
+                    .computeIfAbsent(node.getMetadata().getName(), id -> nowUTC);
+
+            if (nodeLostTimestamp.plusSeconds(tolerantSeconds).isBefore(nowUTC)) {
                 LOGGER.debug("Node is out of order: {}", conditions);
                 return false;
             }
+        } else {
+            // clean up since we check that here is no reasons from NODE_OUT_OF_ORDER_REASONS
+            // so we need to reset "node lost" timestamp
+            nodeLostTimestamps.remove(node.getMetadata().getName());
```
mzueva, line 558: Nodes may also be terminated using API,  in this case nodeLostTimestamps won't be cleared forever.

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/cluster/AutoscaleManager.java
SilinPavel: I have add cleanup process for nodeLostTimestamps map each step of `Autoscaler` loop

mzueva (approved):

Issue #837, opened 2019-12-13
Implement compute nodes' disks/swap autoscaling


sidoruka:
**Background**
At the moment, Cloud Pipeline allows setting the compute nodes' local disk size at a job startup time.
For certain use cases, users may fill the disk and get the out of disk errors, leading to job failure. 
The same issue is observed for the out of memory errors. While do have the SWAP support, it is configured as a static value at a job start time.

To address those issues, we shall think of automatic scaling (up) of the filesystem/swap space, according to workload demand.

**Approach**
1. Implement API methods:
  * Increase the disk by a specified delta for the specified run (shall be supported for `AWS`/`GCP`, for `Azure` it shall throw `operation not supported` error)
  * Attach a new disk  of a specified size to the specified run (shall be supported for `AWS`/`GCP`/`Azure`)
2. Add a number of preferences to control the scaling behavior:
  * `cluster.instance.hdd.scale.threshold.ratio` - float, default: `0.75`
  * `cluster.instance.hdd.scale.delta.ratio` - float, default: `0.5`
  * `cluster.instance.swap.scale.threshold.ratio` - float, default: `0.75`
3. Implement a daemonset that will:
  * Monitor compute node's disk/swap
  * If a threshold (`cluster.instance.hdd.scale.threshold.ratio`) is reached for the FS disk - increase it's size and resize the FS (by `cluster.instance.hdd.scale.delta.ratio` times)
  * If a threshold is reached for the SWAP disk - attach a new disk (of the same size as the existing swap volume) and enable it for swapping

tcibinan:
Pull request #855 was cherry-picked to `release/0.15` via 6b27ba22b878027c37fc0679f4be7ef48a078f35.

tcibinan:
Pull request #913 was cherry-picked to `release/0.15` via 4453ee5841edbb5904481f81b824ae363d00271d.

PR #913, opened 2020-01-23
Add support for autoscaling filesystem for AWS

tcibinan:
## Description

Relates to #837.

The pull request brings support for autoscaling filesystems. *Currently only AWS provider is supported.*

Pull request #855 brought support for additional disks attaching to Cloud Pipeline. Since then it is possible to use several disks and some scalable filesystem to automate run container filesystems autoscaling.

### Btrfs

Filesystem that is used in the current implementation uses **btrfs** (B-tree FS or better fs) which is [compatible with docker](https://docs.docker.com/storage/storagedriver/btrfs-driver/). It can be initialized with a single device and new devices can be attached after bringing more available disk space.

The new **btrfs** filesystem replaces the original **ext4** filesystem which was used for run main disks (not the root ones). Moreover the new **btrfs** filesystem replaces the original **overlay2** filesystem that docker used as a container root filesystem.

### /etc/fstab

From now on all disk will be defined in `/etc/fstab` using their **UUID**s rather then device names which can vary from boot to boot.

### AMIs

The pull request makes changes to all cloud amis adding a `btrfs-progs` and `jq` dependencies which bring support for btrfs filesystem and json parsing.

### User data

User data self-deployment script was updated. From now on additional files can be embedded to the user data along with `init_script`. To specify and additional embedded file one can update `cluster.networks.config` preference with the corresponding entry.

### Filesystem autoscaler

Filesystem autoscaler introduced in the pull request launches once on the node startup / node boot as a `fsautoscale` systemd service. Every 10 seconds the daemon checks if there is a corresponding run and if the corresponding run has file system autoscaling capability being enabled. Otherwise it does nothing.

For all runs with the file system autoscaling capability being enabled the daemon checks current filesystem usage and attach additional disks in case there is no enough space in the filesystem.

There are several system preferences that can be used to control filesystem autoscaler behavior:

- `cluster.instance.hdd.scale.enabled` specifies if a filesystem autoscaling capabilities are enabled for all runs or not. Disabled by default.
- `cluster.instance.hdd.scale.monitoring.delay` specifies filesystem monitoring delay in seconds. Defaults to 10 seconds.
- `cluster.instance.hdd.scale.threshold.ratio` specifies an allowed filesystem usage ratio which defaults to 0.75. It means that if filesystem usage is above 75% then it will be autoscaled.
- `cluster.instance.hdd.scale.delta.ratio` specifies a multiplier for a filesystem autoscaling which defaults to 0.5. It means that filesystem will be autoscaled by 50%, f.e. 50G filesystem will be scaled up to 75G.
- `cluster.instance.hdd.scale.max.devices` specifies a max number of devices (disks) for a single filesystem. Defaults to 40.
- `cluster.instance.hdd.scale.max.size` specifies filesystem max size in GB. Defaults to 16384 GB.
- `cluster.instance.hdd.scale.disk.min.size` specifies additional disks min size in GB. Defaults to 10 GB.
- `cluster.instance.hdd.scale.disk.max.size` specifies additional disks max size in GB. Defaults to 16384 GB.

**Notice that all the preferences specified above affect all existing runs including previously launched ones. Use with caution.**

## Limitations

Nevertheless the current implementation has some limitations that can affect usage of btrfs in practice. According to the same [docker docs](https://docs.docker.com/storage/storagedriver/btrfs-driver/) linked above only Ubuntu and Debian distributions are expected to be used with btrfs.

## Testing

The pull request changes were tested in several use cases including:
- aws and azure (autoscaling is not available) deployments,
- `ubuntu`, `centos` and `debian` run filesystems autoscaling,
- pausing and resuming runs,
- committing runs,
- dind runs.

Since the pull request changes both backing and container root filesystems and we've [previously faced](https://github.com/epam/cloud-pipeline/issues/291#issuecomment-498174114) problems regarding such filesystems it is required to test the new filesystems behaviour using production-like workloads. In order to do so RNASeq nextflow pipeline was used to test out disks stability.


tcibinan:
Just linking 39efe230dbe431128857fb6936a5d9d22d30f65c that backports filesystem autoscaling to `release/0.15`.

sidoruka (approved):
Need to double-check PAUSE/RESUME operations once merged and deployed into DEV

tcibinan:
Checked that `PAUSE/RESUME` and `COMMIT` works in version `0.16.0.4715.6d66122b4865148cccb4f14e6b8da54d8233d49f`.

tcibinan:
It turned out that `0.16` release was affected too but with different resources monitoring backend. See the following [comment](https://github.com/epam/cloud-pipeline/issues/998#issuecomment-598633089) for more info.

Issue #904, opened 2020-01-20
S3 bucket shall be created with broken Multipart Upload cleanup policy


sidoruka:
**Background**

At the moment, if the data upload to the S3 buckets is terminated (including docker pushes) - an incomplete multipart upload is left. This consumes the storage volumes and introduces extra charges.

**Approach**

* Research how this behaves across the other cloud providers
* Introduce a preference `storage.incomplete.upload.clean.days` (int, nullable, default: 5)
* If the `storage.incomplete.upload.clean` is enabled - all the object storages shall be created with the corresponding policy (in addition to the other policies), e.g.:
```
{
    "Rules": [
        {
            "ID": "Clean Incomplete Multipart Uploads",
            "Filter": {
                "Prefix": ""
            },
            "Status": "Enabled",
            "AbortIncompleteMultipartUpload": {
                "DaysAfterInitiation": <storage.incomplete.upload.clean.days>
            }
        }
    ]
}
```

sidoruka:
Shall be backported to `release/0.15`

PR #920, opened 2020-01-27
Issue #904 Apply AbortIncompleteMultipartUpload rule for S3 buckets

SilinPavel:
This PR related to #904 
Now AbortIncompleteMultipartUpload will be automatically added to s3 bucket life cycle policy, while bucket creation time.


api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -311,6 +312,16 @@ public AbstractDataStorage updatePolicy(DataStorageVO dataStorageVO) {
         return createdStorage;
     }
 
+    private StoragePolicy applySpecialRulesToStoragePolicy(final StoragePolicy policy) {
+        final Integer incompleteUploadCleanupDays = preferenceManager
```
mzueva: It seems to me that `policy` maybe nullable here?

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
SilinPavel: right, now it is fixed

mzueva (approved):

Issue #919, opened 2020-01-27
Administrator shall be able to restrict which run statuses trigger the email notification


sidoruka:
**Background**
At the moment, Cloud Pipeline can be configured to send the email to the owner of the job if its status is changed.
In certain cases - it is not desired to get a notification in certain cases (e.g. if the run is started). 
To reduce a number of the emails we shall allow to configure, which statuses are triggering the email notifications.

**Approach**
1. `PIPELINE_RUN_STATUS` shall expose an additional option, which shall accept a list of possible run's statuses
2. The email notifications shall be sent only if the run enters one of the specified states
3. GUI shall provide a separate control to manage this as well

sidoruka:
@rodichenko please take care of (3) once backend is ready

NShaforostov:
Docs were updated via #991

NShaforostov:
Verified on GCP stand

PR #925, opened 2020-01-29
Issue 919 send notification only for certain statuses

SilinPavel:
This PR related to #919 
It proposes an implementation of that issue, to send notification only in case when status of a run has changed to only specific ones


mzueva (approved):

api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
```diff
@@ -206,6 +208,14 @@ public void notifyRunStatusChanged(PipelineRun pipelineRun) {
             return;
         }
 
+        final List<TaskStatus> runStatusesToReport = ListUtils.emptyIfNull(runStatusSettings.getStatusesToInform());
+        if (CollectionUtils.isEmpty(runStatusesToReport) || !runStatusesToReport.contains(pipelineRun.getStatus())) {
```
mzueva: If statuses are not defined we shall proceed with notification

mzueva (changes requested):

mzueva (approved):

Issue #865, opened 2019-12-19
SLURM support for the static clusters


sidoruka:
**Background**
At the moment, Cloud Pipeline allows creating the clusters, that are managed by the 3rd party schedulers:
* GridEngine
* Spark

This feature is mostly used for backward compatibility with the existing code-base (which is a super common usage scenario).

To extend this compatibility - we need to implement [SLURM scheduler](https://slurm.schedmd.com/documentation.html) support as well

**Approach**
* Overall approach shall be the same as for the GridEngine implementation
* Only static cluster shall be implemented for now. Autoscaling feature shall be added further

sidoruka:
@rodichenko you can proceed with the GUI part (use the same PR/branch please)

sidoruka:
@NShaforostov, ready for the tests and docs

NShaforostov:
I've tested this feature. Generally, it works.
But the following issues were identified:
1) After performing the simple script via the `sbatch` command over several nodes - the first node in the allocation (where the script was run) gets the `drained` state and remains with it. Also, such `sbatch` ends with the error.
   More details:
   - I used the following script (from [here](https://slurm.schedmd.com/quickstart.html))
   ![image](https://user-images.githubusercontent.com/45459424/73361292-b9169480-42b5-11ea-8c7d-e398afa72ef6.png)
   - The batch command was: `sbatch -N3 -o test.stdout test.script`
   - The nodes state during the job performing:
   ![image](https://user-images.githubusercontent.com/45459424/73361450-06930180-42b6-11ea-807d-6940e40619f6.png)
   - The nodes state after the job finish:
   ![image](https://user-images.githubusercontent.com/45459424/73361634-5f629a00-42b6-11ea-8378-93394c03c98b.png)
   - The output that contains the error:
   ![image](https://user-images.githubusercontent.com/45459424/73362030-2545c800-42b7-11ea-9e1a-3ce3bbc4cd51.png)
2) I tried to perform the script specified above on nodes without the master-node. In the result, the first node in the allocation (where the script was run) also got the `drained` state and remained with it. But in that case, the output contained also other errors:
   - The batch command was: `sbatch -N2 -w "pipeline-[21358-21359]" -o test.stdout test.script` (pipeline-21357 - master-node)
   - The output:
   ![image](https://user-images.githubusercontent.com/45459424/73364121-0d704300-42bb-11ea-9b1c-a9e645121b34.png)

Such behavior was observed for both - `Debian` and `Centos` images

[...]

PR #937, opened 2020-02-03
(issue #865) fixes for slurm intergation

SilinPavel:
The following problems are solved by this PR

- Unable to run slurm jobs from worker hosts (NodeAddr was added for each node to be able to recognize node ip from worker node)

- Strange error `error: *** JOB X STEPD TERMINATED ON xxxxxxx AT xxxxxxx DUE TO JOB NOT ENDING WITH SIGNALS ***` (change `ProctrackType` to cgroup according to (https://slurm.schedmd.com/slurm.conf.html)


NShaforostov:
Verified.
Works well.
Was tested for both - `Debian` and `Centos` images

PR #957, opened 2020-02-14
Reimplement truncate and support fallocate pipe fuse operations

### LINKED ISSUE #821: Support fallocate and ftruncate for the pipe-fuse
**Background**
At the moment, any `ftruncate` or `fallocate` operation fail on the directories, mounted by `pipe storage mount`.

But this is a quite common use case, especially when dealing with data transfer via `rsync`.

The approach is described in the PR https://github.com/epam/cloud-pipeline/pull/661#issuecomment-538399907

tcibinan:
Resolves issue #821.

The pull request consists of two parts. Firstly it updates *truncate* operation implementation to support efficient file length increasing and decreasing for both webdav and s3 backends. Secondly the pull request brings support for *fallocate* operation for most of the modern distribution.

## Truncate

Previously the truncate operation worked only if the target file length was 0. From now on truncate operation can be used in order to increase or decrease file length.

## Fallocate

Currently *fallocate* operation support includes only the basic command behavior. It can be used only to either verify or increase file length. All the allowed fallocate [modes](http://man7.org/linux/man-pages/man2/fallocate.2.html) are not supported yet.

## Libfuse

From now pipe mount executables are bundled with either libfuse version *2.9.2* or libfuse version *2.8.3*. All the linux distributions that use common pipe executable will use the newer libfuse version. Only the linux distibutions that use so-called *el6* pipe executable will use the same libfuse version as previously.

Notice that *fallocate* operation is supported only in the newer libfuse version. Therefore linux distributions that use *el6* pipe executable won't have support for *fallocate* operation.

Also notice that if pipe is installed as a python package rather than a pre-built executable then the host machine libfuse version will be used.

## Testing

Fallocate and truncate testing were performed for a variety of platforms listed below. Some of the platforms such as centos:6 use older libfuse version and therefore fallocate operation is not support ed there.

| distribution | version | libfuse version | truncate | fallocate | 
| --- | --- | --- | --- | --- |
| ubuntu | 16.04 | 2.9.2 | supported | supported |
| ubuntu | 18.04 | 2.9.2 | supported | supported |
| ubuntu | 19.04 | 2.9.2 | supported | supported |
| centos | 6 | 2.8.3 | supported | not supported |
| centos | 7 | 2.9.2 | supported | supported |
| debian | 8 | 2.9.2 | supported | supported |
| debian | 9 | 2.9.2 | supported | supported |


NShaforostov:
Verified at `Ubuntu` (16.04, 18.04, 19.04), `Centos` (6, 7, 8), `Debian` (8, 9).
All works well!

Issue #548, opened 2019-07-23
Implement storage usage statistics retrieval via pipe


sidoruka:
Expose `pipe storage du` command which shall provide "disk usage" information on the supported data storages/path:
* Storage/Path name
* Number of files in the storage/path
* Size of the files in the storage/path

The following use cases shall be supported:

## All storages
```
$ pipe storage du
Storage    Files count    Size(Gb)
bucket1    121            17
bucket2    6227           218
bucket3    1              8
```

## Specific storage
```
$ pipe storage du s3://bucket1
Storage    Files count    Size(Gb)
bucket1    121            17
```

## Storage path
```
$ pipe storage du s3://bucket1/path/to/some/files
Storage                       Files count    Size(Gb)
bucket1/path/to/some/files    102            12
```

## Size formatting (-f|--format G/M/K)
```
$ pipe storage du s3://bucket1 -f M
Storage    Files count    Size(Mb)
bucket1    121            17408
```

## du depth (-d|--depth)
```
$ pipe storage du s3://bucket1 -d 3
Storage                 Files count    Size(Gb)
bucket1/path/to/some    102            12
bucket1/path/to/other   2              1
bucket1/path/to/more    1              2
bucket1/path/to/less    4              1
```

NShaforostov:
I've tried to verify this feature. In general, it works.
But some issues were detected:
1) `pipe storage du` (with or without any options) is being failed for NFS storages with the following error message: `Failed to load usage for datastorage <NFS_storage_name>. Cause: Failed to fetch data from server. Server responded with message: Elasticsearch exception [type=index_not_found_exception, reason=no such index]`
2) Can we implement the output of a warning message when the user specifies a non-existing path with `-p` option?
3) Can we implement the sorting of the command output when it's used with `-d` option? It would be clearer for the user.
4) Imagine the situation: there are two folders in the storage - `folder` and `folder2`. Each folder contains 2 files.
In that case, the output of the command `pipe storage du <datastorage> -p folder` will be:
```
Storage                 Files count      Size (Mb)
<datastorage>/folder    4                ...
```
But for the command `pipe storage du <datastorage> -p folder/` will be:
```
Storage                 Files count      Size (Mb)
<datastorage>/folder/   2                ...
```
Probably, we shouldn't use the inputted path with `-p` option as a "mask" but only as the specific path to the directory?

NShaforostov:
Docs were updated via #991

PR #979, opened 2020-02-21
pipe storage du implementation

ekazachkova:
This PR provides implementation for issue #548 

- a new API method `GET datastorage/path/usage` that calculates datastorage usage statistics. The data is taken from elasticsearch.
- method for loading NFS datastorages by name or ID was fixed.
- a new CLI option `pipe storage du` was added. If `path` is not specified all buckets will be listed with `GET datastorage/path/usage` API method call. If path was specified and it's not NFS the bucket will be listed with cloud provider directly. If path has NFS type the data will be loaded with `GET datastorage/path/usage` API method call

The command  `pipe storage du` has the following options:
- `name` - the data storage name [optional]
- `-p (--relative-path)` - the relative path [optional]
- `-f (--format)` - the size unit format [default: Mb]
- `-d (--depth)` - the maximum depth level [optional]

Usage examples:
- `pipe storage du` - lists all storages usage
```
  Storage                 Files count     Size (Mb)  
  storage_1                    26863          357.7  
  storage_2                        7            0.1  
```
- `pipe storage du storage_1` - prints usage for required storage
```
  Storage                              Files count     Size (Mb)  
  storage_1                                  26863         357.7  
```
- `pipe storage du storage_1 -p common` - prints usage for required storage and path
```
  Storage                              Files count     Size (Mb)  
  storage_1/common                           26863         357.7  
```
- `pipe storage du storage_1 -p common -d 1` - prints usage for required storage and path for depth level 1
```
  Storage                              Files count     Size (Mb)  
  storage_1/common/folder1                   26863         357.7  
  storage_1/common/folder2                       7           0.1  
```
- `pipe storage du storage_1 -f G` - prints usage for required storage with specified size format
```
  Storage                              Files count     Size (Gb)  
  storage_1                                  26863           2.2 
```

Note: `--depth` option is not supported for NFS storages now


pipe-cli/pipe.py
```diff
@@ -901,6 +902,16 @@ def storage_copy_item(source, destination, recursive, force, exclude, include, q
                              exclude, include, quiet, tags, file_list, symlinks, skip_existing=skip_existing)
 
 
+@storage.command('du')
+@click.argument('path', required=False)
+@click.option('-f', '--format', help='Format for size [G/M/K]',
+              type=click.Choice(DuFormatType.possible_types()), required=False, default='M')
+@click.option('-d', '--depth', help='Depth level', type=int, required=False)
+@Config.validate_access_token(quiet_flag_property_name='quiet')
+def storage_copy_item(path, format, depth):
```
mzueva: It seems that method name is misleading

pipe-cli/src/utilities/storage/s3.py
```diff
@@ -411,6 +414,83 @@ def list_items(self, relative_path=None, recursive=False, page_size=StorageOpera
...
+            if 'Contents' in page:
+                for file in page['Contents']:
+                    name = self.get_file_name(file, prefix, True)
+                    size = file['Size']
+
+                    tokens = name.split(delimiter)
+                    root.data.add_item(size)
+
+                    if len(tokens) - relative_path_len > 1:
+                        if relative_path:
```
mzueva: Could we extract this logic into a separate method to make it more readable?

api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
```diff
@@ -412,6 +412,15 @@ public boolean storagePermission(Long storageId, String permissionName) {
         }
     }
 
+    public boolean storagePermissionByName(final String identifier, final String permissionName) {
+        AbstractSecuredEntity storage = entityManager.loadByNameOrId(AclClass.DATA_STORAGE, identifier);
```
mzueva: missing `final`

mzueva (changes requested):

mzueva:
@ekazachkova Could you please add nfs storage path format to the PR description?

api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
mzueva: still missing `final`

mzueva (changes requested):

ekazachkova:
Several remarks for this implementation:
1. It would be better to specify storage path with single option, e.g. `pipe storage du s3://storage_1/folder1`. But in this case it's not clear how to parse NFS storage path since it is difficult to determine where the storage name and where relative path. 
The idea is to implement NFS path parsing into the server part:
- determine deepest folder path from specified path
- try to load storage by this name
- if the storage does not exists - level up folder path and try to load storage by new name
- repeat until the storage is found
2. `--depth` option is not supported for NFS storages. One of the possible ways to solve this problem is to use elasticsearch. The idea is to store depth information for paths.

Issue #983, opened 2020-02-27
Pass proxy settings to the DIND containers


sidoruka:
**Background**
At the moment, DIND containers configuration includes only registry credentials and a couple of driver settings.
In certain environments, it is not possible to access external networks (e.g. for the packages installation) without the `http(s)/no_proxy` settings.
So the users have to pass this manually using `docker run -e http_proxy=...` option.

**Approach**
* Introduce a system preference `launch.dind.container.vars` (nullable, string) listing all the variables, which shall be passed to the DIND containers (if they are set for the host environment)
* By default, the following variables shall be set for the `launch.dind.container.vars` preference: `http_proxy`,`https_proxy`,`no_proxy`,`API`,`API_TOKEN`
* Introduce a system parameter (per run) `CP_CAP_DIND_CONTAINER_NO_VARS`, which disables this behavior

sidoruka:
Backport to `release/0.15`

tcibinan:
Backported to `release/0.15` via 476f521aed3f308e1e8f30febae341385f32fe8c.

tcibinan:
It will be also nice to have the same proxy parameters to be passed automatically during `docker build` call as well as `docker run`.

NShaforostov:
Verified.
Docs were updated via the #991

PR #987, opened 2020-02-28
Add environment variables inheritance docker wrapper

tcibinan:
Resolves issue #983.

The following pull request introduces docker wrapper for runs with DIND capability being enabled. The wrapper is implemented just the same way it is done for #158 and #500.

A system preference and a system parameter are introduced by the pull request:

- `launch.dind.container.vars` system preference can be used to specify a comma-separated list of environment variables that will be automatically set for all launched DIND containers in all runs. Environment variable values will be inherited from the run's container.
- `CP_CAP_DIND_CONTAINER_NO_VARS` system parameter can be used for a single run to disable this environment variables inheritance capability.

Just as a neat addition `pipe_get_preference` script is added to pipe-common in order to simplify system preferences retrieving.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
+++ b/api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
@@ -345,6 +345,13 @@ public class SystemPreferences {
     public static final ObjectPreference<List<DockerMount>> DOCKER_IN_DOCKER_MOUNTS = new ObjectPreference<>(
             "launch.dind.mounts", null, new TypeReference<List<DockerMount>>() {},
             LAUNCH_GROUP, isNullOrValidJson(new TypeReference<List<DockerMount>>() {}));
+    /**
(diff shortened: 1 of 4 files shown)


sidoruka (changes requested):
@tcibinan please double-check what happens in the following cases:
```
echo 'http_proxy=http://proxy:port' > ~/my.env
docker run -e http_proxy="http://proxy:port" image:tag bash
docker run -env ~/my.env image:tag bash
```

Are the variables merged? If so, in what order?
Keep in mind: we shall never overwrite user's inputs

tcibinan:
@sidoruka Thanks for raising the concern. The wrapper appends additional `--env-file` argument right after `run` keyword and all the user specified `-e\--env-file` arguments goes after. The [documentation](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file) is not clear on the priority of such arguments. Nevertheless according to the manual testing it seems that the user specified environment variables always overrides wrapper ones because of the arguments order.

sidoruka (approved):

Issue #1030, opened 2020-04-03
Extend `pipe storage mount` logging


sidoruka:
**Background**
At the moment, it's almost impossible to trace any issue, that may happen during the `pipe storage mount ...` lifetime.
E.g. If it was working for some time and, eventually, mountpoint becomes unavailable - we don't know how to troubleshoot.

**Approach**
We shall extend logs and (optionally) print any event, that occurs within an application.

sidoruka:
Shall be backported to `release/0.15`

tcibinan:
Cherry-picked to `release/0.15` via db1022ffbc056a71c869ba2690db58481ecade3a.

PR #1032, opened 2020-04-06
Add log level and custom option pipe mount arguments

tcibinan:
Relates to #1030.

The pull request improves pipe fuse integration in several ways in order to simplify troubleshooting.

## Logging level

The pull request brings support for an additional `pipe mount` argument `--log-level` which can be used to specify python [logging level](https://docs.python.org/2.7/howto/logging.html) which defaults to `ERROR`.

```bash
pipe storage mount --file \
                   --log-file log.txt \
                   --log-level INFO \
                   mountpoint
```

For the simplest cases `INFO` logging level should be enough. It registers all the file system events but `write` and `read` calls which are available only in `DEBUG` logging level mode. The `DEBUG` logging level collects all the available logs which can results in very large logs file especially if `threading` is enabled. Moreover pipe fuse performance is significantly worse in case of `DEBUG` logging level.

## Libfuse logging

There is a possibility to enable libfuse low level logging with fuse `debug` option. The pull request doesn't introduce the feature but it seems fine to point it out along other logging configurations.

```bash
pipe storage mount --file \
                   --options debug \
                   mountpoint
```

## Custom options

Moreover the pull request brings an additional `--custom-options` argument which can be used to fine tune pipe fuse behavior. All the custom options are transferred to pipe fuse command line arguments.

```bash
pipe storage mount --file \
                   --custom-options buffer-size=10000,cache-size=10 \
                   mountpoint
```

### DIFF
--- a/pipe-cli/mount/pipe-fuse.py
+++ b/pipe-cli/mount/pipe-fuse.py
@@ -190,6 +190,7 @@ if __name__ == '__main__':
     recording = args.logging_level == _debug_logging_level
     logging.basicConfig(format='[%(levelname)s] %(asctime)s %(filename)s - %(message)s',
                         level=_allowed_logging_level_names[args.logging_level])
+    logging.getLogger('botocore').setLevel(logging.ERROR)
 
     if is_frozen:
         logging.info('Frozen installation found. Bundled libfuse will be used.')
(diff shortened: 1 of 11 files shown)


mzueva (approved):
I'd suggest to refactor `mount` methods in pipe cli, but in a separate branch

NShaforostov:
I've tried to verify. Works well.

Minor note: might be, it would be useful to add possible values of log-levels into the help of the `storage mount` command?

tcibinan:
@NShaforostov Thanks for checking it out. Added available logging levels to `pipe storage mount` command's help message. Also cleaned up the help message a bit.

Issue #989, opened 2020-03-02
Implement central storage system for log data


SilinPavel:
In additional to #684, this issue is for proposing and implementing solution for storing and managing any type of log data from Cloud-Pipeline system (Security logs, pipeline node logs, etc)

The main aim of this issue will be implementing whole log storage system.
Acceptance criteria:

- Concept of log storage system is provided 
  - where to store logs
  - which tools to use
  - structure of log entry
  - how to search and filter logs
  - how to configure log rotation
  - how to integrate and deploy it with Cloud-Pipeline
- Solution of log storage system is provided

Solution should include the next parts of a system:
- Storage point (f.e. elasticsearch instance)
- Mutation and normalization point (f.e. Logstash)
- Scripts and configurations to deploy this solution

As additional acceptance criteria it is proposed to implement logging of the next `Security` events within the new system:
  -   Successful user authentication
  -   Failed user authentication
  -   User profile/rights modification
  -   Applications accessed

SilinPavel:
@sidoruka  @mzueva please feel free to express any other thoughts and concerns about this issue

SilinPavel:
# Concept of log storage system

Here is a diagram that represent an interaction between components in a system:

![Untitled Diagram](https://user-images.githubusercontent.com/11179595/75783391-8dce0c00-5d71-11ea-8556-7d956cf30727.png)

## The next technology stack is proposed to use:
- Elasticsearch as a storage to persist and manage log data
- Logstash as a mutation point, here all filtering and log mapping could be performed 
- Kibana to visualize and inspect logs for developer needs
- Filebeat (?) / Our solution (daemon script)  to send logs from services to logstash

## Functional artifacts:
- Deployment configurations for new ELK, Logstash, and Kibana including configuration of elk index lifecycle (rotation, backup etc)
- Daemon script to send different types of logs to Logstash
- Cloud-Pipeline API to filter, search and load logs to be able to inspect logs by admins of the system on our UI

## Log structure
To be able to effectively filter and search through the logs we need to have non-plain structure of it, this approach will allow us to split log data by different indexes and also build filter query to fetch only data that we interested in.
A structure of the log entry is proposed to be as following:
```
{
    "service" : "<Name of the service that sends a log entry, f.e. api/edge/pipeline_container>",
    "service-id": "<Optional identifier of a service, f.e. for pipeline_container - run_id>",
    "timestamp" : "<timestamp of the event>",
    "from" : "<address of a reported service>",
[...]

PR #1043, opened 2020-04-08
Issue 989 log filter endpoint. Part 3

SilinPavel:
The third part of #989. 
This part bring new functionality to API server, now all admins can view and search system logs


api/src/main/java/com/epam/pipeline/manager/log/LogManager.java
```diff
@@ -0,0 +1,269 @@
...
+        final LogPaginationRequest pagination = logFilter.getPagination();
+
+        Assert.notNull(pagination, messageHelper.getMessage(MessageConstants.ERROR_PAGINATION_IS_NOT_PROVIDED));
+        Assert.isTrue(pagination.getPageSize() > 0,
+                messageHelper.getMessage(MessageConstants.ERROR_INVALID_PAGE_INDEX_OR_SIZE));
+
+        final SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
+                .query(constructQueryFilter(logFilter))
+                .sort(MESSAGE_TIMESTAMP, SortOrder.ASC)
+                .sort(ID, SortOrder.ASC)
```
mzueva, line 113: As far as I remember sorting by id is not recommended due to performance considerations: "The _id field has a unique value per document but it is not recommended to use it as a tiebreaker directly" https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-search-after.html.  
I think sorting by timestamp shall be enough?

api/src/main/java/com/epam/pipeline/manager/log/LogManager.java
```diff
@@ -0,0 +1,269 @@
...
+
+        return LogPagination.builder()
+                .logEntries(entries.stream().limit(pagination.getPageSize()).collect(Collectors.toList()))
+                .pageSize(pagination.getPageSize())
+                .token(getToken(entries, pagination.getPageSize()))
+                .totalHits(hits.totalHits)
+                .build();
+    }
+
+    private PageMarker getToken(List<LogEntry> items, int pageSize) {
```
mzueva: Please, check usage of `final` here and below

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/log/LogManager.java
SilinPavel: Two fields is used here to have unique value from pair. Here I try to resolve problem when two messages can be acquired at the same time. But indeed we should get rid of using `_id` as  tiebreaker, now ingest pipeline creates second field `event_id` to use it instead of `_id`

PR #1078, opened 2020-04-28
Issue 1060 SMB support for web dav

SilinPavel:
This PR adds new functionality for mounting SMB share storages to webDAV Service.

Because of specific mount prerequisites (like azure storage creds) for mounting, not only WebDAV scripts are changed, but additional changes on deployment process are needed.

To be able to pass storage creds to WebDAV additional kube secret 'cp-region-creds-secret' was added. This secret is used to store all azure region credentials for datastorage accounts.

Each time region is created, api adds storage creds for this region to this kube secret.
WebDAV service mounts this secret as folder and can access creds file for mounting purpose. 

NOTE.
While redeploy existing CP instance with this changes, please remember that you need to create  'cp-region-creds-secret' manually, (when CP instance is created from scratch it will create this secret automatically by pipectl)

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/KubernetesManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/KubernetesManager.java
@@ -156,7 +156,7 @@ public class KubernetesManager {
 
     public String refreshSecret(final String secretName, final Map<String, String> data) {
         Secret secret;
-        try (final KubernetesClient client = getKubernetesClient()) {
+        try (KubernetesClient client = getKubernetesClient()) {
             secret = client.secrets()
                     .inNamespace(kubeNamespace)
                     .withName(secretName)
@@ -200,7 +200,7 @@ public class KubernetesManager {
     }
 
     public boolean doesSecretExist(final String name) {
-        try (final KubernetesClient client = getKubernetesClient()) {
+        try (KubernetesClient client = getKubernetesClient()) {
             return Objects.nonNull(client.secrets().inNamespace(kubeNamespace).withName(name).get());
         } catch (KubernetesClientException e) {
             LOGGER.error(e.getMessage(), e);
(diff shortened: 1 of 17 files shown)


api/src/main/java/com/epam/pipeline/manager/region/CloudRegionManager.java
```diff
@@ -223,6 +238,33 @@ public AbstractCloudRegion loadOrDefault(final Long id) {
...
+            log.warn("Secret: " + CP_REGION_CREDS_SECRET + " doesn't exist!");
+            return;
+        }
+        kubernetesManager.refreshSecret(CP_REGION_CREDS_SECRET,
+                loadAll()
+                        .stream()
+                        .filter(region -> region.getProvider().equals(CloudProvider.AZURE))
+                        .map(region -> (AzureRegion) region)
+                        .collect(
+                                Collectors.toMap(azureRegion -> azureRegion.getId().toString(),
```
mzueva: This code duplicates logic in aspect, I'd suggest to move it to a separate service and use it in aspect and in `onApplicationEvent ` method

api/src/main/java/com/epam/pipeline/manager/region/CloudRegionManager.java
```diff
@@ -223,6 +238,33 @@ public AbstractCloudRegion loadOrDefault(final Long id) {
...
+                                        azureRegion ->  {
+                                            final HashMap<String, String> creds = new HashMap<>();
+                                            creds.put(STORAGE_ACCOUNT, azureRegion.getStorageAccount());
+                                            creds.put(STORAGE_KEY, loadCredentials(azureRegion).getStorageAccountKey());
+                                            try {
+                                                return Base64.encodeBase64String(mapper.writeValueAsString(creds).getBytes());
```
mzueva: Maybe extract this code to some util method or class, since such conversion is used several times

mzueva (changes requested):
Please review usage of `final` keyword in all new methods/classes

api/src/main/java/com/epam/pipeline/manager/cluster/KubernetesManager.java
```diff
@@ -167,6 +199,15 @@ public String createDockerRegistrySecret(DockerRegistrySecret secret) {
         }
     }
 
+    public boolean isSecretExist(final String name) {
```
mzueva: I think it is better to call methods `doesSecretExist`

SilinPavel:
@mzueva done

mzueva (approved):

Issue #911, opened 2020-01-22
`pipe` CLI shall allow to list the users/groups/objects permissions globally


sidoruka:
**Background**
The main idea is to automate the retrieval of:
* Which objects are accessible by a user?
* Which objects are accessible by a group?
* Which user/group have access to the object?

**Approach**

**Other options**

mzueva:
@SilinPavel Could you please assist with implementation of new API methods:
- load all permissions for user/group (including inherited permissions)

ekazachkova:
**Approach**

1. Which objects are accessible by a user?
To solve this problem we could use API method `POST /entities` introduced in this PR #949 and introduce a new CLI command `pipe view-user-objects`.

Usage example:
```
$ pipe view-user-objects <USER_NAME>
+--------------+------------+
| Type         | Name       | 
+--------------+------------+
| DATA_STORAGE | storage_1  | 
+--------------+------------+
| PIPELINE     | pipeline_1 | 
+--------------+------------+
```
2. Which objects are accessible by a group?
To solve this problem we could use API method `POST /entities` introduced in this PR #949 and introduce a new CLI command `pipe view-group-objects`.

Usage example:
```
$ pipe view-group-objects <GROUP_NAME>
+--------------+------------+
| Type         | Name       | 
+--------------+------------+
| DATA_STORAGE | storage_1  | 
+--------------+------------+
| PIPELINE     | pipeline_1 | 
+--------------+------------+
```
3. Which user/group have access to the object?
In this case we could use existing CLI command `pipe view-acl`.

Usage example:
```
$ pipe view-acl pipeline_1 -t pipeline
+-----------+-----------+----------------------+------+
|       SID | Principal |                Allow | Deny |
+-----------+-----------+----------------------+------+
| ROLE_USER |     False | Read, Write, Execute |      |
+-----------+-----------+----------------------+------+
```
But the implementation is too old and does not respect permissions for object parents. To fix it we could use `GET /permissions` instead of `GET /grant`.

NShaforostov:
@ekazachkova I've tried to verify this feature. In general, it works well.
Some notes about it:
1. No tools are printed in the full object accessible list for any user (including admins). And `pipe view-user-objects -t tool <user_name>` prints empty table.
2. No Docker registries are printed in the full object accessible list for any user (including admins). And `pipe view-user-objects -t docker_registry <user_name>` prints empty table.
[...]

PR #1082, opened 2020-04-30
Issue #911: `pipe` CLI shall allow to list the users/groups/objects permissions globally

ekazachkova:
This PR contains fixes according to this https://github.com/epam/cloud-pipeline/issues/911#issuecomment-620874885
- if specified group/role does not exists method `POST entities` will return error 
- fix for listing `view-user-objects`/`view-group-objects` commands for `tool`/`docker_registry` types

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
@@ -65,7 +65,6 @@ import com.epam.pipeline.manager.pipeline.ToolGroupManager;
 import com.epam.pipeline.manager.pipeline.ToolManager;
 import com.epam.pipeline.manager.pipeline.runner.ConfigurationProviderManager;
 import com.epam.pipeline.manager.security.run.RunPermissionManager;
-import com.epam.pipeline.manager.user.RoleManager;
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
```diff
@@ -382,6 +385,7 @@ public void filterTree(AclSid aclSid, AbstractHierarchicalEntity entity, Permiss
         if (aclSid.isPrincipal()) {
             filterTree(aclSid.getName(), entity, permission);
         } else {
+            roleManager.loadRole(aclSid.getName());
```
mzueva: Result of method is ignored, is it correct? Also I assume non-principal sid may be a group not only role

ekazachkova: @mzueva  This method is used to determine exists group or not. We do not need result of this method itself.

mzueva: Result of the method call is ignored, is it ok? I assume non-principal sid may be a `group` not only `role`.

mzueva (changes requested):

Issue #1104, opened 2020-05-13
NoMachine images begin to be unusable after commit


SilinPavel:
**Describe the bug**
Any NoMachine-based images that user creates with CloudPipeline commit operation will be unusable by other user.

**To Reproduce**
1. Start any NoMachine-based image
2. commit it to some other image
3. Start commited image with other user
4. try to login to virtual session with NoMachine client
5. See a massage like "wait for confirmation from logged user"

**Expected behavior**
User shall be able to login in to commited images

**Screenshots**

PR #1106, opened 2020-05-13
Issue 1104 nomachine image is unusable after commit

SilinPavel:
This PR is related to #1104 
It's fix NoMachine-based image buildings to be able to use this images after commit operation.

This situation was caused by bug in `nomachine_launcher.sh` script.
This script performs some pre-start actions (that actually change some config parameters by replacing strings) to run NX processes as user that start this `pipeline run`. But when we commit image NoMachine server config file is already changed and next time we will try to start this image pre-start action will not be applied.

This PR adds `server config template` file  that will be copied to `server config` file each time we run the image and all pre-start actions will be applied to the new and clear config file.


SilinPavel:
@sidoruka setup tools was updated just because of suddenly newly built image fail to start on `install pipe-common` step. Updating this package solves the problem. I assume that this action will be required for some newly built images.

sidoruka (approved):

Issue #1096, opened 2020-05-08
Billing shall include the disks costs, produced by the FSAutoscaler 


sidoruka:
**Background**
A clear and concise description of what the problem is. Ex. At the moment, Cloud Pipeline does not allow [...] and it would be nice to [...] because [...]

**Approach**

**Other options**

sidoruka:
GUI "estimated price" shall behave in the same manner as now.
@rodichenko just add the tooltip with the following info when hovering the price (wherever it is shown):
```
The estimated price is calculated using a machine type only.
During the job lifetime, it's configuration can be changed (e.g. disks added).
This additional information will be reflected in the "Billing Reports".
```

mzueva:
@tcibinan, @sidoruka  As we discussed:
- All API methods returning estimated price for a run shall be left as is (this prices includes only compute price and requested by user disk size)
- `PipelineRun` field `pricePerHour`, which is used for GUI, shall include: compute price, requested disk (adjusted to docker size), boot disk and swap disk
- Billing service shall calculate full price for instance including: compute price excluding paused periods, all disks (requested disk (adjusted to docker size), boot disk, swap disk, autoscaled disks) for full run period, including paused periods

rodichenko:
@sidoruka tooltip added (to Log form, Launch form, Configuration form, Runs table, Home Dashboard, Global Search)

mzueva:
@tcibinan We need to change the way of compute (run) billing is calculated in `billing-report-service`. Currently `RunToBillingRequestConverter` class uses run's `pricePerHour` field to calculate price for instance, change it to include compute (instance) price for VM running time and price for all disks for all instance existence time.

PR #1114, opened 2020-05-15
Provide node disks API

tcibinan:
Relates to #1096.

The pull request brings support for node disks api. From now node disks history data can be accessed via a new API method `GET /cluster/node/{name}/disks`. All attached disks of all nodes are automatically registered including os, data, swap and autoscaled disks.

Additionally the pull request expands run price calculation capabilities. Two additional fields are added to pipeline run entity that can be used in billing reports:
- `computePricePerHour` which specifies only virtual machine price per hour,
- `diskPricePerHour` which specifies 1GB disk price per hour.

Moreover overall run `pricePerHour` gets updated once compute node is assigned to the run to more precisely reflect node disks cost.

**Notice** that node disks API only covers runs that were created after the patch is applied.


api/src/main/java/com/epam/pipeline/dao/cluster/NodeDiskDao.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.stream.Collectors;
+
+@RequiredArgsConstructor
+public class NodeDiskDao extends NamedParameterJdbcDaoSupport {
+    
+    private final String insertNodeDiskQuery;
+    private final String loadNodeDisksByNodeIdQuery;
+    private final String deleteNodeDisksByNodeIdQuery;
+
+    public List<NodeDisk> insert(final String nodeId, final List<DiskRegistrationRequest> requests) {
```
mzueva, line 24: Please add `Transactional` annotation

api/src/test/java/com/epam/pipeline/dao/cluster/NodeDiskDaoTest.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final String NULL_NODE_ID = null;
+    private static final Long SIZE = 1L;
+    private static final Long NULL_SIZE = null;
+
+    @Autowired
+    private NodeDiskDao dao;
+
+    @After
+    public void deleteDisksByNodeId() {
+        dao.deleteByNodeId(NODE_ID);
```
mzueva: You can just add `@Transactional` annotation to the class and all DB changes will be rolled-back.

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/dao/cluster/NodeDiskDao.java
tcibinan: Added.

api/src/test/java/com/epam/pipeline/dao/cluster/NodeDiskDaoTest.java
tcibinan: That's good. Thanks. Added.

mzueva (approved):

PR #1132, opened 2020-05-22
Consider run node disks in billing

tcibinan:
Relates to #1096.

The pull request introduces support for enhanced node disks billing. From now on run billing reports include costs for all disks of run's node including os, data, swap and attached disks.

### DIFF
--- a/billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/model/PipelineRunWithType.java
+++ b/billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/model/PipelineRunWithType.java
@@ -16,12 +16,16 @@
 
 package com.epam.pipeline.billingreportagent.model;
 
+import com.epam.pipeline.entity.cluster.NodeDisk;
 import com.epam.pipeline.entity.pipeline.PipelineRun;
 import lombok.Value;
+
+import java.util.List;
 
 @Value
 public class PipelineRunWithType {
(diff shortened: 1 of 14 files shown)


billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/model/PipelineRunWithType.java
```diff
@@ -16,12 +16,16 @@
...
 import lombok.Value;
 
+import java.util.List;
+
 @Value
 public class PipelineRunWithType {
 
-    private PipelineRun pipelineRun;
-    private ComputeType runType;
+    PipelineRun pipelineRun;
```
mzueva, line 28: Can we make these fields `private`?

cloud-pipeline-common/model/src/main/java/com/epam/pipeline/entity/cluster/NodeDisk.java
```diff
@@ -0,0 +1,12 @@
+package com.epam.pipeline.entity.cluster;
+
+import lombok.Value;
+
+import java.time.LocalDateTime;
+
+@Value
+public class NodeDisk {
+    Long size;
```
mzueva, line 9: Can we make all fields `private`?

mzueva (changes requested):

Issue #1126, opened 2020-05-21
`pipe run` custom parameters parsing issue


sidoruka:
**Describe the bug**
For some reason, `pipe` can't execute a `run` command if it uses a custom parameter with a `-q main.q` value. If I change this to `-q my.q` (or whatever name) - it works.

**To Reproduce**
1. Execute 
```
pipe run -n 11111 \
    CUSTOM_PARAM '-q main.q' \
    -y -id 100 \
    -it m5.xlarge \
    -di library/centos \
    -cmd 'python $SCRIPTS_DIR/src/[main_file]'
```
2. Observed
```
Failed to fetch data from server. Server responded with message: Pipeline with requested id: '.q' was not found.
```

**Expected behavior**
Command shall fail with:
```
Failed to fetch data from server. Server responded with message: Pipeline with requested id: '11111' was not found.
```

mzueva:
Seems to be related to https://github.com/pallets/click/issues/249

mzueva:
@ekazachkova We need to change the generation of `pipe run` command on the server-side: pipeline run parameters shall be attached at the end of the command and shall be preceded by **"--"** chars.

It also would be great add a new line sign (**"\\"** for Linux and **"^"** for Windows) to each chunk of generated command.

mzueva:
cherry-pick to release/0.15 beb66d138c02a30284aaa93781629eaa275448c6

PR #1133, opened 2020-05-22
Issue #1126: `pipe run` custom parameters parsing issue

ekazachkova:
This PR provides fix for issue #1126 
The issue was that some run parameters values were parsed incorrectly via pipe CLI. To fix it we should put all run parameters at the end of the command and specify `--` before them.
Also, this PR provide style fixes for launch cmd: each options group (option name and value) should be placed at the new line.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/runner/PipeRunCmdBuilder.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/runner/PipeRunCmdBuilder.java
@@ -169,7 +169,11 @@ public class PipeRunCmdBuilder {
             cmd.add(parametersCommand);
         }
         if (Objects.nonNull(runVO.getParentRunId())) {
-            addCmd(String.format("parent-id %d", runVO.getParentRunId()));
+            cmd.add(getNewLineIndicator());
+            if (MapUtils.isEmpty(runVO.getParams())) {
+                cmd.add(CMD_OPTIONS_DELIMITER);
+            }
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/pipeline/runner/PipeRunCmdBuilder.java
```diff
@@ -130,49 +133,50 @@ public PipeRunCmdBuilder config() {
...
+            addCmd(CMD_OPTIONS_DELIMITER);
             final String parametersCommand = runVO.getParams().entrySet()
                     .stream()
                     .map(this::prepareParams)
-                    .collect(Collectors.joining(WHITESPACE));
+                    .collect(Collectors.joining(WHITESPACE + getNewLineIndicator() + WHITESPACE));
             cmd.add(parametersCommand);
         }
         if (Objects.nonNull(runVO.getParentRunId())) {
-            cmd.add(String.format("parent-id %d", runVO.getParentRunId()));
```
mzueva, line 168: If parameters are empty, we need to add '--' here as well

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/pipeline/runner/PipeRunCmdBuilder.java
ekazachkova: added

mzueva (approved):

Issue #1122, opened 2020-05-21
Allow to "symlink" the tools between the tools groups


sidoruka:
**Background**
At the moment, the majority of the tools are managed by the administrators and are available via the `library` tool group.
But some of the users would like to have separate tool groups, which are going to contain a mix of the custom tools (managed by the users) and the `library` tools (managed by the admins).
The latter ones shall be "symlinked" to the users' tool groups. The following behavior is expected:
* "Symlink" creation operation shall be available to the users which:
  * Have read access to the source tool
  * Have write access to the destination tool group
* It shall not be possible to create two levels of "symlinks"
* Users shall be able to run "symlinks" as usual, the following permissions shall be checked to run or view the "symlink":
  * "Symlink" permissions
  * Source image permissions
* It shall not be possible to "push" into the "symlinked" tool. Two options can be considered:
  * Just deny any push operation
  * Behave as real symlinks and push into the source tool
* "Symlinks" shall display all the same description, icon, settings as the source image
* It shall not be possible to make any changes to the "symlink" data (description, icon, settings), even for the admins
* OWNERs shall be still able to manage the permissions for the symlinks

**Approach**
A new field `linkTo` shall be added to `Tool` object to indicate that such a tool is only a link to a real `Tool`.
* Add a new API methods to create a new `symlink` tool from `original` tool
* When a `Tool` is loaded from DB some fields `[description, labels, endpoints, default_command, instance_type, short_description, icon_id]` shall be loaded from `original` tool
* All API methods  to update `symlink` tool shall be forbidden
* All API methods to read versions/attributes/settings/scan/vulnerabilities for `symlink` tool shall load data for `original` tool
* All API methods to update/delete versions/attributes/settings/scan/vulnerabilities for `symlink` tool shall be forbidden
* All push and pull events shall be rejected for `symlink` tools in `/dockerRegistry/oauth` endpoint
* `symlink` tools shall be skipped during scheduled security scanning
* When a run is started with `symlink` tool as docker image it shall be replaced with `original` image for k8s pod spec, but stored as is for DB
* For now permissions shall be handled separately for `symlink` and `original` tool

NShaforostov:
I've tried to verify the feature.
In general, it works nicely.

The following issues were encountered during the verifying:

1. Imagine the case:
- admin create tool group (no matters - personal or not)
- admin give all permissions (R/W/X) to that group for the user (not admin, but with the `ROLE_TOOL_GROUP_MANAGER` role)
[...]

PR #1150, opened 2020-06-04
Support tool symlinks

tcibinan:
Resolves issue #1122.

The pull request brings support for so-called tool symlinks. Tool symlink is a representation of some tool in a different tool group. Symlink inherits all the source tool parameters such as name, description, versions, issues and etc. If some of the source tool parameters are changed then its symlink parameters are changed as well. At the same time none of the symlink parameters can be modified. 

Permissions and owner of tool symlink can be modified separately from the source tool.

An additional API method `POST /tool/symlink` is added in order to create new symlink. It accepts body with just two fields:
- `toolId` of the source tool,
- `groupId` of the tool group to create symlink in.


api/src/main/java/com/epam/pipeline/manager/docker/DockerRegistryManager.java
```diff
@@ -591,6 +591,12 @@ private Tool buildTool(DockerRegistry dockerRegistry, ToolGroup toolGroup, Strin
                 }
             }
 
+            if (isSymlinkWriting(entity, permissions)) {
```
mzueva: I think all operations `read` & `write` shall be forbidden (since user won't be able to pull a symlink anyway)

api/src/main/java/com/epam/pipeline/manager/issue/IssueManager.java
```diff
@@ -97,6 +103,7 @@ public Issue createIssue(IssueVO issueVO) {
         EntityVO entityVO = issueVO.getEntity();
         validateEntityParameters(entityVO);
         AbstractSecuredEntity entity = ensureEntityExists(entityVO);
+        validateEntityCanBeModified(entityVO);
```
mzueva, line 106: Issues can be attached to symlinks

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/docker/DockerRegistryManager.java
tcibinan: Seems reasonable, done.

api/src/main/java/com/epam/pipeline/manager/issue/IssueManager.java
tcibinan: As discussed with @sidoruka issues creation shall be forbidden for symlinks.

PR #1164, opened 2020-06-10
Execute data transfer under specific user

Wedds:
Previously, synchronous CmdExecutors were able to launch command only under the active user.
But sometimes, such as during data transfer, it is useful to execute commands under a specific user, to match access control rules. 

- `CmdExecutor` and some implementations were updated to proceed `username` argument to execute a command with a specific user
- `data-transfer-service` is upgraded to handle `user` argument correctly as a part of the transfer task to execute underlying commands with CmdExecutor under the privileges of the selected user


mzueva (changes requested):
Two points to consider:
1. Let's do not require to pass used in the DTO object and take username from JWT token which was used for authentication. In order to convert authentication username to linux username add a new application property: `dts.username.transformation` with supported values `LOWERCASE`, `UPPERCASE`, `NONE` (default), if `LOWERCASE` or `UPPERCASE` value is provided, corresponding function shall be applied to username from JWT token.
2. I assume that file listing shall be executed under specific user as well (@sidoruka correct me if I am wrong). This may require to execute file system user using shell instead of java code.

tcibinan:
Certain changes were performed to the original pull request.

From now data transfer service support two impersonation modes:
- *impersonation mode*. All the local file system operations are performed by different users depending on the request. The user that will be impersonated during file system calls is retrieved from either request jwt token or from the corresponding user metadata in Cloud Pipeline. *The impersonating mode is used by default.*
- *non impersonation mode*. All the local file system operations are performed by the same user which launched the data transfer application itself.

Impersonation mode works in the following manner for listing and transfer requests. If request contains complete credentials object in it then data transfer service tries to request the corresponding user metadata from Cloud Pipeline. If the corresponding user metadata contains a value by some predefined key (see `dts.impersonation.name.metadata.key`) then it will be used as an impersonating user for file system calls. If such metadata value is not found or the credentials are not provided then the original request jwt token user will be used as an impersonating user after its name is transformed with the configured transformation (see `dts.impersonation.name.transformation`).

The following required application properties have been added:
- `dts.listing.listScript` specifies location of the list script `data-transfer-service/src/main/resources/list.py`. Has no default value.

The following optional application properties have been added:
- `dts.impersonation.enabled` enables user impersonation mode. Enabled by default.
- `dts.impersonation.name.transformation` specfies request jwt token user name transformation function which can be set to `LOWERCASE`, `UPPERCASE` or `NONE` . Defaults to `LOWERCASE`.
- `dts.impersonation.name.metadata.key` specifies user metadata key to retrieve impersonating user name. Defaults to `dts_name`.
- `dts.listing.listCommand` specifies list command template. The following arguments will be substituted in the template: list script location (see `dts.listing.listScript`), listing absolute path, listing page offset, listing page size. Defaults to `%s --path '%s' --offset '%d' --size '%d'`.

mzueva (approved):

Issue #1173, opened 2020-06-17
Pipe mount shall support Google Cloud Storage buckets


mzueva:
**Background**
`pipe mount` command shall support mounting Google Cloud Storage buckets with `gs://` scheme.

PR #661, opened 2019-09-13
Introduce AWS storages support for pipe fuse

tcibinan:
Implerments #679: Along with #629 brings support for another mount type to built-in pipe fuse.

Brings support for AWS S3 storage mounts to pipe cli. In-memory file system is inspired by both [goofys](https://github.com/kahing/goofys) and [s3fs](https://github.com/s3fs-fuse/s3fs-fuse) fuses.

# Details

Enables listing caching, in-memory read and write buffering for both WebDAV and AWS S3 mounts. 
Read buffering uses so-called read ahead approach which retrieves data before it is requested.

## Limitations

AWS fuse uses so-called multipart upload to upload most of the files. It allows random writes and even non-sequential random writes. Nevertheless AWS S3 multipart upload introduces several limitations on the uploading file:
 - it should be smaller than 5TB,
 - it can be split up to 10000 parts.

In order to support non-sequential random writes pipe fuse uses chunked multipart upload strategy. It means that each object being uploaded is split into the series of chunks with a predefined size. It allows pipe fuse to upload only the updated chunks keeping the remaining ones untouched. In the picture below only the yellow sections are the actual updates. It can be seen that partially updated chunks should be uploaded entirely including the regions that were untouched.

pipe-aws-fuse-chunks

*Nevertheless the predefined chunk size ends up in the uploading file size limitation.* The allowed uploading file size is basically a chunk size multiplied by 10000. F.e. for chunk size of *10MB* the allowed uploading file size would be *100GB*, *100MB* -> *1TB*, *500MB* -> *5TB*.

It's hard to say what chunk size should be used by default. On the one hand the bigger the chunk size is the faster simple writing to the end of file will be. On the other hand the smaller the chunk size is the less data will be transferred from cloud and to the cloud.

## CentOS

Currently *centos:7* runs requires additional `fuse-libs` package to be installed to use pipe AWS fuse.

# Usage

## pipe cli

Two new arguments were added to `pipe storage mount` command. These arguments are mutually exclusive.

- `-f, --file` specifies that all available file systems should be mounted into mountpoint.
- `-b, --bucket BUCKET_NAME` specifies bucket name to mount into the mountpoint.

## pipe fuse

Several new arguments were added directly to pipe fuse cli.

- `-b, --bucket BUCKET_NAME` specifies bucket name to mount into the mountpoint.
[...]

PR #1187, opened 2020-06-26
Support google cloud storages in pipe mount

tcibinan:
Resolves issue #1173.

The pull request brings support for Google Cloud to pipe mount tool. The implementation mostly reuses functionality from AWS  support resolved in #661. All of the provided independent features such as caching and buffering are supported for google provider as well. The single notable difference between the AWS and Google Cloud implementations is the usage of so-called [composite objects](https://cloud.google.com/storage/docs/composite-objects) in the second case.

# Details

Google Cloud has similar yet different approach for multipart uploading in compare with AWS. 

In AWS an uploading object can be constructed from parts of already uploaded object or from additionally uploaded parts. All the temporary parts are not visible to the user.

Google Cloud allows to merge several already uploaded objects to a single composite one. Using such mechanism pipe mount uploads all the parts as temporary files and then composes them. All the temporary files are visible to the user from moment they are uploaded to the storage till the moment the upload is completed.

_As a result Google Cloud allows to support sequential and non-sequential random writes without any uploading file size limitations._ Nevertheless default chunk size considerations is the same as for AWS.

## Benchmarks

| operation | gcsfuse | pipe |
| --- | --- | --- |
| list 1000 | 0m31.384s | 0m0.365s |
|           | 0m30.648s* | 0m0.012s |
| read 1G   | 0m7.136s | 0m18.125s |
|           | 0m7.377s | 0m18.663s |
| read 10G  | 1m56.705s | 2m52.147s |
|           | 1m55.685s | 2m55.278s |
| write 1G  | 0m10.926s | 0m37.361s |
|           | 0m15.681s | 0m35.497s |
| write 10G | 4m32.414s | 5m53.244s |
|           | 6m46.451s | 5m51.303s |

> \* Most subsequent attempts to list 1000 files using gcsfuse stuck for more than 4 minutes and was interrupted after.

### DIFF
--- a/pipe-cli/build_linux.sh
+++ b/pipe-cli/build_linux.sh
@@ -65,7 +65,7 @@ cd $PIPE_MOUNT_SOURCES_DIR && \
 python2 $PYINSTALLER_PATH/pyinstaller/pyinstaller.py \
                                 --hidden-import=UserList \
                                 --hidden-import=UserString \
-                                --additional-hooks-dir="{$PIPE_MOUNT_SOURCES_DIR}/hooks" \
+                                --additional-hooks-dir="${PIPE_MOUNT_SOURCES_DIR}/hooks" \
                                 -y \
                                 --clean \
                                 --runtime-tmpdir $PIPE_CLI_RUNTIME_TMP_DIR \
--- a/pipe-cli/build_linux.sh
+++ b/pipe-cli/build_linux.sh
@@ -65,6 +65,7 @@ cd $PIPE_MOUNT_SOURCES_DIR && \
 python2 $PYINSTALLER_PATH/pyinstaller/pyinstaller.py \
                                 --hidden-import=UserList \
                                 --hidden-import=UserString \
+                                --additional-hooks-dir="{$PIPE_MOUNT_SOURCES_DIR}/hooks" \
                                 -y \
                                 --clean \
                                 --runtime-tmpdir $PIPE_CLI_RUNTIME_TMP_DIR \
--- a/pipe-cli/mount/build_mount.sh
+++ b/pipe-cli/mount/build_mount.sh
@@ -43,6 +43,7 @@ cd $PIPE_MOUNT_SOURCES_DIR && \
 python2 $PYINSTALLER_PATH/pyinstaller/pyinstaller.py \
                                 --hidden-import=UserList \
                                 --hidden-import=UserString \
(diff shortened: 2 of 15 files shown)


NShaforostov:
@tcibinan 
I've tried to verify the version 5808. In general, it works well. Some minor issues that were encountered during the testing:
1. Trying to perform `cp -r` or `mv` command, when the _source_ is the **folder with sub-folders** on the local workstation and the _destination_ is the mountpoint or its sub-folders, throws the "input/output" error:
image
2. Trying to perform `cp -r` or `mv` command, when the _source_ and the _destination_ are mountpoint's subfolders, throws the "input/output" error:
image
3. `mkdir` command doesn't work in the mountpoint's sub-folders:
image
4. Imagine the case:
- try to execute `truncate`/`fallocate` command with the big-size file in the mount directory, e.g. `truncate -s +100G file100`
- pause the run until the end of the previous command execution
- view the storage that was mounted - see lots of small temp files (each 10Mb):
image
Could we do something with these temp files in such situations - e.g., hide them all in separate temp folder - to not clog the working folder?

tcibinan:
Points 1,2,3 should be fixed by e5c3ebc1fcf32cf019e7aff1c01bf21bc286d768. Point 4 cannot be resolved due to the way GCP handles composite uploads. Nevertheless all the temporary files should be hidden in temporary folders by 1dee286e3ccfef01aec80f59498fd7ace2910228.

NShaforostov:
@tcibinan 
Verified, points 1-3 are fixed. Approach with the point 4 also works well.

NShaforostov:
Updates were verified on `GCP` and `AWS` stands. Works well.
Were checked:
- multi threads

tcibinan:
According to [documentation](https://cloud.google.com/storage/docs/composite-objects#integrity-checking) composite objects hashsum calculation requires additional package CRC32C to be installed. If it is not installed then such objects downloading using gsutil will require additional option to be specified which disables hashes checking.

```
gsutil -o GSUtil:check_hashes=if_fast_else_skip cp gs://fuse-dev-storage-1/file_100M_C2 ./
```

Details on how to install the required package can be found using the following command.

```
gsutil help crcmod
```

Both pipe storage mount and Cloud Pipeline works fine with such objects because hashes checksum is not performed.

mzueva (approved):

Issue #1192, opened 2020-06-30
API HA configuration improvements


sidoruka:
**Background**
Since the #292 implementation - we have a couple of issues, introduced by the HA configuration:
* User cannot log in via SAML, as the session is stored "in-memory" and different API replicas are not aware of the `JSESSIONID`
* It's hard to distinguish logs from different replicas (all written into the same file)
* The `MY_POD_NAME` environment variable, used to specify the replica hostname, is not configurable and hard to figure out

**Approach**
* Store user's session in the database
* Add `MY_POD_NAME` value into the API log entries (make sure that the #1010 is not broken - Filebeat config)
* `MY_POD_NAME` environment variable name shall be configurable via `application.properties`

PR #1194, opened 2020-06-30
Issue #1192 API HA config improvements

Wedds:
This PR is related to issue #1192 

- User's session is stored in the database, so the authentication will remain if leading pod is changed
- Active API pod name is added to the beginning of the API log entries 
- Environment variable to look for pod name is configurable via `application.properties`

### DIFF
--- a/api/src/main/resources/log4j2-spring.xml
+++ b/api/src/main/resources/log4j2-spring.xml
@@ -18,7 +18,7 @@
 <Configuration monitorInterval="60">
     <Properties>
         <Property name="log-path">logs</Property>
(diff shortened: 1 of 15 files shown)


api/src/main/resources/log4j2-spring.xml
```diff
@@ -18,12 +18,13 @@
 <Configuration monitorInterval="60">
     <Properties>
         <Property name="log-path">logs</Property>
+        <Property name="cp-api-logger-hostname">$${env:MY_POD_NAME:-cp-api-srv}</Property>
```
mzueva: Shouldn't we change env var name here as well?

Wedds: Done

Issue #1038, opened 2020-04-06
"Serverless" Cloud Pipeline API


sidoruka:
**Background**
At the moment, Cloud Pipeline is used to run on-demand jobs, which then can be consumed in a different manner. E.g. some data processing service, exposing own API can be launched.
The current approach for using such a service is:
* Launch the `tool` via GUI/CLI/API
* Wait for the initialization
* Grab the interactive endpoint URL
* Submit a request to the URL

While this works fine (and can be automated easily) - it is quite useful to reduce the number of the required operations.
To address this - a sort of "serverless" approach shall be implemented. I.e. when the "target" API can be requested as a "step 1" so that the workflow will look like:
* Submit a request to the URL

_Under the hood the following will happen automatically:_

* `tool` is launched and initialized
* Interactive endpoint is exposed
* Request is submitted to the URL

**Approach**
* Implement a service, that is capable of:
  * Accepting HTTP requests
  * Parsing the request according to the template
  * Based on the parsing results - a `tool` and an `endpoint` shall be retrieved
  * Tool shall be launched
  * Original request shall be submitted to the `endpoint`
  * Response to the client shall be returned in a sync manner (to the tool's response)
* Tools, that are launched in this manner - shall keep running for a configurable amount of time, since the last request

mzueva:
### Approach

#### API
Main Cloud Pipeline API server shall provide a new endpoint for `serverless` tasks management with the following methods:
- `POST` submits new task via the following request:
```
{
   "nodeConfiguration": {
        "dockerImage": "library/centos:latest",
        // all other fields supported by POST run endpoint
   },
   "endpoint": "Application",
   "path": "api/tasks",
   "method": "POST",
   "body": "{ \"name\": \"test task\" }",
   "retryCount": 3,
   "stopAfter": false,
   "notify": true
}
```
- `GET` returns task status by id and results for completed tasks
- `DELETE` cancels task execution

#### Implementation
API server shall process submitted tasks asynchronously. Task processing consists of three main steps and one optional: 
- **task submission**: task is created and saved to DB
- **run initialisation**: at this step run is created with the requested parameters, step ends when runs is initialised and requested endpoint is available
- **request submission**: request is submitted via endpoint with specified `path`, `method` and `body`. Http result of the request is saved as task result. If request fails due to some reason it is retried, if `retryCount` parameter is specified
- **run stop**: if `stopAfter` parameter is `true`, run is stopped after request results are received

[...]

PR #1222, opened 2020-07-16
Issue #1038: "Serverless" Cloud Pipeline API - fixes

ekazachkova:
This PR provides updates for current implementation for issue #1038 

- the field `RunConfiguration#stopAfter` is numeric value now and indicates how long corresponding run will be alive after the last request (in minutes). If this value is not specified the value `launch.serverless.stop.timeout` from system preferences will be used
- bearer token will be added to request cookies if not specified (required for `edge` auth)

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
@@ -25,6 +25,7 @@ import com.epam.pipeline.entity.pipeline.CommitStatus;
 import com.epam.pipeline.entity.pipeline.Pipeline;
 import com.epam.pipeline.entity.pipeline.PipelineRun;
 import com.epam.pipeline.entity.pipeline.RunInstance;
+import com.epam.pipeline.entity.pipeline.StopServerlessRun;
 import com.epam.pipeline.entity.pipeline.TaskStatus;
 import com.epam.pipeline.entity.pipeline.run.ExecutionPreferences;
 import com.epam.pipeline.entity.pipeline.run.parameter.RunAccessType;
@@ -449,11 +450,9 @@ public class PipelineRunDao extends NamedParameterJdbcDaoSupport {
                 .findFirst();
     }
 
-    public List<PipelineRun> loadServerlessRunsToStop(final LocalDateTime maxLastUpdate) {
-        final MapSqlParameterSource params = new MapSqlParameterSource();
-        params.addValue("MAX_LAST_UPDATE", maxLastUpdate);
+    public List<StopServerlessRun> loadServerlessRunsToStop() {
         return ListUtils.emptyIfNull(getNamedParameterJdbcTemplate()
-                .query(loadServerlessRunsToStopQuery, params, PipelineRunParameters.getRowMapper()));
+                .query(loadServerlessRunsToStopQuery, StopServerlessRunDao.StopServerlessRunParameters.getRowMapper()));
     }
 
     private MapSqlParameterSource getPagingParameters(PagingRunFilterVO filter) {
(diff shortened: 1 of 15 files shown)


api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
```diff
@@ -449,11 +450,9 @@ public void deleteRunSidsByPipelineId(final Long pipelineId) {
                 .findFirst();
     }
 
-    public List<PipelineRun> loadServerlessRunsToStop(final LocalDateTime maxLastUpdate) {
-        final MapSqlParameterSource params = new MapSqlParameterSource();
-        params.addValue("MAX_LAST_UPDATE", maxLastUpdate);
+    public List<StopServerlessRun> loadServerlessRunsToStop() {
         return ListUtils.emptyIfNull(getNamedParameterJdbcTemplate()
```
mzueva: I'd suggest to move this method to `StopServerlessRunDao` class

api/src/main/java/com/epam/pipeline/dao/pipeline/StopServerlessRunDao.java
```diff
@@ -38,21 +39,22 @@
...
+    @Transactional(propagation = Propagation.REQUIRED)
     public void createServerlessRun(final StopServerlessRun run) {
         final Long id = createServerlessRunId();
         run.setId(id);
         getNamedParameterJdbcTemplate().update(saveServerlessRunQuery,
                 StopServerlessRunParameters.getParameters(run));
     }
 
-    @Transactional(propagation = Propagation.MANDATORY)
+    @Transactional(propagation = Propagation.REQUIRED)
```
mzueva: For most classes we keep `Mandatory` propagation at DAO level, why do we need `Required` in this case?

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -408,11 +409,21 @@ private void setInstanceTypeMap(final Map<String, InstanceType> instanceTypeMap)
...
+        }
+
+        private boolean serverlessRunIsExpired(final StopServerlessRun run) {
+            return run.getLastUpdate().isBefore(LocalDateTime.now().minusMinutes(getTimeoutMinutes(run)));
+        }
+
+        private Long getTimeoutMinutes(final StopServerlessRun run) {
+            return Objects.nonNull(run.getStopAfter())
+                    ? run.getStopAfter()
+                    : preferenceManager.getPreference(SystemPreferences.LAUNCH_SERVERLESS_STOP_TIMEOUT).longValue();
```
mzueva, line 431: We also need to handle situation when `LAUNCH_SERVERLESS_STOP_TIMEOUT` is null - consider run non-expired in this case

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
ekazachkova: Thanks, done

api/src/main/java/com/epam/pipeline/dao/pipeline/StopServerlessRunDao.java
ekazachkova: We need `Required` here due to this method called from non-transaction context.
I added a new class `StopServerlessRunManager` to manage transactions.

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
ekazachkova: Thanks, done

mzueva (approved):

Issue #1214, opened 2020-07-13
Pipe storage cp/mv/rm commands shall support multithreaded execution


mzueva:
**Background**
Currently `pipe storage cp/mv/rm` commands process each object (file) in the storage sequentially although file upload/download is implemented using multipart approach, which may use several threads. It would be great to implement a parallel by file upload/download/deletion approach to improve performance of operations with a large number of small files.

**Approach**
Add a new option '-t --threads N' to `pipe storage cp/mv/rm` commands. If this option is specified `pipe` shall use N threads to process files. 

In order to evenly distribute workload between threads I'd suggest the following algorithm:
- list all files in the destination and sort them by files size descending
- distribute files by threads in this order (round-robin)

tcibinan:
Recently I've found support for [batch requests in gcp](https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch). Probably it could be useful here.

tcibinan:
Cherry-picked to `release/0.16` via 485c9a45b26604f4179c0273ea7b8ec117ff5585.

PR #1241, opened 2020-07-27
Pipe storage cp/mv commands shall support multithreaded execution

ekazachkova:
This PR provides implementation for issue #1214 

A new CLI option `--threads (-n)` was added to `pipe storage cp/mv` operations. Available for folders only. This option indicates a multithreded files processing. The processing based on the following algorithm:
1. lists all files in folder recursively
2. sorts them by size descending
3. splits into N chunks where N is the `--threads (-n)` option value

This option is aimed at working with a huge number of small files.

The progress bar is disabled if  `--threads (-n)` is specified.

### Benchmarks

The benchmarks were taken for `AWS` provider only.

The following tests datasets were generated:
- 1000 files with 7 Mb size
- 100 files with 20 Mb size

Instance info: m5.large (2 CPU, 8 RAM)

1000 files with 7 Mb size:

|Operation	|n = 1		|n = 2		|n = 3		|n = 4		|n = 8|n = 16|
| --- | --- | --- | --- | --- | --- |--- |
|Upload		|4m42.889s	|2m8.634s	|1m37.565s	|1m13.003s	|1m1.270s|0m52.974s|
|Download	|4m47.033s	|2m25.133s	|1m38.428s	|1m11.698s	|0m52.849s|0m52.069s|

100 files with 20 Mb size:

|Operation|	n = 1|		n = 2|		n = 3|		n = 4|		n = 8|		n = 16|
| --- | --- | --- | --- | --- | --- | --- |		
|Upload|	0m48.492s|	0m22.963s|	0m18.422s|	0m15.642s|	0m13.278s|	0m13.423s|
|Download|	0m27.859s|	0m16.909s|	0m14.886s|	0m14.190s|	0m13.931s|	0m15.489s|

In case with 20 Mb size files a low performance gain is observed with n >= 3. This is due to the fact that `boto3` has its own parallel implementation by file. The parallel computation starts when the file size is greater than [multi-part threshold](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-threshold) (default : 8Mb). This way, threads start to competing for resource and affect the performance.

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -884,7 +884,7 @@ def storage_remove_item(path, yes, version, hard_delete, recursive, exclude, inc
               "filter - follow symlinks but check for cyclic links")
 @click.option('-n', '--threads', type=int, required=False,
               help='The number of threads that will work to perform operation. Allowed for folders only. '
-                   'Use to move a huge number of small files. Not supported for Windows OS')
+                   'Use to move a huge number of small files. Not supported for Windows OS. Progress bar is disabled')
 @click.option('-u', '--user', required=False, callback=set_user_token, expose_value=False, help=USER_OPTION_DESCRIPTION)
 @Config.validate_access_token(quiet_flag_property_name='quiet')
 def storage_move_item(source, destination, recursive, force, exclude, include, quiet, skip_existing, tags, file_list,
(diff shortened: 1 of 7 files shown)


tcibinan (approved):

maryvictol:
Verified for Linux OS. Works well

Issue #441, opened 2019-06-24
Support AWS FSx for Lustre


sidoruka:
1. Add Lustre package installation into the host AWS AMIs, according to https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html (`Amazon Linux 2`)
2. Introduce a new mount type: `Lustre` (in addition to existing NFS/SMB)
3. If `Lustre` is specified - install client packages into the run, being launched, according to https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
4. Mount `Lustre` FS according to https://docs.aws.amazon.com/fsx/latest/LustreGuide/mounting-ec2-instance.html

mzueva:
### Implementation details

I'd suggest to implement this feature in three steps:
1. Add support of `Lustre` storages on the API server
      - support new `MountType` `Lustre` for `FileShareMount` object and 
      - implement mounting of `Lustre` storages to the API server (`NFSStorageProvider.mount` method)
      - add installation of lustre client to the API server docker (`deploy/docker/cp-api-srv/Dockerfile`)
2. Add support of mounting `Lustre` storages to runs
      - handle `Lustre` storages in `workflows/pipe-common/scripts/mount_storage.py` script
3. Support `Lustre` storages in WedDAV synchronisation
     - handle `Lustre` storages in `scripts/nfs-roles-management`
     - add installation of lustre client to the WEBDAV docker `deploy/docker/cp-dav/Dockerfile`

PR #1252, opened 2020-08-03
Issue #441 Lustre path handling improvements

Wedds:
This PR is related to issue #441.
It updates the handling of Lustre paths. 
Lustre allows specifying complicated ("multi-host") mount roots [[1]](http://wiki.lustre.org/Lustre_Networking_(LNET)_Overview) [[2]](http://wiki.lustre.org/Mounting_a_Lustre_File_System_on_Client_Nodes). As the same char ':' is used for many purposes:
- split port from a server in host specification
- split LNet Network Identifiers (NID) in host specification
- part of sequence splitting host from the other part of mount path (':/')
user input is verified against the specific regex, describing valid Lustre path.
The format is checking during `NFSStorageProvider.mount` operation, as well, and operation is being aborted if the format is invalid.
Also, Lustre mount fails, if path ends with a separator (even if valid), so additional operation is applied if necessary.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSStorageProvider.java
+++ b/api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSStorageProvider.java
@@ -32,6 +32,7 @@ import com.epam.pipeline.entity.datastorage.DataStorageListing;
 import com.epam.pipeline.entity.datastorage.DataStorageStreamingContent;
 import com.epam.pipeline.entity.datastorage.DataStorageType;
 import com.epam.pipeline.entity.datastorage.FileShareMount;
+import com.epam.pipeline.entity.datastorage.MountType;
 import com.epam.pipeline.entity.datastorage.PathDescription;
 import com.epam.pipeline.entity.datastorage.nfs.NFSDataStorage;
 import com.epam.pipeline.entity.region.AbstractCloudRegion;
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSStorageProvider.java
```diff
@@ -206,8 +214,10 @@ private synchronized File mount(NFSDataStorage dataStorage) {
 
     private synchronized void unmountNFSIfEmpty(AbstractDataStorage storage) {
         FileShareMount fileShareMount = shareMountManager.load(storage.getFileShareMountId());
-        File rootMount = Paths.get(rootMountPoint, normalizePath(fileShareMount.getMountRoot())).toFile();
-
+        final String shareMountPath = MountType.LUSTRE == fileShareMount.getMountType()
+                                      ? normalizeLustrePath(fileShareMount.getMountRoot())
+                                      : normalizePath(fileShareMount.getMountRoot());
+        File rootMount = Paths.get(rootMountPoint, shareMountPath).toFile();
```
tcibinan: It would be nice to extract this to a separate method since this peace of code is duplicated several times in the file.

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSStorageProvider.java
```diff
@@ -166,7 +167,14 @@ private synchronized File mount(NFSDataStorage dataStorage) {
         File mntDir = Paths.get(rootMountPoint, getMountDirName(dataStorage.getPath())).toFile();
         try {
             FileShareMount fileShareMount = shareMountManager.load(dataStorage.getFileShareMountId());
-            File rootMount = Paths.get(rootMountPoint, normalizePath(fileShareMount.getMountRoot())).toFile();
+            final String mountPath;
+            if (MountType.LUSTRE == fileShareMount.getMountType()) {
+                mntDir = Paths.get(rootMountPoint, getLustreMountDirName(dataStorage.getPath())).toFile();
```
tcibinan: I suppose that If we can use final or at least effectively final variables than we should do that. Let's assign `mntDir` variable only once. Moreover we can extract this to a separate method.

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -522,6 +522,8 @@ def build_mount_command(self, params):
                 params['path'] = '//' + params['path']
         elif self.share_mount.mount_type == "LUSTRE":
             command = command.format(protocol="lustre")
+            if params['path'].count('/') > 1 and params['path'].endswith('/'):
+                params['path'] = params['path'][:-1]
```
tcibinan: I'm just wondering if it's ok that we have trailing slashes in path for only lustre mounts. If there are some formal reasons for this then ok. Otherwise it would be great to unify the behavior between different storage types.

Please share what you think before implementing it.

Wedds: Sorry, it was a part of a debugging routine, following commits brings more thorough Lustre support.

PR #1256, opened 2020-08-04
Handle listing caches in thread safe manner in pipe fuse

tcibinan:
The pull request introduces thread safe listing caches in pipe fuse. Previously pipe fuse plainly used [cache implementation](https://cachetools.readthedocs.io/en/latest/#cache-implementations) that is not thread safe which probably was the cause of some issues in the past. From now on all interactions with the cache implementation are synchronized.


NShaforostov:
Verified. Works well

tcibinan:
Cherry-picked to `release/0.16` via 5a17d10dfe80e0d1b2da3b786f5b53b954742457.

PR #1264, opened 2020-08-10
Log pipe fuse errors explicitly

tcibinan:
Relates to #1030.

The pull request brings explicit implementation errors logging to pipe fuse. Previously, some of the exceptions were visible only in DEBUG mode.

Also pull request introduces caching for existence calls and preliminary invalidates caches for missing requested files.


tcibinan:
Cherry-picked to #1272.

PR #1272, opened 2020-08-14
Recover from unmanageable writes in pipe fuse

### LINKED ISSUE #1271: Reinitialize upload in case of unmanageable writes in pipe fuse
**Background**

Currently some complicated random write scenarios is not supported in pipe fuse yet. For instance if writing to file is being performed to some regions that were already uploaded to cloud within the same multipart upload then it will fail.

**Approach**

Efficient managing of such cases is pretty complicated thing to implement and requires some changes in pipe fuse core. At the same time there is simple yet effective way to support such cases without the need to introduce some core changes. In case of any unmanageable write calls pipe fuse can just complete a multipart upload at its current state and initialize another one right after. Surely the approach performance highly depends on the number of such unmanageable writes as well as the target file size because for each of the writes a new multipart upload should be created.

tcibinan:
Resolves #1271 and relates to #1030.

The pull introduces support for multipart uploads reinitialization in case of unmanageable writes.

Also it includes changes previously introduced in #1264.

The pull request brings explicit implementation errors logging to pipe fuse. Previously, some of the exceptions were visible only in DEBUG mode.

Also pull request introduces caching for existence calls and preliminary invalidates caches for missing requested files.

### DIFF
--- a/pipe-cli/mount/pipefuse/buff.py
+++ b/pipe-cli/mount/pipefuse/buff.py
@@ -132,7 +132,7 @@ class BufferedFileSystemClient(FileSystemClientDecorator):
         buf_key = fh, path
         file_buf = self._read_file_buffs.get(buf_key)
         if not file_buf:
-            file_size = self.attrs(path).size
+            file_size = self._inner.attrs(path).size
             if not file_size:
                 return
             file_buf = self._new_read_buf(fh, path, file_size, offset, length)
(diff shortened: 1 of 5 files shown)


NShaforostov:
I've tried to verify. Works well.
Were tested:
- multi threads
- logging

mzueva (approved):

Issue #1299, opened 2020-08-19
Pipeline run store RESUMED_RUN env in cp_env.sh


SilinPavel:
Currently RESUMED_RUN env stored in cp_env.sh it can lead to wrong behavior of launch sh while resuming

SilinPavel:
closed as not implemented

PR #1300, opened 2020-08-19
issue 1299 don't store RESUMED_RUN env

SilinPavel:
Related to #1299


sidoruka (changes requested):
I'm against removing the `RESUMED_RUN` from the user's session variables. As some of the apps (that are configured within a run) can depend on that.
If the value of this variable gets corrupted - let's diagnose the real cause and fix that specific issue.

SilinPavel:
closed since main issue can be fixed with other solution

Issue #1268, opened 2020-08-12
`pipe storage cp` windows with the GCS - sslv3 error


sidoruka:
**Describe the bug**
We observe the issue with copying data to/from the GCS using the Windows workstation:
```
C:\pipe>pipe.exe storage cp file.txt gs://cloud-pipeline-gcs/file.txt
Error: HTTPSConnectionPool(host='www.googleapis.com', port=443): Max retries exc
eeded with url: /storage/v1/b/cloud-pipeline-gcs/o?projection=noAcl&prefix=file.txt
&delimiter=%2F&versions=False (Caused by SSLError(SSLError("bad handshake: Er
ror([('SSL routines', 'ssl3_read_bytes', 'tlsv1 alert protocol version')],)",),)
```

**Additional context**
The windows `Internet Options` allow the following protocols:
```
SSL 3
TLS 1.0
SSL 1.1
SSL 1.2
```
It might be that the `pipe` uses these settings and the GCP rejects SSLv3 requests

sidoruka:
Fix shall be backported to `release/0.16`, if the issue is caused by the `pipe`

mzueva:
cherry pick 832b6a863bc6817396eeda1bfc1c80a8aa571ba4

PR #1311, opened 2020-08-21
`pipe storage cp` windows with the GCS - sslv3 error

ekazachkova:
This PR provides fix for issue #1268
 The current fix contains implementation for `S3` and `GS` providers.

### DIFF
--- a/pipe-cli/src/utilities/storage/gs.py
+++ b/pipe-cli/src/utilities/storage/gs.py
@@ -726,11 +726,11 @@ class TlsAdapter(HTTPAdapter):
         super(TlsAdapter, self).__init__(**kwargs)
 
     def init_poolmanager(self, *pool_args, **pool_kwargs):
-        ctx = ssl_.create_urllib3_context(ssl.PROTOCOL_TLS)
-
-        ctx.options |= ssl.OP_NO_SSLv2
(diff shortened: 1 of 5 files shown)


pipe-cli/src/utilities/storage/gs.py
```diff
@@ -651,6 +662,51 @@ def apply(self, headers, token=None):
         headers['authorization'] = 'Bearer {}'.format(_helpers.from_bytes(self.temporary_credentials.session_token))
 
 
+class VerifiedHTTPSConnectionWithHeaders(VerifiedHTTPSConnection):
```
mzueva, line 665: Can't we use the same code for VerifiedHTTPSConnectionWithHeaders for AWS and GCP?

mzueva (changes requested):

pipe-cli/src/utilities/storage/gs.py
ekazachkova: @mzueva  The version of the `boto3` used by `pipe cli` contains its own implementation of the `urllib3` library. This way, the `GS` provider uses the class `VerifiedHTTPSConnection` from `urllib3.connection` but `VerifiedHTTPSConnection` from `botocore.vendored.requests.packages.urllib3.connection` is used by `S3`.

PR #1330, opened 2020-08-26
Support writing with gaps in pipe fuse

### LINKED ISSUE #1329: Support writing with gaps in pipe fuse
**Background**

Currently pipe fuse doesn't handle writing with gaps. For instance if writing to empty file is started from any position but 0 then the resulting file may be corrupted.

**Approach**

Pipe fuse shall handle writing with gaps in order to become more POSIX compliant.

tcibinan:
Resolves #1329.

The pull request introduces support for writing with gaps.

### DIFF
--- a/pipe-cli/mount/pipefuse/mpu.py
+++ b/pipe-cli/mount/pipefuse/mpu.py
@@ -355,12 +355,13 @@ class OutOfBoundsFillingMultipartCopyUpload(MultipartUploadDecorator):
 
     def upload_copy_part(self, start, end, offset=None, part_number=None):
         if self._original_size <= start:
-            logging.debug('Filling out of bounds copy upload part %s whole %d-%d for %s' % (part_number, start, end, self.path))
-            modified_buf = bytearray(end - start)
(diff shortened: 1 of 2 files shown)


mzueva (approved):

NShaforostov:
Verified. Works well.
Were tested:
- multi threads

tcibinan:
The changes also help with some specific workloads f.e. saving `.mat` and `.fig` files in Matlab.

To make sure that everything works fine the following steps can be performed.
```matlab
% Generate mock environment
A = rand(2000000, 1);

% Save the environment in data.mat file
save('data.mat');

% Load the environment back to check that it is not corrupted
% If there are no errors then the file is not corrupted
load('data.mat');

% Generate the figure
figure;
plot(1:3000000,1:3000000);

% Save the opened figure using the corresponding save button to data.fig
% Wait for a minute to make sure that the figure is saved, close the figure
% Try to open the saved data.fig file using double click to check that it is not corrupted
% If the figure is opened then the file is not corrupted
```

PR #1344, opened 2020-08-31
ES billing modifications for aggregation performance

Wedds:
This PR brings changes to `BillingManager`, which allows receiving aggregation results faster.

- Count unique runs with `ValueCount` aggregation instead of` Terms`
- Use `filter_path` param for grouping requests to reduce the size of JSON response from ES (only aggregation results are received without search details)


tcibinan:
Cherry-picked to `release/0.16` via b176fd9bf6e8bfd4ba50e79c4d102edac9845f50.

PR #1364, opened 2020-09-03
Improve pipe fuse performance and memory consumption for big directories

### LINKED ISSUE #1328: Caches and buffers should be explicitly cleaned up in pipe fuse in case of low-level errors
**Background**

Currently pipe fuse doesn't clean up caches and buffers in case of some low-level error. It does the cleaning on a subsequent flush call which is usually performed by libfuse. Nevertheless in some specific cases such flush calls won't happen.

**Approach**

In order to have better error handling explicit caches and buffers cleaning should be performed in case of any low-level errors.

### LINKED ISSUE #1338: PipeFUSE sometimes fails when creating or listing of a large number (~100 000) of files
**Describe the bug**
When we create or list a large number (~100 000) of files in PipeFUSE in S3 storage, at some point, the storage stops working correctly.  In all cases, the common mistake:
`cannot access '/storage_path/': Transport endpoint is not connected`

**To Reproduce**
1. Run a pipeline with mounted PipeFUSE in S3 storage
2. Create a large number (~100 000) of files in the created pipeline. Example command:
`
for j in {1..100000}; do 
 	head -c 500kB </dev/urandom > /path-to-storage/small/randfile$j.txt
done
`
3. List `ls /path-to-storage/small/ ` the created files in the storage, if the `2` step was a success. 

**Screenshots**
image

image

image

**Environment:**
 - Cloud Provider: **AWS**
 - Version **0.17.0.6626**
 - Browser: **chrome**

### LINKED ISSUE #1358: Invalidate caches partially in pipe fuse
**Background**

Currently any terminal operation (mv, mkdir, flush and etc.) in pipe fuse causes file's directory cache invalidation. It works fine for small directories but performs badly for directories with thousands of files. Lots of operations on small files (creation lots of small files) end up with constant directory listing which is time-consuming actions for bigger directories.

**Approach**

Most of the terminal operations allows to perform partial caches invalidation. For example if mv command is called than we can just move cached object from source cached listing to a target one without the need to list both directories.

### LINKED ISSUE #1359: Improve buffers merging performance in pipe fuse
**Background**

According to some resources such as [this one](https://www.guyrutenberg.com/2020/04/04/fast-bytes-concatenation-in-python/) plain bytearrays concatenation in python performs several times faster than bytearrays preallocation.

It is even possible to reduce memory consumption if bytearrays concatenation is used. To do so already concatenated buffers can be deleted explicitly to possibly freed up some space.

**Approach**

Bytearrays concatenation should be used for all buffers merging-like operations.

tcibinan:
Relates to #1338. Resolves #1328, resolves #1358 and resolves #1359.

The pull request resolves several issues which tend to improve both performance and memory consumption of pipe fuse. It seems to reduce the overall workload for big directories processing and possibly resolves issue #1338.

### DIFF
--- a/pipe-cli/mount/pipefuse/buff.py
+++ b/pipe-cli/mount/pipefuse/buff.py
@@ -28,7 +28,7 @@ class _FileBuffer(object):
         self._offset = offset
         self._current_offset = self._offset
         self._capacity = capacity
-        self._buffs = []
+        self._buf = bytearray()
 
     @property
     def offset(self):
@@ -43,7 +43,7 @@ class _FileBuffer(object):
         return self._capacity
 
     def append(self, buf):
-        self._buffs.append(buf)
+        self._buf += buf
         self._current_offset += len(buf)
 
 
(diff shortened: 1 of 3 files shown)


tcibinan:
Merged via #1187.

Issue #1340, opened 2020-08-28
Embed the "Cloud Pipeline File Browser" (electron app) into the API and expose to the users


sidoruka:
As now we have the "Cloud Pipeline File Browser" application implemented ([gui-cloud-pipeline-webdav-client](https://github.com/epam/cloud-pipeline/tree/gui-cloud-pipeline-webdav-client)) which can be used a local data storage browser. We'd need to include it into the Cloud Pipeline distribution (in the same manner as the "pipe").

This would greatly simplify the data access from the Windows workstation, which have lots of issues with the native webdav client.

rodichenko:
@ekazachkova @mzueva to build a client

for linux:
```
npm install
npm run package:linux
# distributives will be located at out/cloud-pipeline-webdav-client-linux-x64
```
for windows:
```
npm install
npm run package:win32
# distributives will be located at out/cloud-pipeline-webdav-client-win32-ia32
```
for macOS:
```
npm install
npm run package:macos
# distributives will be located at out/cloud-pipeline-webdav-client-darwin-x64
```

If you build a client for windows platform on *non-windows* machine, you should pre-install [wine](https://www.winehq.org).

mzueva:
@sidoruka @NShaforostov @rodichenko `Cloud Pipeline File Browser` may be downloaded via `/webdav-linux.tar.gz, /webdav-win32.zip, /webdav-win64.zip` paths.

PR #1384, opened 2020-09-09
Embed the "Cloud Pipeline File Browser" (electron app) into the API and expose to the users

ekazachkova:
This PR provides implementation for issue #1340

This PR contains build scripts for Linux and Windows OS.


api/src/main/java/com/epam/pipeline/app/JWTSecurityConfiguration.java
```diff
@@ -142,7 +142,8 @@ protected String getSecuredResources() {
             "/launch.sh", "/PipelineCLI.tar.gz",
             "/pipe-common.tar.gz", "/commit-run-scripts/**", "/pipe",
             "/fsbrowser.tar.gz", "/error", "/error/**", "/pipe.zip", "/pipe.tar.gz",
-            "/pipe-el6", "/pipe-el6.tar.gz"
+            "/pipe-el6", "/pipe-el6.tar.gz",
```
mzueva: Could you please move this list to application.properties and use it in both security configurations

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/app/JWTSecurityConfiguration.java
ekazachkova: added a new application property `static.common.scripts.exclude` which contains a list of paths to exclude

mzueva (approved):

Issue #1387, opened 2020-09-10
Introduce "ROLE_USER_READER" role with RO access to users/groups/roles info


sidoruka:
Please add a "built-in" role `ROLE_USER_READER` which shall allow read-only access to the API endpoints, responsible for the users, groups, roles information.
Only "general" user/groups information is required: name/email/etc. Metadata is not needed for this use-case

sidoruka:
@mzueva please backport this to `release/0.16`

mzueva:
@sidoruka backported to `release/0.16`

NShaforostov:
Verified. Docs were updated via #1423

PR #1390, opened 2020-09-10
Issue 1387 Role `USER_READER`

Wedds:
This PR is related to issue #1387

It brings a new predefined role `ROLE_USER_READER`, that grants read-only access to information regarding users, groups, roles.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/security/acl/AclExpressions.java
+++ b/api/src/main/java/com/epam/pipeline/security/acl/AclExpressions.java
@@ -22,7 +22,9 @@ public final class AclExpressions {
 
     public static final String ADMIN_ONLY = "hasRole('ADMIN')";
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/epam/pipeline/security/acl/AclExpressions.java
```diff
@@ -22,6 +22,8 @@
 
     public static final String ADMIN_ONLY = "hasRole('ADMIN')";
 
+    public static final String OR_USER_READER = OR + "hasRole('ROLE_USER_READER')";
```
mzueva: You can use shorthand for role in `hasRole` expression -> `USER_READER`

mzueva (changes requested):

mzueva (approved):

Issue #1385, opened 2020-09-09
Advanced configuration of notifications/automatic actions for "idle" runs


NShaforostov:
**Background**
When the platform is used by many users, there a situation can occur when a large count of "idle" runs is being slept in paused state for a real long time, users can forget about them and the stand can become "littered" by such forgotten runs.
In this case, to "clear" the stand - admins should manually clarify - whether a specific run needs for a user or could be terminated.

It would be convenient to have an automatic mechanism to rule such situations.
The following scenario for management of the "idle" runs should be supported:
_**Suppose, there is a run that has been idle for some time.**_ 
1. After a certain time period, if the run has still "idle" state, the corresponding notification (about the "idle" run and possibility of its automatic pause) should be sent to the owner user.
2. After an additional time period from the first notification and if the run has still "idle" state, the new notification should be sent to the owner user. The notification content can be different from the first one. This action can be repeated several times, if the run state doesn't change during the time.
3. If the user still takes no actions - after `N` notifications, the set by preferences action (`PAUSE`/`STOP`) is being performed with the run. The owner user receives the corresponding notification.
_**If the run was paused:**_
4. After a certain time period, if the user still takes no actions and the run is paused, the new notification (about the "long paused" run and possibility of its automatic termination) should be sent to the owner user.
5. After an additional time period from the previous notification and if the run has still "paused" state, the notification should be sent again to the owner user. The notification content can be different from the previous one. This action can be repeated several times, if the run state doesn't change during the time.
6. If the user still takes no actions - after `M` notifications, the paused run should be terminated. The owner user receives the corresponding notification.

**Approach**
Currently, scenario described above can be performed on the platform only partly.
Suggest to expand the existing functionality with the following features:
- Add a new notification type - _**LONG\_PAUSED**_ - the notification shall be sent when the run is in "PAUSED" state for a long time. 
- This new notification type shall respect the following parameters (similar to _**IDLE\_RUN**_):
  - Notification time threshold, once elapsed - notification is triggered (e.g. `system.max.long.paused.timeout.minutes`)
  - Action time threshold, once elapsed - a certain action is triggered (e.g. `system.long.paused.action.timeout.minutes`)
  - Action type - one of two keywords: `NOTIFY` or `STOP` (e.g. `system.long.paused.action`)
    - If `STOP` is used - the corresponding run shall be terminated
    - If `NOTIFY` is used - no action is performed - keep sending emails according to the thresholds
[...]

PR #1410, opened 2020-09-17
Advanced configuration of notifications/automatic actions for "idle" runs

ekazachkova:
This PR provides implementation for issue #1385 

- a new notification type `LONG_PAUSED` - indicates that run has been paused for a long time and notification shell be sent to users periodically with a certain time interval.
-  a new notification type `LONG_PAUSED_STOPPED` - indicates that run has been paused for a long time and shell be stopped after timeout.
- a new system property `system.long.paused.action` - indicated long paused runs action strategy: notify or stop
- a new email template files for a new notification types

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
@@ -32,6 +32,7 @@ import java.util.stream.Stream;
 import javax.annotation.PostConstruct;
 
 import com.epam.pipeline.entity.cluster.monitoring.ELKUsageMetric;
+import com.epam.pipeline.entity.monitoring.LongPausedRunAction;
 import com.epam.pipeline.entity.pipeline.StopServerlessRun;
 import com.epam.pipeline.entity.pipeline.TaskStatus;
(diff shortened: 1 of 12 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -434,5 +436,30 @@ public void updateInstanceMap(Map<String, InstanceType> types) {
...
+        }
+
+        private void processLongPausedRuns(final List<PipelineRun> pausedRuns, final LongPausedRunAction action) {
+            if (CollectionUtils.isEmpty(pausedRuns)) {
+                return;
+            }
+
+            if (LongPausedRunAction.STOP.equals(action)) {
+                ListUtils.emptyIfNull(notificationManager.notifyLongPausedRunsBeforeStop(pausedRuns))
+                        .forEach(run -> pipelineRunManager.stop(run.getId()));
```
mzueva: Just `stop` is not enough in this case as paused instance will stay in the cloud: I'd suggest to use `NodesManager.terminateNode` method here.

mzueva (changes requested):

mzueva (approved):

PR #1435, opened 2020-09-24
Support newest data transfer service in api

tcibinan:
Relates to #1164.

The pull request contains changes originally presented in #1164. It only introduces changes in api module. *Notice* that the pull request should be merged after #1164 because it depends on some core classes added in the original pull request.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/exception/SystemPreferenceNotSetException.java
+++ b/api/src/main/java/com/epam/pipeline/exception/SystemPreferenceNotSetException.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


api/src/main/java/com/epam/pipeline/exception/SystemPreferenceNotSetException.java
```diff
@@ -0,0 +1,13 @@
+package com.epam.pipeline.exception;
```
mzueva, line 17: Missing license

api/src/main/java/com/epam/pipeline/manager/dts/DtsClient.java
```diff
@@ -19,6 +19,7 @@
 import com.epam.pipeline.controller.Result;
 import com.epam.pipeline.entity.dts.DtsClusterConfiguration;
 import com.epam.pipeline.entity.dts.DtsDataStorageListing;
+import com.epam.pipeline.entity.dts.DtsDataStorageListingRequest;
```
mzueva, line 22: This class shall be included into PR

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/exception/SystemPreferenceNotSetException.java
tcibinan: Added.

api/src/main/java/com/epam/pipeline/manager/dts/DtsClient.java
tcibinan: Oops! Thanks for catching this.

PR #1448, opened 2020-09-30
Fallback to default user while data transferring

tcibinan:
Relates to #1164.

The pull request allows data transfer service to fallback to default user in case of any error using impersonated user.


tcibinan:
Rejected because it introduces several security issues.

Issue #1404, opened 2020-09-15
Implement tests for ACL layer


mzueva:
**Background**
Currently we have a bunch of integration tests testing ACL security configuration and corresponding classes, but most of this tests use full Spring `ApplicationContext`, which takes up to two minutes to initialise and are marked with `@DirtiesContext` annotation which forces creation of a new `ApplicationContext` for each test. Such configuration leads to unacceptable run time of tests and low test coverage as we cannot afford a large number of such tests.

**Approach**
In order to reduce runtime of tests we need to create isolated tests aiming only one application layer: ACL configuration and corresponding `APIService` classes.
- Provide test context configuration for ACL layer
- Implement tests for the following packages:
    - billing
    - cluster
    - configuration
    - contextual
    - datastorage
    - dts
    - entity
    - issue
    - log
    - metadata
    - notification
    - pipeline
    - preference
    - region
    - search
    - security
    - user

mzueva:
@cryteq 

Next classes for tests:
- `DockerRegistryApiService`
- `ToolApiService`
- `ToolGroupApiService`

These classes shall be moved to `acl/docker` package. Existing `ToolApiServiceTest` shall be replaced with a new test.

PR #1460, opened 2020-10-05
Issue 1404 implemented test for acl layer region package

youkofan:
Implement tests for ACL layer (for region package) - #1404
Implemented tests for CloudRegionApiService.
Signed-off-by: Iurii_Kofanov youkofan@gmail.com


api/src/test/java/com/epam/pipeline/app/AclTestConfiguration.java
```diff
@@ -45,6 +45,7 @@
 import com.epam.pipeline.manager.pipeline.ToolManager;
 import com.epam.pipeline.manager.pipeline.runner.ConfigurationProviderManager;
 import com.epam.pipeline.manager.pipeline.runner.ConfigurationRunner;
+import com.epam.pipeline.manager.region.CloudRegionManager;
```
ekazachkova, line 47: Why should we add changes to this class? Does this configuration affect new tests?

api/src/test/java/com/epam/pipeline/test/acl/AclTestBeans.java
```diff
@@ -179,24 +210,15 @@
...
-
-    @MockBean
-    protected PermissionEvaluator mockPermissionEvaluator;
-
     @MockBean
     protected MessageHelper mockMessageHelper;
 
-    @MockBean
+    @Autowired
     protected CheckPermissionHelper mockPermissionHelper;
```
ekazachkova: To avoid confusion, let's give the fields actual names, depending on their current mock state

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+    @WithMockUser(username = SIMPLE_USER_ROLE)
+    public void shouldReturnAllCloudRegionsWhenPermissionIsGranted() {
+        initAclEntity(region,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.READ.getMask())));
+        doReturn(clouds).when(mockCloudRegionManager).loadAll();
+
+        final List<? extends AbstractCloudRegion> abstractCloudRegions = cloudRegionApiService.loadAll();
+
+        assertThat(abstractCloudRegions).isEqualTo(clouds);
+        assertThat(abstractCloudRegions.get(0).getId()).isEqualTo(clouds.get(0).getId());
```
ekazachkova: Lets make a new expected object that will not be mutated during the test execution

ekazachkova (changes requested):
Also, lets run the following actions before each push:
- `compileTestJava`
- `checkstyleTest`
- `pmdTest`
- `test`

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER_ROLE)
+    public void shouldReturnAllCloudRegionsWhenPermissionIsGranted() {
```
ekazachkova, line 176: Lets add test to the following case:
- create two regions: region1, region2
- grant permissions for user only for region1
- mock `mockCloudRegionManager` to return two regions
- check that user gets only region1

api/src/test/java/com/epam/pipeline/test/acl/AclTestBeans.java
youkofan: Thank you, fixed.

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: Done. I hope that i understand you correctly.

api/src/test/java/com/epam/pipeline/app/AclTestConfiguration.java
youkofan: If we don't add this manager, tests in the RunApiServiceTest will be failed. So, if I may say so, new tests affect the old tests and their configuration. But thanks to your remark, I found a redundant bin in this configuration, which I added earlier.

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: Test is added.

youkofan:
All checks and test are passed (i find incorrect import in the ControllerTestBeans.class as i moved CloudRegionApiService.class)

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+        regionWithoutPermission = new AwsRegion();
+        regionWithoutPermission.setId(2L);
+        regionWithoutPermission.setName("SIMPLE_USER");
+        regionWithoutPermission.setOwner(OWNER_USER);
+
+        singleRegionList = new ArrayList<>();
+        singleRegionList.add(region);
+
+        twoRegionsList = new ArrayList<>();
```
ekazachkova: To my mind it would be cleaner to create this object directly in test since `twoRegionList` used in one test only. No need to create this object for each test.

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,261 @@
...
+        region.setId(1L);
+        region.setName("OWNER");
+        region.setOwner(OWNER_USER);
+
+        regionWithoutPermission = new AwsRegion();
+        regionWithoutPermission.setId(2L);
+        regionWithoutPermission.setName("SIMPLE_USER");
+        regionWithoutPermission.setOwner(OWNER_USER);
+
+        singleRegionList = new ArrayList<>();
```
ekazachkova, line 69: Could we use `Collections.singletonList` here?

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,261 @@
...
+        singleRegionList.add(region);
+
+        twoRegionsList = new ArrayList<>();
+        twoRegionsList.add(region);
+        twoRegionsList.add(regionWithoutPermission);
+
+        cloudRegionInfoList = new ArrayList<>();
+        cloudRegionInfoList.add(new CloudRegionInfo(region));
```
ekazachkova: Could we use `Collections.singletonList` here?

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER_ROLE)
+    public void shouldReturnAllCloudRegionsWhenPermissionIsGranted() {
+        initAclEntity(region,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.READ.getMask())));
+        doReturn(singleRegionList).when(mockCloudRegionManager).loadAll();
+
+        final List<? extends AbstractCloudRegion> abstractCloudRegions = cloudRegionApiService.loadAll();
+        System.out.println(abstractCloudRegions);
```
ekazachkova: Cleanup please

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldReturnAllCloudRegionsForAdmin() {
+        doReturn(singleRegionList).when(mockCloudRegionManager).loadAll();
+
+        final List<? extends AbstractCloudRegion> abstractCloudRegions = cloudRegionApiService.loadAll();
+        final AbstractCloudRegion resultRegion = abstractCloudRegions.get(0);
+
+        assertThat(abstractCloudRegions).isEqualTo(singleRegionList);
```
ekazachkova: The `singleRegionList` object was mutated during `@PostFilter` operation. This way this assertion is useless and we need another one. For example, let's check elements count (in this case we shall move `abstractCloudRegions.get(0)` after checking size).

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,261 @@
...
+    public void shouldReturnListWithRegionsPermissionsForWhichIsGranted() {
+        initAclEntity(region,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.READ.getMask())));
+        initAclEntity(regionWithoutPermission,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.NO_READ.getMask())));
+        doReturn(twoRegionsList).when(mockCloudRegionManager).loadAll();
+        
+        final List<? extends AbstractCloudRegion> abstractCloudRegions = cloudRegionApiService.loadAll();
+
+        assertThat(abstractCloudRegions.size()).isEqualTo(1);
```
ekazachkova: Let's add check chat returned region id as expected.

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
ekazachkova: Also, we can use `Arrays.asList` here

ekazachkova (changes requested):

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: `Collections.singletonList`, as well as the list created with the help of `Arrays.asList`, call UnsupportedOperationException when trying to change them (the stack trace comes to the `remove `method in the `AbstractList `class).

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: `twoRegionList` moved to the method that calls them (also moved `regionWithoutPermission`object as it used at once too). 
About using` Arrays.asList` please see following comment.

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: Done

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: Done

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: Done

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: Done

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER_ROLE)
+    public void shouldReturnAllCloudRegionsWhenPermissionIsGranted() {
+        initAclEntity(region,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.READ.getMask())));
+        doReturn(singleRegionList).when(mockCloudRegionManager).loadAll();
+
+        final List<AwsRegion> resultList = (List<AwsRegion>) cloudRegionApiService.loadAll();
```
ekazachkova: Could you please change object casting to something like:
```
final List<? extends AbstractCloudRegion> resultList = cloudRegionApiService.loadAll();

assertThat(resultList).hasSize(1);
assertThat(resultList.get(0)).isEqualTo(region);
```

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Test(expected = AccessDeniedException.class)
+    @WithMockUser(username = SIMPLE_USER_ROLE)
+    public void shouldFailReturningCloudRegionWithoutPermission() {
+        initAclEntity(region);
+        doReturn(region).when(mockCloudRegionManager).load(region.getId());
+
+        cloudRegionApiService.load(region.getId());
+
+        assertThat(region).isNull();
```
ekazachkova: Will this code ever get executed?

ekazachkova (changes requested):

ekazachkova (approved):

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,252 @@
...
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldAllowLoadRegionsInfoForAdmin() {
+        doReturn(cloudRegionInfoList).when(mockCloudRegionManager).loadAllRegionsInfo();
+
+        assertThat(cloudRegionApiService.loadAllRegionsInfo()).isEqualTo(cloudRegionInfoList);
+    }
+
+    @Test
+    @WithMockUser(roles = GENERAL_USER_ROLE)
```
mzueva, line 84: Can't we use here `SIMPLE_USER_ROLE`? I think it is better to stuck to one role if both have equal meaning

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
```diff
@@ -0,0 +1,252 @@
...
+        doReturn(singleRegionList).when(mockCloudRegionManager).loadAll();
+
+        final List<? extends AbstractCloudRegion> resultList = cloudRegionApiService.loadAll();
+
+        assertThat(resultList).hasSize(1);
+        assertThat(resultList.get(0)).isEqualTo(region);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER_ROLE)
```
mzueva: Please, do not mix `role` and `username`, let's introduce a separate constant for `username`

api/src/test/java/com/epam/pipeline/test/acl/AclTestConfiguration.java
```diff
@@ -37,10 +39,18 @@
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
+@ContextConfiguration(classes = {AclTestBeans.class, AclSecurityConfiguration.class})
+@ComponentScan(
```
mzueva: It seems that you can just use `basePackages = {"com.epam.pipeline.security.acl", "com.epam.pipeline.manager.security"}` without excludes

mzueva (changes requested):

api/src/test/java/com/epam/pipeline/acl/region/CloudRegionApiServiceTest.java
youkofan: The method that is tested here has the condition ADMIN_OR_GENERAL_USER in PreAuthorize. I thought the General User had access as opposed to the Simple User. So I made a separate variable for General user.

mzueva: `General` user is any user with role named `USER` or `ROLE_USER`, so there is no actual difference between General user and Simple user

Issue #1374, opened 2020-09-08
Implement parallel transferring in pipe cli for Google storages


tcibinan:
**Background**

Currently pipe cli uses default single-threaded approach for data transferring in Google Cloud storages.

**Approach**

Nevertheless it would be quite useful to have support for parallel transferring the same way it is done for AWS S3 in boto3 library.

tcibinan:
# Benchmarks

Average values of 5 attempts are shown for each operation below.

| Operation | Size | Before, s | After, s | Faster, times |
| --- | --- | --- | --- | -- |
| Upload | 10M | 1 | 1 | - |
| Upload | 100M | 2 | 2 | - |
| Upload | 1G | 8 | 5 | ▲ 1.6 |
| Upload | 10G | 66 | 35 | ▲ 1.89 |
| Upload | 100G | 674 | 336 | ▲ 2.01 |
| Upload | 200G | 1422 | 670 | ▲ 2.12 |
| Download | 10M | 1 | 1 | - |
| Download | 100M | 2 | 2 | - |
| Download | 1G | 8 | 7 | ▲ 1.14 |
| Download | 10G | 63 | 53 | ▲ 1.19 |
| Download | 100G | 656 | 539 | ▲ 1.22 |
| Download | 200G | 1272  | 1000 | ▲ 1.27 |

## Environment

OS: Ubuntu 18.04
Disk: 1000GB
Instance: n2-standard-4 (cpu: 4, ram: 16)
Cloud region: europe-west1-b
Deployment cloud provider: GCP
Before commit: bc7f5fce8f6e9d77ce1aa1172a6ba5c6b31e0148
After commit: 8de45b4f73da5cbaea5f31a97a6c325691b60d59

<details><summary>Source code</summary>
<p>

```
./measure_performance.sh > measured_performance.tsv
python3 collect_summary.py
```

<details><summary>measure_performance.sh</summary>
<p>

```bash
executables=("./pipe_old" "./pipe_new")
operations=("upload" "download")
sizes=(10M 100M 1G 10G 100G 200G)
destination="gs://fuse-dev-storage-1"
attempts=5

for size in "${sizes[@]}"
do
  file=file_$size
  fallocate -l $size $file
done

echo -e "operation\texecutable\tsize\telapsed, s"
for executable in "${executables[@]}"
do
  for operation in "${operations[@]}"
  do
    if [[ $operation == "upload" ]]
    then
      for size in "${sizes[@]}"
      do
[...]

PR #1468, opened 2020-10-06
Repair parallel transfer in pipe cli for gcs on windows

tcibinan:
Relates to #1374.

The pull request brings two fixes for gcs parallel transfer implementation:
- First one replaces python 2 `iterator.next()` call with python 2 and 3 `next(iterator)` call. It previously broke pipe cli win build where python target is 3 rather than 2 that is used for pipe cli linux build.
- Second one simplifies proxies resolving mechanism for gcs in order to avoid `pypac` library calls in multithreading mode which tends to fail on windows.

### DIFF
--- a/pipe-cli/src/utilities/storage/gs.py
+++ b/pipe-cli/src/utilities/storage/gs.py
@@ -934,21 +934,13 @@ class ProxyConnectWithHeadersHTTPSAdapter(HTTPAdapter):
         return manager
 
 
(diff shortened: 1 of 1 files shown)


mzueva (approved):

tcibinan:
Cherry-picked to `release/0.16` via 0d8dce3e8c1c09dd67f803d52c2b64414758af5a.

Issue #1469, opened 2020-10-06
Allow to transfer objects between AWS regions using pipe cp/mv commands


mzueva:
**Background**
At the moment `pipe cp/mv` commands allow to transfer objects only within one AWS region. It would be useful to allow object transfer between regions since AWS provides such functionality.

mzueva:
@sidoruka backported to `release/0.16`

ekazachkova:
@sidoruka  There are some benchmarks for this feature. The tests were performed with file 1.06 GB.

| case  | aws s3 cp | pipe storage cp |
| ------------- | ------------- | ------------- |
| one region (`eu-central-1`)  | 4 sec  | 7 sec |
| between regions (`eu-central-1` to `us-east-1`)  | 57 sec  | 58 sec |

NShaforostov:
Verified. Works well

NShaforostov:
**Docs** were updated via #1516

NShaforostov:
**Test case** was created by af3dc801a3a64a2b152308975e1c3f2e8f3dac1a and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/testcases/UI/Data_storages/1469.md).

NShaforostov:
**e2e-tests** were implemented via #1574.

NShaforostov:
Daily e2e-tests results are located and updated [here](https://github.com/epam/cloud-pipeline/blob/e2e-reports/e2e/reports/AWS/test/classes/com.epam.pipeline.autotests.DataStoragesCLITest.html) (for `AWS`).

PR #1479, opened 2020-10-07
Allow to transfer objects between AWS regions using pipe cp/mv commands

ekazachkova:
This PR provides implementation for issue #1469 

-  multi-region support added for AWS provider
-  multi-region actions disabled for AZURE and GCP

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cloud/TemporaryCredentialsManagerImpl.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cloud/TemporaryCredentialsManagerImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 EPAM Systems, Inc. (https://www.epam.com/)
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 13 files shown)


api/src/main/java/com/epam/pipeline/manager/cloud/TemporaryCredentialsManagerImpl.java
```diff
@@ -56,32 +59,48 @@ public TemporaryCredentialsManagerImpl(final List<TemporaryCredentialsGenerator>
...
+        final AbstractDataStorage loadedDataStorage = storagesById.get(action.getId());
+        action.setBucketName(loadedDataStorage.getRoot());
+        action.setPath(loadedDataStorage.getPath());
+    }
+
+    private AbstractDataStorage verifyAllTypesAreSameAngGetStorage(final List<AbstractDataStorage> storages) {
+        Assert.state(storages.stream()
+                .map(AbstractDataStorage::getType)
+                .distinct()
+                .count() <= 1, "Storage types shall be the same");
```
mzueva: Could you please move all error messages (in `Assert` methods to message constants)

ekazachkova: done

PR #1481, opened 2020-10-08
Issue #1404: implemented tests for ClusterApiService

youkofan:
First variant of tests for the first method (with @PostFilter) in the ClusterApiService
Issue - https://github.com/epam/cloud-pipeline/issues/1404

### DIFF
--- a/api/src/main/java/com/epam/pipeline/entity/cluster/NodeInstance.java
+++ b/api/src/main/java/com/epam/pipeline/entity/cluster/NodeInstance.java
@@ -45,7 +45,6 @@ import static com.epam.pipeline.manager.cluster.KubernetesConstants.RUN_ID_LABEL
 public class NodeInstance extends AbstractSecuredEntity {
 
     private UUID uid;
-    private String name;
     private String creationTimestamp;
     private List<NodeInstanceAddress> addresses;
 
--- a/api/src/main/java/com/epam/pipeline/entity/cluster/NodeInstance.java
+++ b/api/src/main/java/com/epam/pipeline/entity/cluster/NodeInstance.java
@@ -45,6 +45,7 @@ import static com.epam.pipeline.manager.cluster.KubernetesConstants.RUN_ID_LABEL
 public class NodeInstance extends AbstractSecuredEntity {
 
     private UUID uid;
+    private String name;
     private String creationTimestamp;
     private List<NodeInstanceAddress> addresses;
 
--- a/api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
+++ b/api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
@@ -50,6 +50,8 @@ import java.time.LocalDateTime;
 import java.util.Collections;
 import java.util.List;
 
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID_2;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
 import static com.epam.pipeline.util.CustomAssertions.assertThrows;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -59,9 +61,9 @@ import static org.mockito.Mockito.doReturn;
 public class ClusterApiServiceTest extends AbstractAclTest {
 
     private final FilterPodsRequest filterPodsRequest = NodeCreatorUtils.getDefaultFilterPodsRequest();
-    private final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, SIMPLE_USER);
-    private final NodeInstance anotherNodeInstance = NodeCreatorUtils.getNodeInstance(2L, TEST_STRING);
-    private final PipelineRun pipelineRun = PipelineCreatorUtils.getPipelineRun(1L, SIMPLE_USER);
+    private final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(ID, SIMPLE_USER);
+    private final NodeInstance anotherNodeInstance = NodeCreatorUtils.getNodeInstance(ID_2, TEST_STRING);
+    private final PipelineRun pipelineRun = PipelineCreatorUtils.getPipelineRun(ID, SIMPLE_USER);
     private final FilterNodesVO filterNodesVO = NodeCreatorUtils.getDefaultFilterNodesVO();
     private final NodeDisk nodeDisk = NodeCreatorUtils.getDefaultNodeDisk();
     private final MonitoringStats monitoringStats = ClusterCreatorUtils.getMonitoringStats();
@@ -182,7 +184,7 @@ public class ClusterApiServiceTest extends AbstractAclTest {
     @Test
     @WithMockUser(username = SIMPLE_USER)
     public void shouldReturnNodeInstanceWhenPermissionIsGranted() {
-        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, TEST_STRING);
+        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(ID, TEST_STRING);
         initAclEntity(nodeInstance, AclPermission.READ);
         doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName());
         mockRun(pipelineRun);
@@ -215,7 +217,7 @@ public class ClusterApiServiceTest extends AbstractAclTest {
     @Test
     @WithMockUser(username = SIMPLE_USER)
     public void shouldReturnNodeThroughRequestWhenPermissionIsGranted() {
-        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, TEST_STRING);
+        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(ID, TEST_STRING);
         initAclEntity(nodeInstance, AclPermission.READ);
         mockRun(pipelineRun);
         mockNode(nodeInstance);
@@ -249,7 +251,7 @@ public class ClusterApiServiceTest extends AbstractAclTest {
     @Test
     @WithMockUser(username = SIMPLE_USER)
     public void shouldTerminateNodeWhenPermissionIsGranted() {
-        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, TEST_STRING);
+        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(ID, TEST_STRING);
         initAclEntity(nodeInstance, AclPermission.READ);
         mockNode(nodeInstance);
         mockRun(pipelineRun);
@@ -359,17 +361,17 @@ public class ClusterApiServiceTest extends AbstractAclTest {
     @Test
     @WithMockUser
     public void shouldReturnInstanceTypes() {
-        doReturn(instanceTypes).when(mockInstanceOfferManager).getAllowedInstanceTypes(1L, true);
+        doReturn(instanceTypes).when(mockInstanceOfferManager).getAllowedInstanceTypes(ID, true);
 
-        assertThat(clusterApiService.getAllowedInstanceTypes(1L, true)).isEqualTo(instanceTypes);
+        assertThat(clusterApiService.getAllowedInstanceTypes(ID, true)).isEqualTo(instanceTypes);
     }
 
     @Test
     @WithMockUser
     public void shouldReturnToolInstanceTypes() {
-        doReturn(instanceTypes).when(mockInstanceOfferManager).getAllowedToolInstanceTypes(1L, true);
+        doReturn(instanceTypes).when(mockInstanceOfferManager).getAllowedToolInstanceTypes(ID, true);
 
-        assertThat(clusterApiService.getAllowedToolInstanceTypes(1L, true)).isEqualTo(instanceTypes);
+        assertThat(clusterApiService.getAllowedToolInstanceTypes(ID, true)).isEqualTo(instanceTypes);
     }
 
     @Test
@@ -378,9 +380,9 @@ public class ClusterApiServiceTest extends AbstractAclTest {
         final AllowedInstanceAndPriceTypes allowedInstanceAndPriceTypes =
                 NodeCreatorUtils.getDefaultAllowedInstanceAndPriceTypes();
         doReturn(allowedInstanceAndPriceTypes).when(mockInstanceOfferManager)
-                .getAllowedInstanceAndPriceTypes(1L, 2L, true);
+                .getAllowedInstanceAndPriceTypes(ID, ID_2, true);
 
-        assertThat(clusterApiService.getAllowedInstanceAndPriceTypes(1L, 2L, true))
+        assertThat(clusterApiService.getAllowedInstanceAndPriceTypes(ID, ID_2, true))
                 .isEqualTo(allowedInstanceAndPriceTypes);
     }
 
--- a/api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
+++ b/api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
@@ -438,7 +438,7 @@ public class ClusterApiServiceTest extends AbstractAclTest {
         doReturn(pipelineRun).when(mockPipelineRunManager).loadPipelineRun(eq(pipelineRun.getId()));
     }
 
-    protected void mockUser() {
+    private void mockUser() {
         doReturn(SIMPLE_USER).when(mockAuthManager).getAuthorizedUser();
     }
 }
--- a/api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
+++ b/api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
@@ -24,21 +24,17 @@ import com.epam.pipeline.entity.cluster.MasterNode;
 import com.epam.pipeline.entity.cluster.NodeDisk;
 import com.epam.pipeline.entity.cluster.NodeInstance;
 import com.epam.pipeline.entity.cluster.monitoring.MonitoringStats;
-import com.epam.pipeline.entity.contextual.ContextualPreference;
 import com.epam.pipeline.entity.pipeline.PipelineRun;
 import com.epam.pipeline.manager.cluster.InstanceOfferManager;
 import com.epam.pipeline.manager.cluster.NodeDiskManager;
 import com.epam.pipeline.manager.cluster.NodesManager;
 import com.epam.pipeline.manager.cluster.performancemonitoring.UsageMonitoringManager;
-import com.epam.pipeline.manager.contextual.ContextualPreferenceManager;
 import com.epam.pipeline.manager.pipeline.PipelineRunManager;
-import com.epam.pipeline.manager.preference.SystemPreferences;
 import com.epam.pipeline.manager.security.AuthManager;
 import com.epam.pipeline.security.acl.AclPermission;
 import com.epam.pipeline.test.acl.AbstractAclTest;
+import com.epam.pipeline.test.creator.cluster.ClusterCreatorUtils;
 import com.epam.pipeline.test.creator.cluster.NodeCreatorUtils;
-import com.epam.pipeline.test.creator.contextual.ContextualPreferenceCreatorUtils;
-import com.epam.pipeline.test.creator.pipeline.PipelineCreatorUtils;
 import org.junit.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.access.AccessDeniedException;
@@ -53,6 +49,7 @@ import java.util.Collections;
 import java.util.List;
 
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+import static com.epam.pipeline.util.CustomAssertions.assertThrows;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doReturn;
@@ -60,16 +57,15 @@ import static org.mockito.Mockito.doReturn;
 public class ClusterApiServiceTest extends AbstractAclTest {
 
     private final FilterPodsRequest filterPodsRequest = NodeCreatorUtils.getDefaultFilterPodsRequest();
-    private final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstanceWithPermission();
-    private final NodeInstance nodeInstanceWithoutPermission = NodeCreatorUtils.getNodeInstanceWithoutPermission();
-    private final PipelineRun pipelineRun = PipelineCreatorUtils.getPipelineRunWithPermission();
-    private final ContextualPreference contextualPreference =
-            ContextualPreferenceCreatorUtils.getContextualPreference();
+    private final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, SIMPLE_USER);
+    private final NodeInstance nodeInstanceWithoutPermission = NodeCreatorUtils.getNodeInstance(2L, TEST_STRING);
+    private final PipelineRun pipelineRun = ClusterCreatorUtils.getPipelineRun(1L, SIMPLE_USER);
     private final FilterNodesVO filterNodesVO = NodeCreatorUtils.getDefaultFilterNodesVO();
     private final NodeDisk nodeDisk = NodeCreatorUtils.getDefaultNodeDisk();
-    private final MonitoringStats monitoringStats = ContextualPreferenceCreatorUtils.getMonitoringStats();
+    private final MonitoringStats monitoringStats = ClusterCreatorUtils.getMonitoringStats();
     private final InstanceType instanceType = NodeCreatorUtils.getDefaultInstanceType();
     private final InputStream inputStream = new ByteArrayInputStream(TEST_STRING.getBytes());
+
     private final List<NodeDisk> nodeDisks = Collections.singletonList(nodeDisk);
     private final List<InstanceType> instanceTypes = Collections.singletonList(instanceType);
     private final List<MonitoringStats> statsList = Collections.singletonList(monitoringStats);
@@ -84,9 +80,6 @@ public class ClusterApiServiceTest extends AbstractAclTest {
     private PipelineRunManager mockPipelineRunManager;
 
     @Autowired
-    private ContextualPreferenceManager mockContextualPreferenceManager;
-
-    @Autowired
     private AuthManager mockAuthManager;
 
     @Autowired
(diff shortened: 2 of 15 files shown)


api/src/test/java/com/epam/pipeline/app/AclTestConfiguration.java
```diff
@@ -144,6 +146,12 @@
     @MockBean
     protected CloudRegionManager mockCloudRegionManager;
 
+    @MockBean
+    protected NodeDiskManager mockNodeDiskManager;
+
+    @MockBean
+    UsageMonitoringManager mockUsageMonitoringManager;
```
ekazachkova: Could you please add access modifier?

api/src/test/java/com/epam/pipeline/test/acl/AclTestBeans.java
```diff
@@ -322,9 +438,9 @@ protected DataStorageManager mockDataStorageManager() {
     }
 
     @Bean
-    protected FolderCrudManager mockCrudManager() {
+    protected FolderCrudManager mockFolderCrudManager() {
         return Mockito.mock(FolderCrudManager.class);
-    }
+    } /////
```
ekazachkova: Cleanup please

api/src/test/java/com/epam/pipeline/test/acl/AclTestConfiguration.java
```diff
@@ -39,7 +41,11 @@
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @ContextConfiguration(classes = {AclTestBeans.class, AclSecurityConfiguration.class})
-@ComponentScan(basePackages = {"com.epam.pipeline.security.acl", "com.epam.pipeline.manager.security"})
+@Import({ContextualPreferenceHandler.class})
```
ekazachkova: Why we need such import?

api/src/test/java/com/epam/pipeline/test/acl/AclTestConfiguration.java
```diff
@@ -39,7 +41,11 @@
...
+@ComponentScan(basePackages = {"com.epam.pipeline.security.acl",
+        "com.epam.pipeline.security.run",
+        "com.epam.pipeline.manager.security",
+        "com.epam.pipeline.manager.user"})
```
ekazachkova: Why we need `UserManager` real bean?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,546 @@
...
+    private UsageMonitoringManager mockUsageMonitoringManager;
+
+    @Autowired
+    private NodeDiskManager mockNodeDiskManager;
+
+    @Autowired
+    private InstanceOfferManager mockInstanceOfferManager;
+
+    private final FilterPodsRequest filterPodsRequest = NodeCreatorUtils.getDefaultFilterPodsRequest();
+
```
ekazachkova: Let's remove empty lines between fields.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,546 @@
...
+        statsList = Collections.singletonList(monitoringStats);
+
+        instanceTypes = Collections.singletonList(instanceType);
+
+        nodeDisks = Collections.singletonList(nodeDisk);
+
+        pipelineRun.setId(1L);
+        pipelineRun.setOwner(SIMPLE_USER);
+
+        final PipelineRun pipelineRunWithoutPermission = NodeCreatorUtils.getPipelineRun();
```
ekazachkova: Let's create a new build method in creators package. This  method may accept `id`, `name`, `owner` arguments.

api/src/test/java/com/epam/pipeline/test/creator/cluster/NodeCreatorUtils.java
```diff
@@ -86,4 +99,20 @@ public static AllowedInstanceAndPriceTypes getDefaultAllowedInstanceAndPriceType
     public static NodeDisk getDefaultNodeDisk() {
         return new NodeDisk(ID, TEST_STRING, LDT);
     }
+
+    public static PipelineRun getPipelineRun() {
```
ekazachkova: Let's not confuse packages: `PipelineRun` is not in cluster package. We need to create a new package/class for `PipelineRun` object creating methods.

ekazachkova: The same for creators below

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,546 @@
...
+
+    private List<MonitoringStats> statsList;
+
+    private List<NodeInstance> singleNodeInstance;
+
+    private List<NodeInstance> twoNodeInstances;
+
+    @Before
+    public void setUp() {
+        statsList = Collections.singletonList(monitoringStats);
```
ekazachkova: Let's  inline test class field initiliazations and get rid of `setUp` method.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,546 @@
...
+        twoNodeInstances.add(nodeInstance);
+        twoNodeInstances.add(nodeInstanceWithoutPermission);
+    }
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldReturnListWithNodeInstancesForAdmin() {
+        doReturn(singleNodeInstance).when(mockNodesManager).getNodes();
+
+        List<NodeInstance> nodes = clusterApiService.getNodes();
```
ekazachkova: Do not forget final please

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,546 @@
...
+        doReturn(SIMPLE_USER).when(mockAuthManager).getAuthorizedUser();
+        initAclEntity(nodeInstance,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.READ.getMask())));
+        doReturn(contextualPreference).when(mockContextualPreferenceManager).search(
+                Collections.singletonList(SystemPreferences.RUN_VISIBILITY_POLICY.getKey()));
+        doReturn(singleNodeInstance).when(mockNodesManager).getNodes();
+
+        List<NodeInstance> nodes = clusterApiService.getNodes();
+
+        assertThat(nodes).hasSize(1);
```
ekazachkova: Could we use `contains` method here?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,546 @@
...
+
+    private final FilterNodesVO filterNodesVO = NodeCreatorUtils.getDefaultFilterNodesVO();
+
+    private final NodeDisk nodeDisk = NodeCreatorUtils.getDefaultNodeDisk();
+
+    private final MonitoringStats monitoringStats = NodeCreatorUtils.getMonitoringStats();
+
+    private final InstanceType instanceType = NodeCreatorUtils.getDefaultInstanceType();
+
+    private InputStream inputStream;
```
ekazachkova: Looks like this field was not initialized

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
ekazachkova: Also, cold you please place fields autowired fields declaration?

ekazachkova (changes requested):

api/src/test/java/com/epam/pipeline/test/creator/cluster/NodeCreatorUtils.java
```diff
@@ -51,6 +53,24 @@ public static NodeInstance getDefaultNodeInstance() {
         return new NodeInstance();
     }
 
+    public static NodeInstance getNodeInstanceWithPermission() {
+        NodeInstance nodeInstance = new NodeInstance();
```
ekazachkova: Could you please add final to fields here and in other places?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+    @Autowired
+    private UsageMonitoringManager mockUsageMonitoringManager;
+
+    @Autowired
+    private NodeDiskManager mockNodeDiskManager;
+
+    @Autowired
+    private InstanceOfferManager mockInstanceOfferManager;
+
+    private List<NodeInstance> initSingleNodeInstanceList() {
```
ekazachkova: The `private` methods shell be after all `public` methods

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+    public void shouldReturnStatsForNodeForAdmin() {
+        final MonitoringStats monitoringStats = ContextualPreferenceCreatorUtils.getMonitoringStats();
+        final List<MonitoringStats> statsList = Collections.singletonList(monitoringStats);
+        doReturn(statsList).when(mockUsageMonitoringManager)
+                .getStatsForNode(nodeInstance.getName(), LocalDateTime.MIN, LocalDateTime.MAX);
+
+        final List<MonitoringStats> resultStatsList =
+                clusterApiService.getStatsForNode(nodeInstance.getName(), LocalDateTime.MIN, LocalDateTime.MAX);
+
+        assertThat(resultStatsList.size()).isEqualTo(1);
```
ekazachkova: Could we use here
```
assertThat(resultStatsList).hasSize(1).contains(monitoringStats);
```
And in other such places too.

api/src/test/java/com/epam/pipeline/test/acl/AclTestConfiguration.java
```diff
@@ -39,7 +39,10 @@
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @ContextConfiguration(classes = {AclTestBeans.class, AclSecurityConfiguration.class})
-@ComponentScan(basePackages = {"com.epam.pipeline.security.acl", "com.epam.pipeline.manager.security"})
+@ComponentScan(basePackages = {"com.epam.pipeline.security.acl",
```
ekazachkova: It seems that we can just use `basePackages = {"com.epam.pipeline.security.acl", "com.epam.pipeline.manager.security"}`

api/src/test/java/com/epam/pipeline/test/creator/cluster/NodeCreatorUtils.java
```diff
@@ -23,6 +23,7 @@
 import com.epam.pipeline.entity.cluster.MasterNode;
 import com.epam.pipeline.entity.cluster.NodeDisk;
 import com.epam.pipeline.entity.cluster.NodeInstance;
+import com.epam.pipeline.test.creator.pipeline.PipelineCreatorUtils;
```
ekazachkova, line 29: Could you please rename this class to `ClusterCreatorUtils`? This class should consist of creators for entities from `cluster` package.

api/src/test/java/com/epam/pipeline/test/creator/contextual/ContextualPreferenceCreatorUtils.java
```diff
@@ -55,4 +56,8 @@ public static ContextualPreferenceVO getContextualPreferenceVO() {
                 TEST_STRING, TEST_STRING, PREFERENCE_TYPE, getCPExternalResource()
         );
     }
+
+    public static MonitoringStats getMonitoringStats() {
```
ekazachkova: Could you move this method to `ClusterCreatorUtils`?

api/src/test/java/com/epam/pipeline/test/creator/pipeline/PipelineCreatorUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.epam.pipeline.entity.pipeline.PipelineRun;
+
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID_2;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+
+public final class PipelineCreatorUtils {
+
+    public static PipelineRun getPipelineRunWithPermission() {
```
ekazachkova: It is absolutely not clear what is meant `permission`

api/src/test/java/com/epam/pipeline/test/creator/pipeline/PipelineCreatorUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.epam.pipeline.test.creator.pipeline;
+
+import com.epam.pipeline.entity.pipeline.PipelineRun;
+
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID_2;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+
+public final class PipelineCreatorUtils {
```
ekazachkova, line 23: Let's avoid some test case specific naming in such classes. For example, for current situation we need only one method  for this class - `getPipelineRun` that will return `PipelineRun` object and accepts all required argument: id, name, owner... You may take the `ObjectCreatorUtils` class as a basis.

api/src/test/java/com/epam/pipeline/test/creator/cluster/NodeCreatorUtils.java
```diff
@@ -51,6 +53,24 @@ public static NodeInstance getDefaultNodeInstance() {
...
+    public static NodeInstance getNodeInstanceWithPermission() {
+        NodeInstance nodeInstance = new NodeInstance();
+        nodeInstance.setId(ID);
+        nodeInstance.setOwner("OWNER");
+        nodeInstance.setPipelineRun(PipelineCreatorUtils.getPipelineRunWithPermission());
+        nodeInstance.setName(TEST_STRING);
+        return nodeInstance;
+    }
+
+    public static NodeInstance getNodeInstanceWithoutPermission() {
```
ekazachkova: see `PipelineCreatorUtils` comment

api/src/test/java/com/epam/pipeline/test/creator/pipeline/PipelineCreatorUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static PipelineRun getPipelineRunWithPermission() {
+        PipelineRun pipelineRun = new PipelineRun();
```
ekazachkova: Do not forget to make variables `final`

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+        doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName());
+        doReturn(pipelineRun).when(mockPipelineRunManager).loadPipelineRun(eq(pipelineRun.getId()));
+
+        clusterApiService.terminateNode(nodeInstance.getName());
+    }
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldReturnStatsForNodeForAdmin() {
+        final MonitoringStats monitoringStats = ContextualPreferenceCreatorUtils.getMonitoringStats();
```
ekazachkova: It seems this initialization is redundant.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+
+        assertThat(nodes).isEmpty();
+    }
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldReturnNodeInstanceForAdmin() {
+        doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName());
+
+       final NodeInstance node = clusterApiService.getNode(nodeInstance.getName());
```
ekazachkova: Format code please

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+        twoNodeInstances.add(nodeInstanceWithoutPermission);
+        return twoNodeInstances;
+    }
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldReturnListWithNodeInstancesForAdmin() {
+        doReturn(initSingleNodeInstanceList()).when(mockNodesManager).getNodes();
+
+        final List<NodeInstance> nodes = clusterApiService.getNodes();
```
ekazachkova: It seems for such simple and short cases we can merge these two lines: 
```
assertThat(clusterApiService.getNodes()).hasSize(1).contains(nodeInstance);
```

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+
+        assertThat(nodes).hasSize(1).contains(nodeInstance);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldReturnListWithNodeInstancesWhenPermissionIsGranted() {
+        doReturn(SIMPLE_USER).when(mockAuthManager).getAuthorizedUser();
+        initAclEntity(nodeInstance,
+                Collections.singletonList(new UserPermission(SIMPLE_USER_ROLE, AclPermission.READ.getMask())));
```
ekazachkova: Maybe we shell consider creating a common for `*ApiServiceTest` classes method to build a single user permissions?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+        doReturn(contextualPreference).when(mockContextualPreferenceManager).search(
```
ekazachkova: Looks like this initialization is redundant

ekazachkova: Could you check it for other methods too?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
ekazachkova: Could you check it for other methods too?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,508 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doReturn;
+
+public class ClusterApiServiceTest extends AbstractAclTest {
```
ekazachkova, line 61: Maybe we shell add `@AclMask` tests? (for merging entity permissions with parent permissions)

P.S.: see `AclAspect`

ekazachkova (changes requested):

api/src/test/java/com/epam/pipeline/test/acl/AclTestBeans.java
```diff
@@ -17,47 +17,79 @@
 package com.epam.pipeline.test.acl;
```
tcibinan: Could you please check if all of the changes in this class are relevant to the pull request? It seems that mocks like `PipelineDao` have nothing to do with acl tests.

api/src/main/java/com/epam/pipeline/entity/cluster/NodeInstance.java
```diff
@@ -45,7 +45,6 @@
 public class NodeInstance extends AbstractSecuredEntity {
 
     private UUID uid;
-    private String name;
```
tcibinan: I'm afraid that this change can be harmful if some api method has `NodeInstance` class in request body. For example as part of `PipelineRun` class. Let's revert the change until we have implemented tests for all the controllers.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,451 @@
...
+import java.util.List;
+
+import static com.epam.pipeline.util.CustomAssertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doReturn;
+
+public class ClusterApiServiceTest extends AbstractAclTest {
+
+    private final String TEST_STRING = "TEST";
```
tcibinan: Could you please replace the field usages with `CommonCreatorConstants.TEST_STRING`.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,451 @@
...
+    private final FilterPodsRequest filterPodsRequest = NodeCreatorUtils.getDefaultFilterPodsRequest();
+    private final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, SIMPLE_USER);
+    private final NodeInstance nodeInstanceWithoutPermission = NodeCreatorUtils.getNodeInstance(2L, TEST_STRING);
```
tcibinan: I suppose we can call the second one just `anotherNodeInstance`.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,451 @@
...
+        assertThat(clusterApiService.getNodes()).hasSize(1).contains(nodeInstance);
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldReturnEmptyNodeInstanceListWhenPermissionIsNotGranted() {
+        initAclEntity(nodeInstanceWithoutPermission);
+        doReturn(initSingleNodeInstanceList()).when(mockNodesManager).getNodes();
+
+        assertThat(clusterApiService.getNodes()).isEmpty();
```
tcibinan: Method `initSingleNodeInstanceList()` returns list which doesn't contain `nodeInstanceWithoutPermission` so this test is not testing what it should. Probably we can even get rid of methods like `initSingleNodeInstanceList` and always initialiaze lists like the following
```java
Collections.singletonList(node);
// or
Arrays.asList(node1, node2);
```

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,451 @@
...
+
+        final NodeInstance returnedNodeInstance = clusterApiService.getNode(nodeInstance.getName());
+
+        assertThat(returnedNodeInstance).isEqualTo(nodeInstance);
+        assertThat(returnedNodeInstance.getMask()).isEqualTo(AclPermission.READ.getMask());
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldDenyAccessToNode() {
```
tcibinan: Let's specify explicitly why the access is denied like `shouldDenyAccessToNodeWhenPermissionIsNotGranted`.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,451 @@
...
+    @Test
+    @WithMockUser
+    public void shouldReturnStatsWhenPermissionIsGranted() {
+        doReturn(SIMPLE_USER).when(mockAuthManager).getAuthorizedUser();
+        initAclEntity(nodeInstance, AclPermission.READ);
+        doReturn(statsList).when(mockUsageMonitoringManager).getStatsForNode(
+                nodeInstance.getName(), LocalDateTime.MIN, LocalDateTime.MAX);
+        doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName(), filterPodsRequest);
+        doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName());
+        doReturn(pipelineRun).when(mockPipelineRunManager).loadPipelineRun(eq(pipelineRun.getId()));
```
tcibinan: I wonder if we can extract all this duplicated mocking calls into some methods like the following:
```java
mockUser();
mockStats();
mockNodeInstance();
mockPipelineRun();
```

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
youkofan: I already have four static imports in this class, and, if i'm not mistaken, PMD avoid amount of static imports more than four.

api/src/test/java/com/epam/pipeline/test/acl/AbstractAclTest.java
```diff
@@ -131,4 +132,10 @@ public Sid toSid() {
             return new GrantedAuthoritySid(authorityName);
         }
     }
+
+    protected List<Object> mutableListOf(Object... objects) {
```
tcibinan: Could you please use generic type to return `List<T>` rather then `List<Object>`. It probably is a good idea because it increases the possibility that this method will be reused.

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,432 @@
...
+
+        assertThat(clusterApiService.getNode(nodeInstance.getName(), filterPodsRequest)).isEqualTo(nodeInstance);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldReturnNodeThroughRequestWhenPermissionIsGranted() {
+        final NodeInstance nodeInstance = NodeCreatorUtils.getNodeInstance(1L, TEST_STRING);
+        initAclEntity(nodeInstance, AclPermission.READ);
+        initMocksBehavior(true);
```
tcibinan: Probably this mocking call makes thing even more complicated then before. For example in this case `nodeInstance` used for acl is not the same instance that will be used in `initMocksBehavior`. Also this `true\false` flag is not intuitive. Just imagine that there will be another boolean parameter.

I still think that it will be better to use separate calls for mocks which will just remove a need to write the whole
`doReturn().when().call()` chain. For example in this case we can use something like the following. It does not contain less lines then the original version. But it is not the point, isn't it? The following code is declarative as plain mockito calls but it is also much simpler. And in case f.e. pipeline run mocking mechanism will have to change in the future then we will just have to change a single method `mockRun`.
```java
mockUser(SIMPLE_USER); // Mock any user related calls that are used in the tests.
mockNode(nodeInstance); // Mock all node instance calls that are used in the tests.
mockRun(pipelineRun); // Mock all pipeline run calls that are used in the tests.
```

```java
public void mockNode(final NodeInstance nodeInstance) {
    doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName(), filterPodsRequest);
    doReturn(nodeInstance).when(mockNodesManager).getNode(nodeInstance.getName());
}
```

What do you think?

api/src/test/java/com/epam/pipeline/acl/cluster/ClusterApiServiceTest.java
```diff
@@ -0,0 +1,444 @@
...
+        assertThat(clusterApiService.loadNodeDisks(nodeDisk.getNodeId())).hasSize(1).contains(nodeDisk);
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldDenyAccessToNodeDisksWhenPermissionIsNotGranted() {
+        initAclEntity(nodeInstance);
+        doReturn(nodeDisks).when(mockNodeDiskManager).loadByNodeId(nodeDisk.getNodeId());
+        doReturn(nodeInstance).when(mockNodesManager).getNode(nodeDisk.getNodeId(), filterPodsRequest);
+        doReturn(nodeInstance).when(mockNodesManager).getNode(nodeDisk.getNodeId());
```
tcibinan: It does look like we can just replace it with the new method `mockNode`. Some similar cases with `terminateNode` calls in different tests too. You can just add `terminateNode` to `mockNode` and everything seems to be just fine.

tcibinan (approved):

Issue #1405, opened 2020-09-15
Implement tests for Controller layer


mzueva:
**Background**
Currently we don't have tests that validate `Controller` layer of our services, but we need to test it including Json (de)serialization and authentication/authorization.

**Approach**
Create isolated tests for Web layer using `@WebMvcTest` annotation:
- Provide test context configuration for Web layer
- Implement tests for the following packages:
     - billing
     - cluster
     - configuration
     - contextual
     - datastorage
     - dts
     - entity
     - issue
     - log
     - metadata
     - notification
     - pipeline
     - preference
     - region
     - search
     - security
     - user

mzueva:
@YouKofan 
Next classes for tests:
- `DockerRegistryController`
- `ToolController`
- `ToolGroupController`

These classes shall be moved to `controller/docker` package.

PR #1487, opened 2020-10-12
Issue 1405 Implemented tests for configuration controllers

cryteq:
This PR is related to issue #1405 and contains tests implementation for `configuration` package

### DIFF
--- a/api/src/test/java/com/epam/pipeline/controller/configuration/ConfigurationControllerTest.java
+++ b/api/src/test/java/com/epam/pipeline/controller/configuration/ConfigurationControllerTest.java
@@ -1,0 +1,216 @@
+/*
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.controller.configuration;
+
+import com.epam.pipeline.config.JsonMapper;
+import com.epam.pipeline.controller.ResponseResult;
+import com.epam.pipeline.controller.Result;
+import com.epam.pipeline.controller.vo.configuration.RunConfigurationVO;
+import com.epam.pipeline.entity.configuration.RunConfiguration;
+import com.epam.pipeline.manager.configuration.RunConfigurationApiService;
+import com.epam.pipeline.test.web.AbstractControllerTest;
+import com.epam.pipeline.util.ControllerTestUtils;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import org.assertj.core.api.Assertions;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.test.context.support.WithMockUser;
+import org.springframework.test.web.servlet.MvcResult;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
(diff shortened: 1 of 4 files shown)


api/src/test/java/com/epam/pipeline/controller/configuration/ServerlessConfigurationControllerTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    @WithMockUser
+    public void shouldRun() throws Exception {
+        Mockito.doReturn(RESULT).when(mockServerlessConfigurationApiService)
+                .run(eq(ID), eq(TEST_CONFIG), Mockito.any());
+
+        performRequest(get(String.format(RUN_URL, ID, TEST_CONFIG)));
+
+        Mockito.verify(mockServerlessConfigurationApiService).run(eq(ID), eq(TEST_CONFIG), Mockito.any());
+    }
```
tcibinan: Could you please check that POST, GET, PUT and DELETE requests can be made. For example parameterized test can be of help here.

api/src/test/java/com/epam/pipeline/controller/configuration/ConfigurationControllerTest.java
```diff
@@ -0,0 +1,138 @@
...
+        final MvcResult mvcResult = performRequest(post(CONFIGURATION_URL).content(content));
+
+        Mockito.verify(mockRunConfigurationApiService).save(Mockito.refEq(runConfigurationVO));
+        assertResponse(mvcResult, mapper, runConfiguration, ConfigurationCreatorUtils.RUN_CONFIGURATION_TYPE);
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldUpdateConfiguration() throws Exception {
+        runConfigurationVO.setId(ID);
```
tcibinan, line 72: You have to create an additional instance of run configuration. Otherwise the test class will fail once tests execution order is different which is possible since junit 4 doesn't have any guarantees on tests execution order.

api/src/test/java/com/epam/pipeline/controller/configuration/ConfigurationControllerTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private final RunConfiguration runConfiguration = ConfigurationCreatorUtils.getRunConfiguration();
+    private final RunConfigurationVO runConfigurationVO = ConfigurationCreatorUtils.getRunConfigurationVO();
+
+    @Autowired
+    private RunConfigurationApiService mockRunConfigurationApiService;
+
+    @Before
+    public void setUp() {
+        mapper = getObjectMapper();
+        mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
tcibinan: Unfortunately this doesn't work as you would expect. Basically `getObjectMapper()` doesn't return a new instance on each call.

To do this properly we have to use two instances of object mappers. Default one which is used for serialization / deserialization in spring cannot be used as is in tests. It can be used for serialization but not for deserealization. Therefore let's use two object mappers in the abstract class. Something like the following can be of help.

```java
final ObjectMapper serializationObjectMapper = currentObjectMapper;
final ObjectMapper deserializationObjectMapper = JsonMapper.newInstance();
deserializationObjectMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
...
final Result<T> actualResult = JsonMapper.parseData(actual, typeReference, deserializationObjectMapper);
```

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -81,13 +83,16 @@ protected final JsonMapper getObjectMapper() {
                                    final JsonMapper objectMapper,
                                    final T payload,
                                    final TypeReference<Result<T>> typeReference) throws Exception {
+        final ObjectMapper deserializationMapper = JsonMapper.newInstance();
+        deserializationMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
tcibinan: Let's move it to a class field.

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -81,13 +83,16 @@ protected final JsonMapper getObjectMapper() {
...
         final ResponseResult<T> expectedResult = buildExpectedResult(payload);
 
         final String actual = mvcResult.getResponse().getContentAsString();
         Assert.assertTrue(StringUtils.isNotBlank(actual));
         assertThat(actual).isEqualToIgnoringWhitespace(objectMapper.writeValueAsString(expectedResult));
 
-        final Result<T> actualResult = JsonMapper.parseData(actual, typeReference);
+        final Result<T> actualResult = JsonMapper.parseData(actual, typeReference, deserializationMapper);
+        assertNotNull(actualResult);
         Assert.assertEquals(expectedResult.getPayload(), actualResult.getPayload());
```
tcibinan: Please use either `Assert.assertSomething` or just `assertSomething` syntax, not both. It is pretty hard to read if the code is not consistent.

tcibinan (approved):

PR #1495, opened 2020-10-14
Issue #1404: Implemented tests for ContextualPreferenceApiService

youkofan:
Implemented tests for ContextualPreferenceApiService
Issue - #1404

### DIFF
--- a/api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
+++ b/api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
@@ -40,11 +40,12 @@ public class ContextualPreferenceApiServiceTest extends AbstractAclTest {
 
     private final ContextualPreference contextualPreference =
             ContextualPreferenceCreatorUtils.getContextualPreference();
-    private final List<ContextualPreference> preferenceList = Collections.singletonList(contextualPreference);
     private final ContextualPreferenceExternalResource externalResource =
             ContextualPreferenceCreatorUtils.getCPExternalResource();
     private final ContextualPreferenceVO contextualPreferenceVO =
             ContextualPreferenceCreatorUtils.getContextualPreferenceVO();
+
+    private final List<ContextualPreference> preferenceList = Collections.singletonList(contextualPreference);
 
     @Autowired
     private ContextualPreferenceApiService preferenceApiService;
@@ -57,10 +58,7 @@ public class ContextualPreferenceApiServiceTest extends AbstractAclTest {
     public void shouldLoadAllContextualPreferencesForAdmin() {
         doReturn(preferenceList).when(mockPreferenceManager).loadAll();
 
-        List<ContextualPreference> loadedPreferences = preferenceApiService.loadAll();
-
-        assertThat(loadedPreferences.size()).isEqualTo(1);
-        assertThat(loadedPreferences.get(0)).isEqualTo(contextualPreference);
+        assertThat(preferenceApiService.loadAll()).hasSize(1).contains(contextualPreference);
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+        assertThat(resultPreferenceList.size()).isEqualTo(1);
+        assertThat(resultPreferenceList.get(0)).isEqualTo(contextualPreference);
+    }
+
+    @Test(expected = AccessDeniedException.class)
+    @WithMockUser(roles = SIMPLE_USER_ROLE)
+    public void shouldDenyLoadingAllContextualPreferencesForNotAdmin() {
+        doReturn(preferenceList).when(mockPreferenceManager).loadAll();
+
+        preferenceApiService.loadAll();
```
tcibinan: There is a helpful method in `CustomAssertions` utility class. Probably it is better to use such method to assert exceptions rather than using `@Test(expected = AccessDeniedException.class)` cause it allows us to specify only a single line that should fail.

```java
assertThrows(AccessDeniedException.class, () -> preferenceApiService.loadAll())
```

api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Autowired
+    private ContextualPreferenceManager mockPreferenceManager;
+
+    private final ContextualPreference contextualPreference =
+            ContextualPreferenceCreatorUtils.getContextualPreference();
+
+    private final List<ContextualPreference> preferenceList = Collections.singletonList(contextualPreference);
+
+    private final ContextualPreferenceExternalResource cpeResource =
```
tcibinan: It seems fine to just call it `externalResource` or even `resource` because we are already in the context of the contextual preference test class.

api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private final ContextualPreferenceVO contextualPreferenceVO =
+            ContextualPreferenceCreatorUtils.getContextualPreferenceVO();
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldLoadAllContextualPreferencesForAdmin() {
+        doReturn(preferenceList).when(mockPreferenceManager).loadAll();
+
+        List<ContextualPreference> resultPreferenceList = preferenceApiService.loadAll();
```
tcibinan: What about calling this loaded preferences list just `loadedPreferences`. Such `result` prefix is just a boilerplate giving no additional meaning to the reader :)

api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @Test(expected = AccessDeniedException.class)
+    @WithMockUser(roles = SIMPLE_USER_ROLE)
+    public void shouldDenyLoadingContextualPreferenceForNotAdmin() {
+        doReturn(contextualPreference).when(mockPreferenceManager).load(TEST_STRING, cpeResource);
+
+        preferenceApiService.load(TEST_STRING, cpeResource);
+    }
+
+    @Test
+    public void shouldSearchContextualPreference() {
```
tcibinan, line 90: We expect here to be at least a simple user role here so let's add corresponding `@WithMockUser` annotation.

api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
tcibinan: In case where there is only one assert in the test method we can even inline the variable and don't think about naming it at all.

api/src/test/java/com/epam/pipeline/acl/contextual/ContextualPreferenceApiServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldLoadAllContextualPreferencesForAdmin() {
+        doReturn(preferenceList).when(mockPreferenceManager).loadAll();
+
+        assertThat(preferenceApiService.loadAll()).hasSize(1).contains(contextualPreference);
+    }
+
+    @Test
+    @WithMockUser(roles = SIMPLE_USER_ROLE)
```
tcibinan: As discussed with the team let's use `@WithMockUser` wherever possible. Even while testing methods without acl annotations at all. Specified without arguments it mocks the test with `user` name and `role_user` authority.

tcibinan (approved):

mzueva (changes requested):
Please, fix checkstyle for ContextualPreferenceApiServiceTest class

PR #1518, opened 2020-10-27
Support custom node images

### LINKED ISSUE #1517: Support custom node images for runs
**Background**

Currently Cloud Pipeline allows to run instances using two default node images. One node image for all CPU instance types and other for GPU ones.

Nevertheless there are cases than some of the tools or pipelines require special node images. For example, some tool may require specific nvidia driver version which default GPU node image doesn't have.

**Approach**

To resolve all possible node image incompatibility issues support for custom node images has to be introduced. If custom node image is specified for pipeline, tool or just a single run then cloud instance with the required node image has to be either instantiated or reassigned.

tcibinan:
Resolves #1517.

The pull request brings support for custom node images on the backend side. From now on `instance_image` settings field will be used to override default node image from cluster configuration. Also pipeline start `instanceImage` entity field can be used to specify custom node image for a single launch.

For example the following configuration entry enforces custom node image for cloud instances.

```json
{
  "name": "default",
  "configuration": {
    "nonPause": true,
    "instance_size": "instance_size",
    "instance_image": "custom node image",
    "instance_disk": "instance_disk",
    "cmd_template": "cmd_template",
    "language": "language",
    "is_spot": true
  },
  "default": true
}
```

And the following pipeline start entity enforces custom node image for a single launch cloud instance.

```json
{
  "instanceType": "instanceType",
  "instanceImage": "custom node image",
  "hddSize": 1,
  "timeout": 1,
  "cmdTemplate": "cmdTemplate",
  "dockerImage": "dockerImage",
  "params": {},
  "isSpot": true,
  "cloudRegionId": 1,
  "nonPause": true,
  "force": true
}
```


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineConfigurationManager.java
```diff
@@ -173,14 +173,14 @@ public PipelineConfiguration mergeParameters(PipelineStart runVO, PipelineConfig
             configuration.setInstanceDisk(defaultConfig.getInstanceDisk());
         }
         if (runVO.getInstanceType() != null) {
-            configuration.setInstanceType(String.valueOf(runVO.getInstanceType()));
+            configuration.setInstanceType(runVO.getInstanceType());
         } else {
             configuration.setInstanceType(defaultConfig.getInstanceType());
         }
-        if (runVO.getTimeout() != null) {
```
mzueva: I think we need to keep 'timeout' field as well

tcibinan: Oops. Fixed.

mzueva (approved):

Issue #1520, opened 2020-10-27
Jobs, that were stopped by a "Timeout", shall move the interim data into the "output" storage


sidoruka:
**Background**
At the moment, if a `timeout` is elapsed - the job is stopped and all the data is erased. 
![image](https://user-images.githubusercontent.com/20998291/97339797-131fa700-1894-11eb-8070-88c621869f26.png)

Some of the users are complaining about it, as the intermediate data is lost.
We need a solution to extract the current data from the timed-out jobs.

**Approach**
* If a job has timed out - it shall not be stopped immediately
* Instead, the `Output Data` task shall be triggered
* All the contents of the `$ANALYSIS_DIR` shall be copied to the `output` storages. In the same manner, as if the job has succeeded

NShaforostov:
Docs were added via #1972 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/release_notes/v.0.17/v.0.17_-_Release_notes.md#saving-of-interim-data-for-jobs-stopped-by-a-timeout).

PR #1522, opened 2020-10-28
Jobs, that were stopped by a "Timeout", shall move the interim data into the "output" storage

ekazachkova:
This PR provides implementation for issue #1520 

- a new system parameter `CP_EXEC_TIMEOUT` was introduced
- if both `Timeout` field and `CP_EXEC_TIMEOUT` system parameter were specified the `CP_EXEC_TIMEOUT` will be used
- if timeout was specified the `launch.sh` script will continue its flow after the timeout elapsed

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/execution/SystemParams.java
+++ b/api/src/main/java/com/epam/pipeline/manager/execution/SystemParams.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 EPAM Systems, Inc. (https://www.epam.com/)
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


api/src/main/java/com/epam/pipeline/manager/execution/SystemParams.java
```diff
@@ -69,7 +69,8 @@
     CONTAINER_CPU_RESOURCE("container-cpu-resource", "CP_CONTAINER_CPU_RESOURCE", false, true),
     CONTAINER_MEMORY_RESOURCE_POLICY("container-memory-resource-policy",
             "CP_CONTAINER_MEMORY_RESOURCE_POLICY", false, true),
-    CP_SENSITIVE_RUN("cp-sensitive-run", "CP_SENSITIVE_RUN", false, false);
+    CP_SENSITIVE_RUN("cp-sensitive-run", "CP_SENSITIVE_RUN", false, false),
+    CP_EXEC_TIMEOUT("cp-exec-timeout", "CP_EXEC_TIMEOUT");
```
mzueva, line 73: Could you also add `CP_EXEC_TIMEOUT` to `deploy/contents/install/preferences/launch.system.parameters.json`?

scripts/pipeline-launch/launch.sh
```diff
@@ -1678,7 +1678,15 @@ pipe_log SUCCESS "Environment initialization finished" "InitializeEnvironment"
 
 echo "Command text:"
 echo "${SCRIPT}"
-bash -c "${SCRIPT}"
+
+if [ ! -z "${CP_EXEC_TIMEOUT}" ] && [ "${CP_EXEC_TIMEOUT}" -gt 0 ];
+then
+  timeout ${CP_EXEC_TIMEOUT}m bash -c "${SCRIPT}"
+  echo "Timeout was elapsed"
```
mzueva: I think that we need to check whether timeout was actually exceeded (possibly using exit code). Also keep in memory that we need to save `CP_EXEC_RESULT` to exit process with correct exit code.

mzueva (changes requested):

mzueva (approved):

Issue #1526, opened 2020-10-29
Billing service enhancements


mzueva:
**Background**
In order to simplify deployment and initial billing data generation process I'd suggest to implement some minor enhancements for the billing service:
- allow to configure billing service deployment via envvars
- allow to optionally enable historical storage billing generation. E.g. if some property is set storage billing is generated not for the current date, but from the specified date
- allow to pass last synchronization time from env vars, it shall have lower precedence than `lastSynchonization` file and shall be applied only if it is not present.

PR #1530, opened 2020-10-29
Issue #1526 Billing improvements

Wedds:
This PR is related to issue #1526

It brings some improvements to billing service

1. Billing properties could be configured via environment variable
2. Billing initial date could be specified using `sync.billing.start.date` - if no last sync time is specified this value will be used
3. Modify AWS price-list loading: skip price rates without USD cost specified to avoid NPE


deploy/docker/cp-billing-srv/config/application.properties
```diff
@@ -31,6 +31,7 @@ sync.submit.threads=1
 sync.billing.schedule=0 0 0 ? * *
```
mzueva: Could you also add env var to configure schedule?

billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/service/impl/converter/StorageToBillingRequestConverter.java
```diff
@@ -97,10 +102,13 @@ public StorageToBillingRequestConverter(final AbstractEntityMapper<StorageBillin
                                                          final LocalDateTime syncStart) {
         final Long storageId = storageContainer.getEntity().getId();
         final DataStorageType storageType = storageContainer.getEntity().getType();
-        final LocalDate reportDate = syncStart.toLocalDate().minusDays(1);
-        final String fullIndex = indexPrefix + parseDateToString(reportDate);
         return requestSumAggregationForStorage(storageId, storageType)
-            .map(searchResponse -> buildRequestFromAggregation(storageContainer, syncStart, searchResponse, fullIndex))
+            .map(searchResponse -> Stream.iterate(previousSync.plusDays(1L), date -> date.plusDays(1))
```
mzueva: Could you please make this behaviour configurable, by default we create documents only for current date, and optionally for the period

mzueva (changes requested):

Issue #1501, opened 2020-10-19
Implement run ports forwarding in pipe cli


tcibinan:
**Background**

It can be quite useful to have support for run ports forwarding in pipe cli. For example run instance ssh port can be forwarded to some local port to be able to ssh into it using regular ssh client.

Such tunneling should forward traffic through edge service similarly to how nomachine traffic is managed now.

**Approach**

Additional `pipe tunnel` command should be added. Once called it will create and serve tcp tunnel in background which forwards local port traffic to some run instance port. The implementation should be based on `pipe ssh` one which already uses ssh ports tunneling. Moreover `pipe tunnel` command should allow to stop tunnels as well as create them.

SSH ports forwarding support for `pipe tunnel` should be extended as it seems to be a common use case for pipe tunnel. Additional command flag should be added which once specified enables automated passwordless ssh configuration for the tunnel.

tcibinan:
Cherry-picked to `release/0.16`.

NShaforostov:
**Docs** were added via #1516.

NShaforostov:
Verified. Tested functionality works well.
Were tested base scenarios:

- creation of the tunnel connection from instance `ssh` port to some local port
   - plus different logging options
   - plus different regimens (background/foreground)
- connection to the run instance using `ssh` client
- copying files up to 5Gb to/from the run instance using `scp` client
- tunnel stop operation
   - plus different logging options

Testing was performed in WSL (`Ubuntu 18.04 LTS`) and `Windows PowerShell`

PR #1534, opened 2020-11-02
Implement passwordless ssh support in pipe tunnel

tcibinan:
Relates to #1501 and depends on #1502.

The pull request introduces additional `-s | --ssh` option for `pipe tunnel` which once specified configures passwordless ssh connections to run instances.

See `pipe tunnel --help` output for more information:
```
  Establishes tunnel connection to specified run instance port and serves it
  as a local port.

  It allows to transfer any tcp traffic from local machine to run instance
  and works both on Linux and Windows.

  Additionally it enables passwordless ssh connections if the corresponding
  option is specified. Once specified ssh is configured both locally and
  remotely to support passwordless connections. Passwordless ssh
  configuration is supported only for openssh client on Linux.

  Examples:

  I. Example of simple tcp port tunnel connection establishing.

  Establish tunnel connection from run (12345) instance port (4567) to the
  same local port.

      pipe tunnel -lp 4567 -rp 4567 12345

  II. Example of ssh port tunnel connection establishing with enabled
  passwordless ssh configuration.

  First of all establish tunnel connection from run (12345) instance ssh
  port (22) to some local port (4567).

      pipe tunnel -lp 4567 -rp 22 --ssh 12345

  Then connect to run instance using regular ssh client.

      ssh root@pipeline-12345

  Or transfer some files to and from run instance using regular scp client.

      scp file.txt root@pipeline-12345:/common/workdir/file.txt

  Additionally the following environment variables can be used to specify
  the exact tunnel properties.

      CP_CLI_TUNNEL_PROXY_HOST
      CP_CLI_TUNNEL_PROXY_PORT
      CP_CLI_TUNNEL_TARGET_HOST
```

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -1273,12 +1273,13 @@ def tunnel(run_id, local_port, remote_port, ssh, log_file, log_level, timeout, f
 
         scp file.txt root@pipeline-12345:/common/workdir/file.txt
 
-    Additionally the following environment variables can be used to specify the exact tunnel properties.
+    Advanced tunnel configuration environment variables:
 
     \b
-        CP_CLI_TUNNEL_PROXY_HOST
-        CP_CLI_TUNNEL_PROXY_PORT
-        CP_CLI_TUNNEL_TARGET_HOST
+        CP_CLI_TUNNEL_PROXY_HOST - tunnel proxy host
+        CP_CLI_TUNNEL_PROXY_PORT - tunnel proxy port
+        CP_CLI_TUNNEL_TARGET_HOST - tunnel target host
+        CP_CLI_TUNNEL_SSH_PATH - .ssh directory path
     """
     if trace:
         create_tunnel(run_id, local_port, remote_port, ssh, log_file, log_level, timeout, foreground, retries)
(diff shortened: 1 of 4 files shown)


mzueva (approved):

NShaforostov:
Verified. Tested functionality works well.
Were tested base scenarios:
- creation of the tunnel connection from instance `ssh` port to some local port (plus logging options)
- connection to the run instance using `ssh` client
- copying files up to 5Gb to/from the run instance using `scp` client

Testing was performed in WSL (Ubuntu 18.04 LTS).

Some suggestions/notes:
1. Could we implement an additional (sub)command to close `ssh` tunnel? In the current implementation, it has be performed via `kill` command. Suspect that it is not really graceful (and could be not intuitive for users).
2. Could we print after the `pipe tunnel` command execution a short message that the tunnel creation was successful/unsuccessful? Or just short error message only in unsuccessful case (like it's done for the `pipe mount` command)? It would be more understandable for users.

mzueva (approved):

pipe-cli/pipe.py
```diff
@@ -1226,42 +1227,82 @@ def ssh(ctx, run_id, retries):
...
+        CP_CLI_TUNNEL_PROXY_PORT - tunnel proxy port
+        CP_CLI_TUNNEL_TARGET_HOST - tunnel target host
+        CP_CLI_TUNNEL_SERVER_ADDRESS - tunnel server address
     """
-    try:
-        create_tunnel(run_id, local_port, remote_port, log_file, log_level, timeout, foreground)
-    except Exception as runtime_error:
-        click.echo('Error: {}'.format(str(runtime_error)), err=True)
-        sys.exit(1)
+    if trace:
```
mzueva: If the goal is to print stack trace we can use something like `traceback.print_exc()` in `except` section

Issue #1542, opened 2020-11-05
Contextual properties shall be respected if set for multiple groups


sidoruka:
At the moment, if the user is a member of two (or more) groups and those groups have `jobs.visibility` set for - this setting is ignored.
While this makes sense if the groups have different values set - but if they equal, it shall not be ignored.

So we need to:
* Check if the contextual property's value is the same for multiple objects at the same level (e.g. "groups")
* If so - use any object's value

sidoruka:
Backport to `release/0.16`

PR #1547, opened 2020-11-06
Issue #1542: Support several contextual preferences at the same level

mzueva:
This PR provides implementation for #1542. 

It allows to resolve contextual preferences if they are specified at the same level with the same value.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/contextual/handler/DefaultContextualPreferenceReducer.java
+++ b/api/src/main/java/com/epam/pipeline/manager/contextual/handler/DefaultContextualPreferenceReducer.java
@@ -22,6 +22,7 @@ import com.epam.pipeline.entity.contextual.ContextualPreference;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/contextual/handler/DefaultContextualPreferenceReducer.java
```diff
@@ -57,15 +58,21 @@
...
     private Optional<ContextualPreference> reduceValidPreferences(final List<ContextualPreference> preferences) {
-        return preferenceReducerMap.getOrDefault(preferences.get(0).getName(), fallbackEmptyReducer())
+        return preferenceReducerMap.getOrDefault(preferences.get(0).getName(), defaultReducer())
                 .reduce(preferences);
     }
 
-    private ContextualPreferenceReducer fallbackEmptyReducer() {
+    private ContextualPreferenceReducer defaultReducer() {
         return preferences -> {
-            log.warn(messageHelper.getMessage(MessageConstants.WARN_CONTEXTUAL_PREFERENCE_REDUCER_NOT_FOUND,
```
tcibinan, line 66: We can get rid of `MessageConstants.WARN_CONTEXTUAL_PREFERENCE_REDUCER_NOT_FOUND` now.

tcibinan (approved):

api/src/main/java/com/epam/pipeline/manager/contextual/handler/DefaultContextualPreferenceReducer.java
mzueva: Thank you, fixed

PR #1583, opened 2020-11-19
Issue #1405: Implemented tests for Metadata package controller layer

youkofan:
Issue #1405

### DIFF
--- a/api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
+++ b/api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
@@ -17,20 +17,26 @@
 package com.epam.pipeline.entity.metadata;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 @Getter
 @Setter
-@NoArgsConstructor
-@AllArgsConstructor
 public class MetadataField {
     private String name;
     @JsonIgnore
     private String dbName;
     private boolean predefined = false;
+
+    public MetadataField(String name, String dbName, boolean predefined) {
+        this.name = name;
+        this.dbName = dbName;
+        this.predefined = predefined;
+    }
+
+    public MetadataField() {
+
+    }
 
     @Override public boolean equals(Object o) {
         if (this == o) {
--- a/api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
+++ b/api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
@@ -30,6 +30,7 @@ public class MetadataField {
     private String dbName;
     private boolean predefined = false;
 
+    @SuppressWarnings("linelength")
     //todo: Replace with @AllArgsConstructor once lombok version 1.16.20 is used.
     // The current version causes JsonMappingException.
     // See more https://stackoverflow.com/questions/40546508/jsoncreator-could-not-find-creator-property-with-name-even-with-ignoreunknown
--- a/api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
+++ b/api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
@@ -18,24 +18,25 @@ package com.epam.pipeline.entity.metadata;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 @Getter
 @Setter
+@NoArgsConstructor
 public class MetadataField {
     private String name;
     @JsonIgnore
     private String dbName;
     private boolean predefined = false;
 
+    //todo: Replace with @AllArgsConstructor once lombok version 1.16.20 is used.
+    // The current version causes JsonMappingException.
+    // See more https://stackoverflow.com/questions/40546508/jsoncreator-could-not-find-creator-property-with-name-even-with-ignoreunknown
     public MetadataField(String name, String dbName, boolean predefined) {
         this.name = name;
         this.dbName = dbName;
         this.predefined = predefined;
-    }
-
-    public MetadataField() {
-
     }
 
     @Override public boolean equals(Object o) {
--- a/api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
+++ b/api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
@@ -252,7 +252,8 @@ public class MetadataControllerTest extends AbstractControllerTest {
 
         verify(mockMetadataApiService).uploadMetadataFromFile(any(EntityVO.class),
                 multipartFileCaptor.capture(), eq(true));
-        assertRequestFile(multipartFileCaptor.getValue(), "file.txt", "content of file.txt".getBytes());
+        assertRequestFile(multipartFileCaptor.getValue(), MULTIPART_CONTENT_FILE_NAME,
+                                                          MULTIPART_CONTENT_FILE_CONTENT);
         assertResponse(mvcResult, metadataEntry, MetadataCreatorUtils.METADATA_ENTRY_TYPE);
     }
 
--- a/api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
+++ b/api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
@@ -29,7 +29,6 @@ import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
-import org.springframework.http.MediaType;
 import org.springframework.security.test.context.support.WithMockUser;
 import org.springframework.test.web.servlet.MvcResult;
 import org.springframework.web.multipart.MultipartFile;
@@ -40,7 +39,6 @@ import java.util.List;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING_SET;
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doReturn;
@@ -254,10 +252,7 @@ public class MetadataControllerTest extends AbstractControllerTest {
 
         verify(mockMetadataApiService).uploadMetadataFromFile(any(EntityVO.class),
                 multipartFileCaptor.capture(), eq(true));
-        final MultipartFile returnedFile = multipartFileCaptor.getValue();
-        assertThat(returnedFile.getOriginalFilename()).isEqualTo("file.txt");
-        assertThat(returnedFile.getBytes()).isEqualTo("file.txt".getBytes());
-        assertRequestFile(multipartFileCaptor, "file", MediaType.APPLICATION_OCTET_STREAM_VALUE);
+        assertRequestFile(multipartFileCaptor.getValue(), "file.txt", "file.txt".getBytes());
         assertResponse(mvcResult, metadataEntry, MetadataCreatorUtils.METADATA_ENTRY_TYPE);
     }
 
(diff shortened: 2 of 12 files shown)


api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
```diff
@@ -17,21 +17,27 @@
...
 @Setter
-@NoArgsConstructor
-@AllArgsConstructor
 public class MetadataField {
     private String name;
     @JsonIgnore
     private String dbName;
     private boolean predefined = false;
 
+    public MetadataField(String name, String dbName, boolean predefined) {
```
tcibinan, line 37: Do you understand why this change resolves jackson deserialization issue? Looks like some lombok \ jackson incompatibility. Probably we should add comment clarifying why we cannot use `@AllArgsConstructor` here. Additionally there seems to be no problems with `NoArgsConstructor` so let's use it as before.

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
```diff
@@ -0,0 +1,296 @@
...
+    @WithMockUser
+    public void shouldUploadMetadataFromFile() {
+        doReturn(metadataEntry).when(mockMetadataApiService)
+                .uploadMetadataFromFile(eq(new EntityVO(ID, aclClass)), any(), eq(true));
+
+        final MvcResult mvcResult = performRequest(post(UPLOAD_URL).content(MULTIPART_CONTENT)
+                        .params(multiValueMapOf(ENTITY_ID, ID,
+                                                ACL_CLASS, aclClass,
+                                                MERGE_WITH_EXISTING_METADATA, true)),
+                MULTIPART_CONTENT_TYPE, EXPECTED_CONTENT_TYPE);
```
tcibinan: Please add the proper indent

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
```diff
@@ -0,0 +1,296 @@
...
+
+        verify(mockMetadataApiService).uploadMetadataFromFile(any(EntityVO.class), any(), eq(true));
```
tcibinan: Let's check the contents of the second argument using `ArgumentCaptor`. This is the only way to be sure that the multipart file was parsed correctly.

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataEntityControllerTest.java
```diff
@@ -0,0 +1,431 @@
...
+    @Test
+    @WithMockUser
+    public void shouldUploadMetadataFromFile() {
+        final List<MetadataEntity> entities = Collections.singletonList(metadataEntity);
+        doReturn(entities).when(mockMetadataEntityApiService).uploadMetadataFromFile(eq(ID), any(MultipartFile.class));
+
+        final MvcResult mvcResult = performRequest(post(METADATA_ENTITY_UPLOAD_URL).content(MULTIPART_CONTENT)
+                .params(multiValueMapOf(PARENT_ID, ID)), MULTIPART_CONTENT_TYPE, EXPECTED_CONTENT_TYPE);
+
+        verify(mockMetadataEntityApiService).uploadMetadataFromFile(eq(ID), any(MultipartFile.class));
```
tcibinan: It would be helpful to use `ArgumentCaptor` here as well.

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataEntityControllerTest.java
```diff
@@ -0,0 +1,431 @@
...
+
+    @Test
+    public void shouldFailDeleteMetadataEntity() {
+        performUnauthorizedRequest(delete(String.format(METADATA_ENTITY_DELETE_URL, ID)));
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldUpdateMetadataEntityWithId() throws Exception {
+        metadataEntityVO.setEntityId(ID);
```
tcibinan, line 303: We need to create a new instance of the class here because there are no guaranties on junit tests execution order.

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataEntityControllerTest.java
```diff
@@ -0,0 +1,431 @@
...
+        doReturn(inputStream).when(mockMetadataEntityApiService).getMetadataEntityFile(ID, TEST_STRING, TEST_STRING);
+
+        final MvcResult mvcResult = performRequest(get(METADATA_ENTITY_DOWNLOAD_URL)
+                        .params(multiValueMapOf(FOLDER_ID, ID,
+                                                ENTITY_CLASS, TEST_STRING,
+                                                FILE_FORMAT, TEST_STRING)),
+                MediaType.APPLICATION_OCTET_STREAM_VALUE);
+
+        verify(mockMetadataEntityApiService).getMetadataEntityFile(ID, TEST_STRING, TEST_STRING);
+        Assert.assertEquals(TEST_STRING, mvcResult.getResponse().getContentAsString());
```
tcibinan: Can we use `assertFileResponse` method here?

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
```diff
@@ -0,0 +1,305 @@
...
+                                                ACL_CLASS, aclClass,
+                                                MERGE_WITH_EXISTING_METADATA, true)),
+                                                MULTIPART_CONTENT_TYPE, EXPECTED_CONTENT_TYPE);
+        final ArgumentCaptor<MultipartFile> multipartFileCaptor = ArgumentCaptor.forClass(MultipartFile.class);
+
+        verify(mockMetadataApiService).uploadMetadataFromFile(any(EntityVO.class),
+                multipartFileCaptor.capture(), eq(true));
+        final MultipartFile returnedFile = multipartFileCaptor.getValue();
+        assertThat(returnedFile.getContentType()).isEqualTo(MediaType.APPLICATION_OCTET_STREAM_VALUE);
+        assertThat(returnedFile.getOriginalFilename()).isEqualTo("file.txt");
```
tcibinan: I'm thinking that we should check `returnedFile.getOriginalFilename()` and `file.getBytes()` because these methods used in `MetadataApiService`.

Additionally it seems like a good idea to extract new assertion method which will accept the captor and check for the uploading file name and its content:

```java
assertRequestFile(captor, expectedFileName, expectedContent);
```

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
```diff
@@ -255,8 +255,9 @@ public void shouldUploadMetadataFromFile() throws Exception {
         verify(mockMetadataApiService).uploadMetadataFromFile(any(EntityVO.class),
                 multipartFileCaptor.capture(), eq(true));
         final MultipartFile returnedFile = multipartFileCaptor.getValue();
-        assertThat(returnedFile.getContentType()).isEqualTo(MediaType.APPLICATION_OCTET_STREAM_VALUE);
         assertThat(returnedFile.getOriginalFilename()).isEqualTo("file.txt");
+        assertThat(returnedFile.getBytes()).isEqualTo("file.txt".getBytes());
+        assertRequestFile(multipartFileCaptor, "file", MediaType.APPLICATION_OCTET_STREAM_VALUE);
```
tcibinan: It looks like all all these asserts can be replaced by a single call which accepts `MultipartFile` rather than captor.

```java
assertRequestFile(captor.getValue(), fileName, fileContentInBytesOrString);
``` 

Content type will probably be the same in all the similar cases. Therefore we don't have to pass it as a parameter and just use a default value `MediaType.APPLICATION_OCTET_STREAM_VALUE`.

Also I believe there was a similar case in this or other test class in the pull request. So it would be nice to use the same  `assertRequestFile` in both cases.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -129,12 +129,13 @@ public void assertContent(final MvcResult mvcResult, final byte[] fileContent) {
...
     }
 
-    public void assertRequestFile(final ArgumentCaptor<MultipartFile> captor,
-                                  final String expectedFileName,
-                                  final String expectedContent) {
-        final MultipartFile capturedValue = captor.getValue();
-        assertThat(capturedValue.getName()).isEqualTo(expectedFileName);
-        assertThat(capturedValue.getContentType()).isEqualTo(expectedContent);
+    @SneakyThrows
+    public <T> void assertRequestFile(final MultipartFile capturedValue,
```
tcibinan: Probably there is no need for generic type here.

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -59,12 +59,12 @@
...
-                    "Content-Disposition: form-data; name=\"file\"; filename=\"file.txt\"\r\n" +
-                    "Content-Type:  application/octet-stream\r\n" +
-                    "\r\n" +
-                    "file.txt" +
-                    "\r\n" +
-                    "----------------------------boundary";
+            "Content-Disposition: form-data; name=\"file\"; filename=\"file.txt\"\r\n" +
+            "Content-Type:  application/octet-stream\r\n" +
+            "\r\n" +
+            "file.txt" +
```
tcibinan: I'm wondering if we can extract `MULTIPART_CONTENT_FILE_CONTENT`. `MULTIPART_CONTENT_FILE_NAME` and `MULTIPART_CONTENT_FILE_CONTENT` fields. Additionally let's replace the content of the file from `file.txt` to some more distinct string like `content of file.txt`. Once changed please check that no tests depending on `MULTIPART_CONTENT` are failing.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/controller/metadata/MetadataControllerTest.java
```diff
@@ -252,7 +252,7 @@ public void shouldUploadMetadataFromFile() throws Exception {
 
         verify(mockMetadataApiService).uploadMetadataFromFile(any(EntityVO.class),
                 multipartFileCaptor.capture(), eq(true));
-        assertRequestFile(multipartFileCaptor.getValue(), "file.txt", "file.txt".getBytes());
+        assertRequestFile(multipartFileCaptor.getValue());
```
tcibinan: It doesn't seem like a good idea to hardcode a file name and its content. Let's pass these parameters explicitly.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -55,11 +55,13 @@
...
+    protected static final byte[] MULTIPART_CONTENT_FILE_CONTENT = "content of file.txt".getBytes();
     protected static final String MULTIPART_CONTENT_TYPE =
             "multipart/form-data; boundary=--------------------------boundary";
     protected static final String MULTIPART_CONTENT =
             "----------------------------boundary\r\n" +
-            "Content-Disposition: form-data; name=\"file\"; filename=\" file.txt \"\r\n" +
+            "Content-Disposition: form-data; name=\"file\"; filename=\" " + MULTIPART_CONTENT_FILE_NAME + " \"\r\n" +
             "Content-Type:  application/octet-stream\r\n" +
             "\r\n" +
             "content of file.txt" +
```
tcibinan: Please replace the string with `MULTIPART_CONTENT_FILE_CONTENT`.

tcibinan (changes requested):

api/src/main/java/com/epam/pipeline/entity/metadata/MetadataField.java
```diff
@@ -17,21 +17,28 @@
...
-@AllArgsConstructor
 public class MetadataField {
     private String name;
     @JsonIgnore
     private String dbName;
     private boolean predefined = false;
 
+    //todo: Replace with @AllArgsConstructor once lombok version 1.16.20 is used.
+    // The current version causes JsonMappingException.
+    // See more https://stackoverflow.com/questions/40546508/jsoncreator-could-not-find-creator-property-with-name-even-with-ignoreunknown
```
tcibinan, line 36: Could you please suppress checkstyle warning for this line using `SuppressWarnings` annotation?

tcibinan (changes requested):

tcibinan (approved):

PR #1588, opened 2020-11-20
Issue #1404: Implemented tests for EntityApiService

youkofan:
Issue #1404

### DIFF
--- a/api/src/test/java/com/epam/pipeline/acl/entity/EntityApiServiceTest.java
+++ b/api/src/test/java/com/epam/pipeline/acl/entity/EntityApiServiceTest.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.acl.entity;
+
+import com.epam.pipeline.entity.AbstractSecuredEntity;
+import com.epam.pipeline.entity.datastorage.aws.S3bucketDataStorage;
+import com.epam.pipeline.entity.security.acl.AclClass;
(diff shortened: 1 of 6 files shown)


api/src/test/java/com/epam/pipeline/acl/entity/EntityApiServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+    public void shouldLoadEntityForAdmin() {
+        doReturn(s3bucket).when(mockEntityManager).loadByNameOrId(AclClass.DATA_STORAGE, TEST_STRING);
+
+        assertThat(entityApiService.loadByNameOrId(AclClass.DATA_STORAGE, TEST_STRING)).isEqualTo(s3bucket);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldLoadEntityWhenPermissionIsGranted() {
+        s3bucket.setAclClass(AclClass.DATA_STORAGE);
```
tcibinan: It seems that `AbstractDataStorage` already has `aclClass` fields set to `AclClass.DATA_STORAGE`. Do we really need that to be specified explicitly?

api/src/test/java/com/epam/pipeline/app/AclTestConfiguration.java
```diff
@@ -214,6 +215,9 @@
     @MockBean
     protected PipelineRunDockerOperationManager pipelineRunDockerOperationManager;
 
+    @MockBean
```
tcibinan: Let's move it to `AclTestBeans`.

api/src/test/java/com/epam/pipeline/acl/entity/EntityApiServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldLoadAvailableEntityForAdmin() {
+        doReturn(aclEntityMap).when(mockHierarchicalEntityManager).loadAvailable(aclSid, AclClass.DATA_STORAGE);
+
+        assertThat(entityApiService.loadAvailable(aclSid, AclClass.DATA_STORAGE)).isEqualTo(aclEntityMap);
+    }
+
+    @Test
+    @WithMockUser()
```
tcibinan: Not required brackets can be omitted.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/app/AclTestConfiguration.java
tcibinan: Class `AclTestConfiguration` configures environment for test classes which doesn't have anything in common with the pull request test classes. The change is only required because acl classes are moved between packages.

tcibinan (approved):

Issue #1590, opened 2020-11-23
Allow to create run without mounts


mzueva:
**Background**
At the moment it is possible to limit which storages will be mounted to the run using `Limit mounts` option. This option doesn't allow to launch a run without mounts at all while it could be useful for some cases.

**Approach**
`Limit mounts` shall accept `None` value and in this case no storages shall be mounted to the run.

mzueva:
@backported to `release/0.16`

rodichenko:
@sidoruka backported to `release/0.16` (ba620c74fa96b7325b2410b76671cf53f7f8e0ca)

NShaforostov:
I've tried to verify.
In general, works well.

Were tested:
- launch without mounts for tools
- launch without mounts for pipelines
- launch without mounts for detach configurations
- launch without mounts from default saved settings and from launch page

Only single note:
when for the tool, in its settings, the checkbox "**Do not mount storages**" is checked, but user unchecks that checkbox at the launch form before the launch - the run still launches with no mounts. (_Expected the launch with all available storages will be mounted_).
Seems that this refers to #297 and https://github.com/epam/cloud-pipeline/issues/1457#issuecomment-702748443.

NShaforostov:
**Test case** was created by 9922a75aab32922b9fd8d8de9ce5072db656b2c2 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/testcases/UI/Run_tools_in_sandbox/Limit_mounts/1590.md).

NShaforostov:
**Docs** were updated via #1516.

NShaforostov:
**e2e-tests** were implemented via #1853.

NShaforostov:
Daily e2e-tests results are located and updated [here](https://github.com/epam/cloud-pipeline/blob/e2e-reports/e2e/reports/AWS/test/classes/com.epam.pipeline.autotests.LaunchLimitMountsTest.html) (for `AWS`).

PR #1596, opened 2020-11-23
issue #1590 work with limit mounts when it is 'none'

SilinPavel:
Handle special case for MountLimits:  'none'

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
@@ -804,7 +804,7 @@ public class PipelineRunManager {
                 .map(Map.Entry::getValue)
                 .flatMap(pipeConfValueVO -> {
                             String limitMounts = pipeConfValueVO.getValue();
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -800,10 +802,15 @@ private boolean checkRunForSensitivity(final Map<String, PipeConfValueVO> parame
         List<Long> datastorageIds = MapUtils.emptyIfNull(parameters).entrySet().stream()
                 .filter(v -> v.getKey().equals(CP_CAP_LIMIT_MOUNTS))
                 .map(Map.Entry::getValue)
-                .flatMap(pipeConfValueVO -> Arrays.stream(
-                        StringUtils.commaDelimitedListToStringArray(pipeConfValueVO.getValue()))
+                .flatMap(pipeConfValueVO -> {
+                            String limitMounts = pipeConfValueVO.getValue();
+                            if (limitMounts.equalsIgnoreCase(LIMIT_MOUNTS_NONE)) {
```
mzueva: It is better to flip equals to prevent NPE :  `LIMIT_MOUNTS_NONE.equalsIgnoreCase(limitMounts)`

Issue #1564, opened 2020-11-13
User Group deleted from system remains in the entities permissions list and can't be removed from it


maryvictol:
**Describe the bug**
User Group deleted from system remains in the entities permissions list and can't be removed from it.

**To Reproduce**
1. Login as admin
2. Create new user group (for ex. TEST_GROUP) or use existing
3. Open any entity (pipeline/folder/storage...) and give access to it for that user group (via Edit >> Permissions)
3. Go to Settings >> User Management >> Groups. Delete user group from step 2
4. Return to entity from step 3 and open Edit >> Permission tab
5. See error:
    - Deleted user group is displayed in Groups and Users list with prefix ROLE_
    - Error "`Failed to find user by name ROLE_TEST_GROUP`" appears when try to delete user group from list

**Expected behavior**
User Group should be removed from all entities permissions lists when it's deleted from system

**Screenshots**
![image](https://user-images.githubusercontent.com/47791878/99064975-c1775c00-25b7-11eb-9fc4-6231a372e9dc.png)

**Environment:**
 - Cloud Provider: AWS
 - Version 0.17.0.8115
 - Browser: chrome

PR #1598, opened 2020-11-23
Issue #1564 Deleted user group remains in permissions tab

Wedds:
This PR is related to issue #1564

Spring caches ACL entities and the cache is updating, when ACL entry is removed. But if the authority, that is used in the ACL entity, was removed - invalidating nor update are not triggered. It leads to displaying deleted entities in permission tabs.
From now, a search for ACL entries, that would be affected by authority removal, is performed, and related data will be deleted from the ACL cache.


SilinPavel:
@mzueva @Wedds  It seems that this PR should solve https://github.com/epam/cloud-pipeline/issues/2145#issuecomment-915952328. How about to resume an activity for this PR, I've merged develop here and resolved problems with tests, now it should be fine.

mzueva (approved):

PR #1602, opened 2020-11-24
Issue 1404: Implemented tests for acl layer notification package

cryteq:
This PR is related to issue #1404 and contains acl tests implementation for `notification` package


api/src/main/java/com/epam/pipeline/acl/notification/SystemNotificationApiService.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 EPAM Systems, Inc. (https://www.epam.com/)
```
tcibinan, line 2: This is not directly connected to the pull request but could I ask you to replace the usage of `SystemNotificationApiService.ADMIN_ROLE` with `AclExpressions.ADMIN_ONLY` in the class?

tcibinan (approved):

PR #1612, opened 2020-11-27
Issue #1404: Implemented tests for Metadata package acl layer

youkofan:
Issue #1404


api/src/main/java/com/epam/pipeline/acl/metadata/MetadataEntityApiService.java
```diff
@@ -129,7 +132,7 @@ public MetadataEntity deleteMetadataItemKey(Long id, String key) {
     public Collection<MetadataClassDescription> getMetadataFields(Long folderId) {
         return metadataEntityManager.getMetadataFields(folderId);
     }
-
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
tcibinan: I wonder how it ended up there :)

It is a good practice to review your pull request changes right before submitting it on GitHub. You can do this either in your IDE or in GitHub GUI during pull request creation.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
```diff
@@ -0,0 +1,453 @@
...
+        doReturn(metadataEntry).when(mockMetadataManager).updateMetadataItemKey(metadataVO);
+
+        assertThat(metadataApiService.updateMetadataItemKey(metadataVO)).isEqualTo(metadataEntry);
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldUpdateMetadataItemKeyWhenPermissionIsGranted() {
+        doReturn(metadataEntry).when(mockMetadataManager).updateMetadataItemKey(metadataVO);
+        doReturn(entity).when(mockEntityManager).load(ENTITY_ACL_CLASS, ID);
```
tcibinan: Looks like we can use `mockLoadEntity` here as well.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldDenyUpdateMetadataItemKeyWhenPermissionIsNotGranted() {
+        doReturn(metadataEntry).when(mockMetadataManager).updateMetadataItemKey(metadataVO);
+        mockLoadEntity(entity, ID);
+
+        assertThrows(AccessDeniedException.class, () -> metadataApiService.updateMetadataItemKey(metadataVO));
+    }
```
tcibinan: I suppose that these two test scenarios can be called `shouldUpdateMetadataItemKeyForOwner` and `shouldDenyUpdateMetadataItemKeyForNonOwner`. It clarifies what kind of permissions are tested.

Could you please also check other test methods for the same issue in the test class below.

api/src/test/java/com/epam/pipeline/test/acl/AbstractAclTest.java
```diff
@@ -148,7 +149,16 @@ public Sid toSid() {
         return list;
     }
 
+    protected void mockAuthentication(final String user) {
```
tcibinan: I'm wondering if we can remove existing `mockAuthUser` method and rename `mockAuthentication` to `mockAuthUser`. I don't think that there is a need to have both methods.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
```diff
@@ -0,0 +1,453 @@
...
+
+    @Test
+    @WithMockUser
+    public void shouldDenySearchMetadataByClassAndKeyValueWhenPermissionIsNotGranted() {
+        initAclEntity(anotherEntity);
+        doReturn(mutableListOf(anotherEntityVO)).when(mockMetadataManager)
+                .searchMetadataByClassAndKeyValue(ENTITY_ACL_CLASS, TEST_STRING, TEST_STRING);
+        mockLoadEntity(anotherEntity, ID_2);
+        mockAuthentication(SIMPLE_USER);
```
tcibinan: What do you think about moving `mockAuthentication` call to the test class setup method? User name can be retrieved from the security context. It will also require to change some of the method `@WithMockUser` annotations.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataEntityApiServiceTest.java
```diff
@@ -0,0 +1,587 @@
...
+    public void shouldDenyDeleteMetadataClassForNotAdmin() {
+        doReturn(metadataClass).when(mockMetadataEntityManager).deleteMetadataClass(ID);
+
+        assertThrows(AccessDeniedException.class, () -> entityApiService.deleteMetadataClass(ID));
+    }
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldUpdateMetadataClassForAdmin() {
+        doReturn(metadataClass).when(mockMetadataEntityManager).updateExternalClassName(ID, FireCloudClass.PARTICIPANT);
```
tcibinan: It would be nice to extract `FireCloudClass.PARTICIPANT` to the class field.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataEntityApiServiceTest.java
```diff
@@ -0,0 +1,587 @@
...
+    }
+
+    @Test
+    @WithMockUser
+    public void shouldDenyCreateMetadataClassForNotAdmin() {
+        doReturn(metadataClass).when(mockMetadataEntityManager).createMetadataClass(TEST_STRING);
+
+        assertThrows(AccessDeniedException.class, () -> entityApiService.createMetadataClass(TEST_STRING));
+    }
+
```
tcibinan, line 108: Let's add test for `MetadataEntityApiService.loadAllMetadataClasses` even if there is no acl on the method yet. Let's test that it works for a regular user.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataEntityApiServiceTest.java
```diff
@@ -0,0 +1,587 @@
...
+    public void shouldFilterMetadataForAdmin() {
+        doReturn(pagedResult).when(mockMetadataEntityManager).filterMetadata(metadataFilter);
+
+        assertThat(entityApiService.filterMetadata(metadataFilter)).isEqualTo(pagedResult);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldFilterMetadataWhenPermissionIsGranted() {
+        initAclEntity(folder, AclPermission.WRITE);
```
tcibinan: I imagine that READ permission would be enough in this case.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataEntityApiServiceTest.java
```diff
@@ -0,0 +1,587 @@
...
+    public void shouldUploadMetadataFromFileForAdmin() {
+        doReturn(metadataEntities).when(mockMetadataUploadManager).uploadFromFile(ID, file);
+
+        assertThat(entityApiService.uploadMetadataFromFile(ID, file)).isEqualTo(metadataEntities);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER, roles = ENTITIES_MANAGER_ROLE)
+    public void shouldUploadMetadataFromFileWhenPermissionIsGranted() {
+        initAclEntity(folder, AclPermission.READ);
```
tcibinan: Seems like WRITE permission is required for the testing method to work. I'm not pretty sure how it passes right now but it should be investigated.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataEntityApiServiceTest.java
```diff
@@ -0,0 +1,587 @@
...
+
+        entityApiService.deleteMetadataFromProject(ID, TEST_STRING);
+
+        verify(mockMetadataEntityManager).deleteMetadataEntitiesInProject(ID, TEST_STRING);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldDeleteMetadataFromProjectWhenPermissionIsGranted() {
+        initAclEntity(folder, AclPermission.READ);
```
tcibinan: Wrong permissions again. How does this test pass?

api/src/test/java/com/epam/pipeline/manager/metadata/MetadataApiServiceTest.java
```diff
@@ -16,6 +16,7 @@
```
tcibinan: This test class should be removed. But before we can do that we should either ensure that all tests in the class are covered by some tests in other classes or move them from the deleting class to some other classes.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
tcibinan: Discussed and refused.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
```diff
@@ -85,19 +111,37 @@ public void shouldUpdateMetadataItemKeyForAdmin() {
...
+    public void shouldUpdateMetadataItemKeyForPipelineUser() {
+        doReturn(pipelineUserEntry).when(mockMetadataManager).updateMetadataItemKey(pipelineUserVO);
+        mockAuthAndLoadUser();
+
+        assertThat(metadataApiService.updateMetadataItemKey(pipelineUserVO)).isEqualTo(pipelineUserEntry);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER, roles = ADMIN_ROLE)
+    public void shouldUpdateMetadataItemKeyForRole() {
```
tcibinan: This method should have suffix `ForAdmin`, don't you think?

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
```diff
@@ -85,19 +111,37 @@ public void shouldUpdateMetadataItemKeyForAdmin() {
...
-        doReturn(entity).when(mockEntityManager).load(ENTITY_ACL_CLASS, ID);
+        mockLoadEntity(entityWithOwner, ID);
         mockAuthUser(SIMPLE_USER);
 
         assertThat(metadataApiService.updateMetadataItemKey(metadataVO)).isEqualTo(metadataEntry);
     }
 
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldUpdateMetadataItemKeyForPipelineUser() {
```
tcibinan, line 132: Let's add the similar test to check if an admin is able to perform the operation for pipeline user metadata.

api/src/test/java/com/epam/pipeline/acl/metadata/MetadataApiServiceTest.java
```diff
@@ -450,4 +693,9 @@ public void shouldDenyGetMetadataKeysWithoutUserRole() {
     private void mockLoadEntity(final AbstractSecuredEntity entity, final Long id) {
         doReturn(entity).when(mockEntityManager).load(ENTITY_ACL_CLASS, id);
     }
+
+    private void mockAuthAndLoadUser() {
```
tcibinan: Seems like `mockUser` would be quite enough for this helper method.

tcibinan (changes requested):

tcibinan (approved):

PR #1629, opened 2020-12-03
Issue #1404: Implemented tests for User package acl layer

youkofan:
Issue #1404


api/src/test/java/com/epam/pipeline/acl/user/RoleApiServiceTest.java
```diff
@@ -0,0 +1,187 @@
...
+    @Test
+    public void shouldLoadRoles() {
+        doReturn(roleList).when(mockRoleManager).loadAllRoles(false);
+
+        assertThat(roleApiService.loadRoles()).isEqualTo(roleList);
+    }
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldLoadRolesForAdmin() {
```
tcibinan: Please use `Role` rather than `Roles` in the test name since the test checks the method returning only a single queue.

tcibinan (changes requested):

tcibinan (approved):

PR #1635, opened 2020-12-08
Issue #1405: Implemented tests for controller layer User package

youkofan:
Issue #1405

### DIFF
--- a/api/src/test/java/com/epam/pipeline/controller/user/RoleControllerTest.java
+++ b/api/src/test/java/com/epam/pipeline/controller/user/RoleControllerTest.java
@@ -86,7 +86,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailLoadRoles() {
+    public void shouldFailLoadRolesForUnauthorizedUser() {
         performUnauthorizedRequest(get(LOAD_ALL_URL));
     }
 
@@ -103,7 +103,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailAssignRole() {
+    public void shouldFailAssignRoleForUnauthorizedUser() {
         performUnauthorizedRequest(post(String.format(ASSIGN_URL, ID)));
     }
 
@@ -120,7 +120,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailRemoveRole() {
+    public void shouldFailRemoveRoleForUnauthorizedUser() {
         performUnauthorizedRequest(delete(String.format(REMOVE_URL, ID)));
     }
 
@@ -139,7 +139,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailCreateRole() {
+    public void shouldFailCreateRoleForUnauthorizedUser() {
         performUnauthorizedRequest(post(CREATE_URL));
     }
 
@@ -156,7 +156,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailUpdateRole() {
+    public void shouldFailUpdateRoleForUnauthorizedUser() {
         performUnauthorizedRequest(put(String.format(ROLE_ID_URL, ID)));
     }
 
@@ -172,7 +172,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailGetRole() {
+    public void shouldFailGetRoleForUnauthorizedUser() {
         performUnauthorizedRequest(get(String.format(ROLE_ID_URL, ID)));
     }
 
@@ -188,7 +188,7 @@ public class RoleControllerTest extends AbstractControllerTest {
     }
 
     @Test
-    public void shouldFailDeleteRole() {
+    public void shouldFailDeleteRoleForUnauthorizedUser() {
         performUnauthorizedRequest(delete(String.format(ROLE_ID_URL, ID)));
     }
 }
--- a/api/src/test/java/com/epam/pipeline/controller/user/RoleControllerTest.java
+++ b/api/src/test/java/com/epam/pipeline/controller/user/RoleControllerTest.java
@@ -1,0 +1,188 @@
+/*
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.controller.user;
+
+import com.epam.pipeline.controller.vo.user.RoleVO;
+import com.epam.pipeline.entity.user.ExtendedRole;
+import com.epam.pipeline.entity.user.Role;
+import com.epam.pipeline.manager.user.RoleApiService;
+import com.epam.pipeline.test.creator.user.UserCreatorUtils;
+import com.epam.pipeline.test.web.AbstractControllerTest;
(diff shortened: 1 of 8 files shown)


api/src/test/java/com/epam/pipeline/test/creator/security/SecurityCreatorUtils.java
```diff
@@ -35,6 +35,8 @@
     public static final TypeReference<Result<Map<AclClass, List<S3bucketDataStorage>>>> ACL_SECURED_ENTITY_MAP_TYPE =
             new TypeReference<Result<Map<AclClass, List<S3bucketDataStorage>>>>() {};
     public static final TypeReference<JwtRawToken> JWT_RAW_TOKEN_INSTANCE_TYPE = new TypeReference<JwtRawToken>() {};
+    public static final TypeReference<Result<JwtRawToken>>JWT_RAW_TOKEN_TYPE =
```
tcibinan: Please add a space between the field's type and its name.

api/src/test/java/com/epam/pipeline/test/creator/user/UserCreatorUtils.java
```diff
@@ -45,6 +69,18 @@ public static PipelineUser getPipelineUser(final String name) {
         return pipelineUser;
     }
 
+    public static PipelineUser getPipelineUser() {
+        PipelineUser pipelineUser = new PipelineUser(TEST_STRING);
```
tcibinan: I suppose the overloaded `getPipelineUser` method can be used rather than the constructor call..

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -197,7 +197,8 @@ public void performRequestWithoutResponse(final MockHttpServletRequestBuilder re
     public MultiValueMap<String, String> multiValueMapOf(Object... objects) {
         final MultiValueMap<String, String> map = new LinkedMultiValueMap<>();
         for (int i = 0; i < objects.length; i += 2) {
-            map.add(String.valueOf(objects[i]), String.valueOf(objects[i + 1]));
+            map.add(String.valueOf(objects[i]).replaceAll("[\\[\\]]", ""),
+                    String.valueOf(objects[i + 1]).replaceAll("[\\[\\]]", ""));
```
tcibinan: The current approach is error-prone. Imagine if some controller method accepts just a string and not a list. If we try to pass f.e. the following string `Array[Sample]` to the method using `multiValueMapOf` it will be cut to `ArraySample`. Imagine debugging something like this :)

As long as this special processing is required for only `List` type then we can handle it explicitly using `instanceof`. Additionally I assume there is no need to process the map keys. Probably we can even check that the keys are of `String` type and fail otherwise.

api/src/test/java/com/epam/pipeline/controller/user/RoleControllerTest.java
```diff
@@ -0,0 +1,194 @@
...
+        doReturn(roleList).when(mockRoleApiService).loadRoles();
+
+        final MvcResult mvcResult = performRequest(get(LOAD_ALL_URL).params(multiValueMapOf(LOAD_USERS, false)));
+
+        verify(mockRoleApiService).loadRoles();
+        assertResponse(mvcResult, roleList, UserCreatorUtils.COLLECTION_ROLE_INSTANCE_TYPE);
+    }
+
+    @Test
+    public void shouldFailLoadRoles() {
```
tcibinan: Could you please add `ForUnauthorizedUser` suffix to all the similar tests?

api/src/test/java/com/epam/pipeline/controller/user/UserControllerTest.java
```diff
@@ -0,0 +1,504 @@
...
+
+    @Autowired
+    private UserApiService mockUserApiService;
+
+    @Autowired
+    private AuthManager mockAuthManager;
+
+    @Test
+    @WithMockUser
+    public void shouldGetSettings() {
```
tcibinan: I see that in controller the method is called `getSettings` but it seems misleading. Probably we can call the test method `shouldGenerateToken`. Please also rename all test methods for the same controller method.

api/src/test/java/com/epam/pipeline/controller/user/UserControllerTest.java
```diff
@@ -0,0 +1,504 @@
...
+    public void shouldRedirectWithFormRouteType() throws Exception {
+        doReturn(token).when(mockAuthManager).issueTokenForCurrentUser(null);
+
+        final MvcResult mvcResult = performRequest(get(ROUTE_URL)
+                .params(multiValueMapOf(URL, TEST_STRING,
+                                        TYPE, RouteType.FORM)),
+                                        TEXT_HTML_UTF8_CONTENT_TYPE);
+
+        verify(mockAuthManager).issueTokenForCurrentUser(null);
+
```
tcibinan: The line break seems non required here.

api/src/test/java/com/epam/pipeline/controller/user/UserControllerTest.java
```diff
@@ -0,0 +1,504 @@
...
+        assertThat(mvcResult.getResponse().getContentAsString()).contains(redirectForm);
+    }
+
```
tcibinan, line 198: The controller method `/route` is one of a very few methods accessible by so-called anonymous users. Currently it is the only one if I'm correct. Could you please add the test that checks that a user with a single role `ROLE_ANONYMOUS_USER` without even `ROLE_USER` can access the method.

api/src/test/java/com/epam/pipeline/controller/user/UserControllerTest.java
```diff
@@ -0,0 +1,504 @@
...
+    @WithMockUser
+    public void shouldExportUsers() throws Exception {
+        final String content = getObjectMapper().writeValueAsString(userExportVO);
+        doReturn(TEST_ARRAY).when(mockUserApiService).exportUsers(userExportVO);
+
+        final MvcResult mvcResult = performRequest(post(USER_EXPORT_URL).content(content),
+                EXPECTED_CONTENT_TYPE, MediaType.APPLICATION_OCTET_STREAM_VALUE);
+
+        verify(mockUserApiService).exportUsers(userExportVO);
+        assertFileResponse(mvcResult, "users.csv", TEST_ARRAY);
```
tcibinan: Let's extract `user.csv` string to the class field.

api/src/test/java/com/epam/pipeline/controller/user/UserControllerTest.java
```diff
@@ -0,0 +1,504 @@
...
+    @WithMockUser
+    public void shouldUpsertGroupBlockingStatusByPutRequest() {
+        doReturn(groupStatus).when(mockUserApiService).upsertGroupBlockingStatus(TEST_STRING, true);
+
+        final MvcResult mvcResult = performRequest(put(String.format(GROUP_NAME_BLOCK_URL, TEST_STRING))
+                .params(multiValueMapOf(BLOCK_STATUS, true)));
+
+        verify(mockUserApiService).upsertGroupBlockingStatus(TEST_STRING, true);
+        assertResponse(mvcResult, groupStatus, UserCreatorUtils.GROUP_STATUS_INSTANCE_TYPE);
+    }
```
tcibinan: Let's use a single methods to test these calls. See `ServerlessConfigurationControllerTest.shouldRun` for a similar case.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -197,8 +199,11 @@ public void performRequestWithoutResponse(final MockHttpServletRequestBuilder re
     public MultiValueMap<String, String> multiValueMapOf(Object... objects) {
         final MultiValueMap<String, String> map = new LinkedMultiValueMap<>();
         for (int i = 0; i < objects.length; i += 2) {
-            map.add(String.valueOf(objects[i]).replaceAll("[\\[\\]]", ""),
-                    String.valueOf(objects[i + 1]).replaceAll("[\\[\\]]", ""));
+            if (objects[i + 1] instanceof List) {
+                map.add(String.valueOf(objects[i]), String.valueOf(objects[i + 1]).replaceAll("[\\[\\]]", ""));
```
tcibinan: I suppose it's cleaner to use `String.valueOf` for each of the list elements and just join them after using f.e. `Collectors.joining`.

tcibinan (changes requested):

tcibinan (approved):

PR #1642, opened 2020-12-09
Issue #1404: Implemented tests for acl layer PipelineApiService

youkofan:
Issue #1404

### DIFF
--- a/api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceGitTest.java
+++ b/api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceGitTest.java
@@ -1,0 +1,546 @@
+/*
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.acl.pipeline;
+
+import com.epam.pipeline.controller.vo.PipelineSourceItemVO;
+import com.epam.pipeline.controller.vo.PipelineSourceItemsVO;
+import com.epam.pipeline.controller.vo.UploadFileMetadata;
+import com.epam.pipeline.entity.git.GitCommitEntry;
+import com.epam.pipeline.entity.git.GitCredentials;
+import com.epam.pipeline.entity.git.GitRepositoryEntry;
+import com.epam.pipeline.entity.git.GitTagEntry;
+import com.epam.pipeline.entity.pipeline.Pipeline;
+import com.epam.pipeline.exception.git.GitClientException;
+import com.epam.pipeline.manager.git.GitManager;
+import com.epam.pipeline.manager.pipeline.PipelineManager;
+import com.epam.pipeline.security.acl.AclPermission;
+import com.epam.pipeline.test.acl.AbstractAclTest;
+import com.epam.pipeline.test.creator.git.GitCreatorUtils;
+import com.epam.pipeline.test.creator.pipeline.PipelineCreatorUtils;
+import org.junit.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.test.context.support.WithMockUser;
+
+import java.util.Collections;
+import java.util.List;
+
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_ARRAY;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_INT;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+import static com.epam.pipeline.util.CustomAssertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.doReturn;
+
+public class PipelineApiServiceGitTest extends AbstractAclTest {
+
+    protected final Pipeline pipeline = PipelineCreatorUtils.getPipeline(ANOTHER_SIMPLE_USER);
+    private final GitTagEntry gitTagEntry = GitCreatorUtils.getGitTagEntry();
+    private final GitCommitEntry gitCommitEntry = GitCreatorUtils.getGitCommitEntry();
+    private final PipelineSourceItemVO sourceItemVO = PipelineCreatorUtils.getPipelineSourceItemVO();
+    private final PipelineSourceItemsVO sourceItemsVO = PipelineCreatorUtils.getPipelineSourceItemsVO();
+    private final GitCredentials gitCredentials = GitCreatorUtils.getGitCredentials();
+    private final GitRepositoryEntry gitRepositoryEntry = GitCreatorUtils.getGitRepositoryEntry();
+    private final UploadFileMetadata fileMetadata = PipelineCreatorUtils.getUploadFileMetadata();
+    private final List<GitRepositoryEntry> gitRepositoryEntries = Collections.singletonList(gitRepositoryEntry);
+    private final List<UploadFileMetadata> files = Collections.singletonList(fileMetadata);
+
+    @Autowired
+    private PipelineApiService pipelineApiService;
+    @Autowired
+    private PipelineManager mockPipelineManager;
+    @Autowired
+    private GitManager mockGitManager;
+
+    @Test
+    @WithMockUser(roles = ADMIN_ROLE)
+    public void shouldLoadRevisionForAdmin() throws GitClientException {
+        doReturn(pipeline).when(mockPipelineManager).load(ID);
+        doReturn(gitTagEntry).when(mockGitManager).loadRevision(pipeline, TEST_STRING);
+
+        assertThat(pipelineApiService.loadRevision(ID, TEST_STRING)).isEqualTo(gitTagEntry);
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldLoadRevisionWhenPermissionIsGranted() throws GitClientException {
+        initAclEntity(pipeline, AclPermission.READ);
+        doReturn(pipeline).when(mockPipelineManager).load(ID);
+        doReturn(gitTagEntry).when(mockGitManager).loadRevision(pipeline, TEST_STRING);
+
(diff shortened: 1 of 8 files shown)


api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceGitTest.java
```diff
@@ -0,0 +1,546 @@
...
+        doReturn(pipeline).when(mockPipelineManager).load(ID);
+        doReturn(gitTagEntry).when(mockGitManager).loadRevision(pipeline, TEST_STRING);
+
+        assertThrows(AccessDeniedException.class, () -> {
+            try {
+                pipelineApiService.loadRevision(ID, TEST_STRING);
+            } catch (GitClientException e) {
+                e.printStackTrace();
+            }
+        });
```
tcibinan: I hope we can get rid of checked exceptions one day but for now we can at least improve the call site. It seams reasonable to create `CustomAssertions.assertThrowsChecked` method which accepts checked lambdas.

To manage checked lambdas you can define a special functional interface `CheckedRunnable` with a single checked method. This functional interface can be used as a lambda parameter type for the new method.

It allows to simplify the call up to the following state:
```
assertThrowsChecked(AccessDeniedException.class, () -> pipelineApiService.loadRevision(ID, TEST_STRING));
```

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,831 @@
...
+    @Autowired
+    private PipelineVersionManager mockVersionManager;
+    @Autowired
+    private InstanceOfferManager mockInstanceOfferManager;
+    @Autowired
+    private PipelineFileGenerationManager mockFileGenerationManager;
+    @Autowired
+    private DocumentGenerationPropertyManager mockPropertyManager;
+    @Autowired
+    private GrantPermissionManager spyGrantPermissionManager;
```
tcibinan, line 89: As long as we share spring context between tests we should be extremely careful with what we do with spy objects. The most safe approach would be to reset spy objects after all tests in a class.

What do you think? Is there any real risk about non resetting the spy?

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,831 @@
...
+                pipelineApiService.create(pipelineVO);
+            } catch (GitClientException e) {
+                e.printStackTrace();
+            }
+        });
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldDenyCreatePipelineForUserRole() throws GitClientException {
```
tcibinan: Seems better to call it `shouldDenyCreatePipelineWithoutPipelineManagerRole`.

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,831 @@
...
+                pipelineApiService.create(pipelineVO);
+            } catch (GitClientException e) {
+                e.printStackTrace();
+            }
+        });
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER, roles = PIPELINE_MANAGER_ROLE)
+    public void shouldDenyCreatePipelineWhenPermissionIsNotGranted() throws GitClientException {
```
tcibinan: Let's be more explicit: `shouldDenyCreatePipelineWhenPermissionForParentFolderIsNotGranted`.

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,831 @@
...
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldDenyLoadAllRunsByPipelineWhenPermissionIsNotGranted() {
+        initAclEntity(pipeline);
+        doReturn(pipelineRunList).when(mockPipelineRunManager).loadAllRunsByPipeline(ID);
+
+        assertThrows(AccessDeniedException.class, () -> pipelineApiService.loadAllRunsByPipeline(ID));
+    }
+
```
tcibinan, line 429: It looks like `PipelineApiService.loadAllRunsByPipeline` method may return runs for which a user doesn't have sufficient permissions.

Could you please check if this is really what's happening. If it is then we should add a test for it and create an issue asking for a fix.

tcibinan: The problem appears when a user has permissions for a pipeline but doesn't have permissions for a run of that pipeline.

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,831 @@
...
+        doReturn(revisionList).when(mockVersionManager).loadAllVersionFromGit(ID);
+
+        assertThrows(AccessDeniedException.class, () -> {
+            try {
+                pipelineApiService.loadAllVersionFromGit(ID);
+            } catch (GitClientException e) {
+                e.printStackTrace();
+            }
+        });
+    }
```
tcibinan: Probably these tests as well as all others depending on `PipelineVersionManager` can be moved to `PipelineApiServiceGitTest`.

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,831 @@
...
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_INT;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+import static com.epam.pipeline.test.creator.CommonCreatorConstants.WRITE_PERMISSION;
+import static com.epam.pipeline.util.CustomAssertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.doReturn;
+
+public class PipelineApiServiceTest extends AbstractAclTest {
+
+    protected final Pipeline pipeline = PipelineCreatorUtils.getPipeline(ANOTHER_SIMPLE_USER);
```
tcibinan: Replace `protected` with `private`.

api/src/test/java/com/epam/pipeline/test/creator/pipeline/PipelineCreatorUtils.java
```diff
@@ -30,9 +42,18 @@ private PipelineCreatorUtils() {
...
+        return pipeline;
+    }
+
+    public static Pipeline getPipeline(final Long id, final String owner, final Long parentId) {
+        final Pipeline pipeline = new Pipeline();
+        pipeline.setId(id);
+        pipeline.setOwner(owner);
+        pipeline.setParentFolderId(parentId);
         return pipeline;
     }
```
tcibinan: It's a good practice to call overloaded methods from each other. 

In this case it is possible to have something like the following with some modifications applied:

```java
    public static Pipeline getPipeline(final String owner) {
        return getPipeline(ID, owner, ID);
    }
```

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
tcibinan: It turned out to be a problem for a separate issue. Closing for now.

api/src/test/java/com/epam/pipeline/test/creator/pipeline/PipelineCreatorUtils.java
```diff
@@ -41,19 +41,17 @@ private PipelineCreatorUtils() {
...
 
-    public static Pipeline getPipeline(final Long id, final String owner, final Long parentId) {
-        final Pipeline pipeline = new Pipeline();
-        pipeline.setId(id);
-        pipeline.setOwner(owner);
-        pipeline.setParentFolderId(parentId);
+    public static Pipeline getPipeline(final String owner) {
+        final Pipeline pipeline = getPipeline(ID, owner, ID);
         return pipeline;
```
tcibinan: There is no need for an additional variable.

api/src/test/java/com/epam/pipeline/util/CustomAssertions.java
```diff
@@ -61,4 +61,20 @@ public static void assertThrows(final Runnable runnable) {
...
+            if (expectedExceptionClass.isInstance(e)) {
+                return;
+            } else {
+                throw new AssertionError(
+                        String.format("Expected exception %s was not thrown, but another exception was: %s.",
+                                expectedExceptionClass.getSimpleName(), e.getClass().getSimpleName()), e);
+            }
+        }
+        fail(String.format("Expected exception %s was not thrown.", expectedExceptionClass.getSimpleName()));
+    }
```
tcibinan: I imagine if we can reuse `assertThrows` method here using `@SneakyThrows` annotation. The call site will look like this:
```
assertThrows(expectedExceptionClass, () -> sneakyThrowing(runnable));
```

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldDenyGetWorkflowGraphWhenPermissionIsNotGranted() {
+        initAclEntity(pipeline);
+        doReturn(taskGraphVO).when(mockVersionManager).getWorkflowGraph(ID, TEST_STRING);
+
+        assertThrows(AccessDeniedException.class, () -> pipelineApiService.getWorkflowGraph(ID, TEST_STRING));
+    }
```
tcibinan: I think these methods should be moved to the git test class.

api/src/test/java/com/epam/pipeline/acl/pipeline/PipelineApiServiceTest.java
```diff
@@ -0,0 +1,774 @@
...
+
+    @Test
+    @WithMockUser(username = SIMPLE_USER)
+    public void shouldDenyRegisterPipelineVersionWhenPermissionIsNotGranted() throws GitClientException {
+        initAclEntity(pipeline);
+        doReturn(revision).when(mockVersionManager).registerPipelineVersion(pipelineVersionVO);
+
+        assertThrowsChecked(AccessDeniedException.class, () ->
+                pipelineApiService.registerPipelineVersion(pipelineVersionVO));
+    }
```
tcibinan: These too.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/util/CustomAssertions.java
tcibinan: I support that something like this should work:
```java
    @SneakyThrows
    public static void sneakyThrowing(final CheckedRunnable runnable) {
        return runnable.run()
    }

    public static void assertThrowsChecked(final Class<? extends Throwable> expectedExceptionClass,
                                           final CheckedRunnable runnable) {
        assertThrows(expectedExceptionClass, () -> sneakyThrowing(runnable));
    }
```

tcibinan (approved):

PR #1650, opened 2020-12-11
Issue #1405: Implemented tests for controller layer pipeline

youkofan:
Issue #1405

### DIFF
--- a/api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
+++ b/api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
@@ -54,6 +54,9 @@ import static com.epam.pipeline.test.creator.CommonCreatorConstants.ID;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_ARRAY;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_INT;
 import static com.epam.pipeline.test.creator.CommonCreatorConstants.TEST_STRING;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.anyListOf;
+import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.verify;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
@@ -84,6 +87,7 @@ public class PipelineControllerTest extends AbstractControllerTest {
     private static final String PIPELINE_ID_FOLDER_URL = PIPELINE_ID_URL + "/folder";
     private static final String PIPELINE_ID_DOCS_URL = PIPELINE_ID_URL + "/docs";
     private static final String PIPELINE_ID_FILE_URL = PIPELINE_ID_URL + "/file";
+    private static final String PIPELINE_ID_FILE_TRUNCATE_URL = PIPELINE_ID_FILE_URL + "/truncate";
     private static final String PIPELINE_ID_FILE_DOWNLOAD_URL = PIPELINE_ID_FILE_URL + "/download";
     private static final String PIPELINE_ID_FILE_UPLOAD_URL = PIPELINE_ID_FILE_URL + "/upload";
     private static final String PIPELINE_ID_FILE_GENERATE_URL = PIPELINE_ID_FILE_URL + "/generate";
@@ -113,6 +117,8 @@ public class PipelineControllerTest extends AbstractControllerTest {
     private static final String URL = "url";
     private static final String PARENT_ID = "parentId";
     private static final String NAME = "name";
+    private static final String FILE_SIZE = "0 Kb";
+    private static final String BYTE_LIMIT = "byteLimit";
 
     private final Pipeline pipeline = PipelineCreatorUtils.getPipeline();
     private final PipelineVO pipelineVO = PipelineCreatorUtils.getPipelineVO();
@@ -239,7 +245,8 @@ public class PipelineControllerTest extends AbstractControllerTest {
                 .loadAllPipelinesWithPermissions(TEST_INT, TEST_INT);
 
         final MvcResult mvcResult = performRequest(get(PIPELINE_LOAD_ALL_PERMISSIONS_URL)
-                .params(multiValueMapOf(PAGE_NUM, TEST_INT, PAGE_SIZE, TEST_INT)));
+                .params(multiValueMapOf(PAGE_NUM, TEST_INT,
+                                        PAGE_SIZE, TEST_INT)));
 
         verify(mockPipelineApiService).loadAllPipelinesWithPermissions(TEST_INT, TEST_INT);
         assertResponse(mvcResult, pipelinesWithPermissionsVO, PipelineCreatorUtils.PIPELINE_WITH_PERMISSIONS_TYPE);
@@ -390,7 +397,8 @@ public class PipelineControllerTest extends AbstractControllerTest {
                 .getInstanceEstimatedPrice(ID, TEST_STRING, TEST_STRING, TEST_STRING, TEST_INT, true, ID);
 
         final MvcResult mvcResult = performRequest(post(String.format(PIPELINE_ID_PRICE_URL, ID)).content(content)
-                .params(multiValueMapOf(VERSION, TEST_STRING, CONFIG, TEST_STRING)));
+                .params(multiValueMapOf(VERSION, TEST_STRING,
+                                        CONFIG, TEST_STRING)));
 
         verify(mockPipelineApiService)
                 .getInstanceEstimatedPrice(ID, TEST_STRING, TEST_STRING, TEST_STRING, TEST_INT, true, ID);
@@ -444,7 +452,9 @@ public class PipelineControllerTest extends AbstractControllerTest {
                 .getPipelineSources(ID, TEST_STRING, TEST_STRING, true, true);
 
         final MvcResult mvcResult = performRequest(get(String.format(PIPELINE_ID_SOURCES_URL, ID))
-                .params(multiValueMapOf(VERSION, TEST_STRING, PATH, TEST_STRING, RECURSIVE, true)));
+                .params(multiValueMapOf(VERSION, TEST_STRING,
+                                        PATH, TEST_STRING,
+                                        RECURSIVE, true)));
 
         verify(mockPipelineApiService).getPipelineSources(ID, TEST_STRING, TEST_STRING, true, true);
         assertResponse(mvcResult, gitRepositoryEntries, GitCreatorUtils.GIT_REPOSITORY_ENTRY_LIST_TYPE);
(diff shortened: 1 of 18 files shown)


api/src/test/java/com/epam/pipeline/test/web/AbstractControllerTest.java
```diff
@@ -72,7 +72,7 @@
     private ObjectMapper deserializationMapper;
 
     @Autowired
-    private JsonMapper objectMapper;
+    protected JsonMapper objectMapper;
```
tcibinan: Probably you can always use `getObjectMapper` in all the subclasses.

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -0,0 +1,827 @@
...
+    private final List<DocumentGenerationProperty> generationProperties =
+            Collections.singletonList(documentGenerationProperty);
+
+    @Autowired
+    private PipelineApiService mockPipelineApiService;
+
+    @Test
+    @WithMockUser
+    public void shouldRegisterPipeline() throws Exception {
+        final String content = getObjectMapper().writeValueAsString(pipelineVO);
```
tcibinan: Could you please extract this call to `stringOf` method of the superclass? I was thinking about this for a long time but now it's the time :) 

Once applied we can inline all the calls in the class slightly reducing the test class size. By the way use the new method only in this pull request. We can upgrade all the old ones later on.

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -0,0 +1,827 @@
...
+    @Test
+    @WithMockUser
+    public void shouldGetPipelineFile() throws Exception {
+        doReturn(TEST_ARRAY).when(mockPipelineApiService).getPipelineFileContents(ID, TEST_STRING, TEST_STRING);
+
+        final MvcResult mvcResult = performRequest(get(String.format(PIPELINE_ID_FILE_URL, ID))
+                .params(multiValueMapOf(VERSION, TEST_STRING,
+                                        PATH, TEST_STRING)));
+        final String contentAsString = mvcResult.getResponse().getContentAsString();
+        final byte[] returnedEntityBody = objectMapper.readValue(contentAsString, byte[].class);
```
tcibinan: Could you please extract these lines along with the corresponding assert to a private method in `PipelineControllerTest`.

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -0,0 +1,827 @@
...
+
+        final MvcResult mvcResult = performRequest(post(String.format(PIPELINE_ID_FILE_UPLOAD_URL, ID))
+                        .params(multiValueMapOf(PATH, TEST_STRING)).content(MULTIPART_CONTENT),
+                MULTIPART_CONTENT_TYPE, EXPECTED_CONTENT_TYPE);
+
+        verify(mockPipelineApiService).uploadFiles(eq(ID), eq(TEST_STRING), anyListOf(UploadFileMetadata.class));
+        final String expectedResult = mvcResult.getResponse().getContentAsString();
+        assertThat(expectedResult).contains(MULTIPART_CONTENT_FILE_NAME)
+                .contains(FILE_SIZE)
+                .contains(MediaType.APPLICATION_OCTET_STREAM_VALUE);
```
tcibinan: Could you please use a regular `assertResponse` method? Also let's check that `pipelineApiService.uploadFiles` is called with the proper file metadata using an argument captor.

The controller method is quite an unusual one so let's test most of its logic.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -600,12 +589,11 @@ public void shouldUploadFile() throws Exception {
...
+        final ArgumentCaptor<UploadFileMetadata> captor = ArgumentCaptor.forClass(UploadFileMetadata.class);
 
-        verify(mockPipelineApiService).uploadFiles(eq(ID), eq(TEST_STRING), anyListOf(UploadFileMetadata.class));
-        final String expectedResult = mvcResult.getResponse().getContentAsString();
-        assertThat(expectedResult).contains(MULTIPART_CONTENT_FILE_NAME)
-                .contains(FILE_SIZE)
-                .contains(MediaType.APPLICATION_OCTET_STREAM_VALUE);
+        verify(mockPipelineApiService).uploadFiles(eq(ID), eq(TEST_STRING),
+                Collections.singletonList(captor.capture()));
+        assertResponse(mvcResult, "[{fileName=file.txt, fileSize=0 Kb, fileType=application/octet-stream}]");
```
tcibinan: Could you please initiate a new instance of expected object? Something like the following:
```java
UploadFileMetadata expectedUploadFileMetadata = new ...;
assertResponse(mvcResult, Collections.singletonList(expectedUploadFileMetadata));
```

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -392,13 +387,12 @@ public void shouldFailGetPipelineCredentialsForUnauthorizedUser() {
...
     public void shouldGetPipelineEstimatedPrice() throws Exception {
-        final String content = getObjectMapper().writeValueAsString(instance);
         doReturn(instancePrice).when(mockPipelineApiService)
                 .getInstanceEstimatedPrice(ID, TEST_STRING, TEST_STRING, TEST_STRING, TEST_INT, true, ID);
 
-        final MvcResult mvcResult = performRequest(post(String.format(PIPELINE_ID_PRICE_URL, ID)).content(content)
-                .params(multiValueMapOf(VERSION, TEST_STRING,
-                                        CONFIG, TEST_STRING)));
+        final MvcResult mvcResult = performRequest(post(String.format(PIPELINE_ID_PRICE_URL, ID))
+                .content(stringOf(instance)).params(multiValueMapOf(VERSION, TEST_STRING,
```
tcibinan: Usually it is a good idea to put chained calls to separate lines:

```java
// BAD
a.b().c()

// GOOD
a.b()
 .c()
```

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
tcibinan: In this case we can also get rid of new `assertResponse(MvcResult, String)` method.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -819,12 +821,11 @@ private void assertResponseAsBytes(MvcResult mvcResult, byte[] bytes) throws Exc
         assertThat(returnedEntityBody).isEqualTo(bytes);
     }
 
-
-    private void assertResponse(MvcResult mvcResult, String expected) throws Exception {
+    private void assertResponseAsList(MvcResult mvcResult) throws Exception{
         final String expectedResult = mvcResult.getResponse().getContentAsString();
+        final List<UploadFileMetadata> expectedList = getObjectMapper().readValue(expectedResult, List.class);
+        final List<UploadFileMetadata> actualList = getObjectMapper().readValue(stringOf(fileMetadataList), List.class);
```
tcibinan: There is no need to convert `List<UploadFileMetadata>` -> `String` -> `List<UploadFileMetadata>`. Additionally it would be nice to extract `fileMetadataList` to a method parameter and name the method something like `assertUnwrappedResponse`.

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -819,12 +821,11 @@ private void assertResponseAsBytes(MvcResult mvcResult, byte[] bytes) throws Exc
         assertThat(returnedEntityBody).isEqualTo(bytes);
     }
 
-
-    private void assertResponse(MvcResult mvcResult, String expected) throws Exception {
+    private void assertResponseAsList(MvcResult mvcResult) throws Exception{
```
tcibinan: Missing space after `Exception`.

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -819,12 +821,11 @@ private void assertResponseAsBytes(MvcResult mvcResult, byte[] bytes) throws Exc
...
         final String expectedResult = mvcResult.getResponse().getContentAsString();
+        final List<UploadFileMetadata> expectedList = getObjectMapper().readValue(expectedResult, List.class);
```
tcibinan: It's better to use a type reference like we always did otherwise the assignment is unchecked.

api/src/test/java/com/epam/pipeline/test/creator/pipeline/PipelineCreatorUtils.java
```diff
@@ -150,7 +150,11 @@ public static PipelineSourceItemsVO getPipelineSourceItemsVO() {
     }
 
     public static UploadFileMetadata getUploadFileMetadata() {
-        return new UploadFileMetadata();
+        final UploadFileMetadata uploadFileMetadata = new UploadFileMetadata();
+        uploadFileMetadata.setFileName("file.txt");
+        uploadFileMetadata.setFileSize("0 Kb");
+        uploadFileMetadata.setFileType("application/octet-stream");
```
tcibinan: Let's pass these values via the method parameters.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/controller/pipeline/PipelineControllerTest.java
```diff
@@ -814,18 +817,19 @@ public void shouldFailCopyPipelineForUnauthorizedUser() {
...
         final String contentAsString = mvcResult.getResponse().getContentAsString();
         final byte[] returnedEntityBody = getObjectMapper().readValue(contentAsString, byte[].class);
 
         assertThat(returnedEntityBody).isEqualTo(bytes);
     }
 
-    private void assertResponseAsList(MvcResult mvcResult) throws Exception{
+    private void assertUnwrappedResponse(final MvcResult mvcResult, final List<UploadFileMetadata> list)
+                                                                                      throws Exception {
         final String expectedResult = mvcResult.getResponse().getContentAsString();
```
tcibinan: It's not the expected result but it's the actual one.

tcibinan (changes requested):

tcibinan (approved):

Issue #1656, opened 2020-12-15
Pass system parameters to autoscaled SGE workers


mzueva:
**Background**
A clear and concise description of what the problem is. Ex. At the moment, Cloud Pipeline does not allow [...] and it would be nice to [...] because [...]

**Approach**

**Other options**

mzueva:
backported to `release/0.16`

PR #1660, opened 2020-12-16
issue #1656 fetch launch system params for the workers

SilinPavel:
This PR relates to #1656 and fetch all parameters that should be pass to workers from master run

### DIFF
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -24,6 +24,7 @@ import multiprocessing
 import requests
 import re
 import sys
+import json
 
 
 class ExecutionError(RuntimeError):
@@ -539,7 +540,7 @@ class GridEngineScaleUpHandler:
 
     def __init__(self, cmd_executor, api, grid_engine, host_storage, instance_helper, parent_run_id, default_hostfile, instance_disk,
                  instance_image, cmd_template, price_type, region_id, polling_timeout=_POLL_TIMEOUT, polling_delay=_POLL_DELAY,
-                 ge_polling_timeout=_GE_POLL_TIMEOUT, instance_family=None, shared_fs_type='lfs', limit_mounts=None):
+                 ge_polling_timeout=_GE_POLL_TIMEOUT, instance_family=None):
         """
         Grid engine scale up implementation. It handles additional nodes launching and hosts configuration (/etc/hosts
         and self.default_hostfile).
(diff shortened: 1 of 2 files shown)


workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -627,19 +627,28 @@ def _launch_additional_worker(self, instance):
...
+        worker_launch_system_params = 'CP_CAP_SGE false ' \
+                                      'CP_CAP_AUTOSCALE false ' \
+                                      'CP_CAP_AUTOSCALE_WORKERS 0 ' \
+                                      'CP_DISABLE_RUN_ENDPOINTS true '
+        for launch_param in system_launch_params:
+            param_name = launch_param.get("name")
+            if launch_param.get("passToWorkers", False) and param_name in master_system_params:
+                worker_launch_system_params = worker_launch_system_params + \
+                                              " {} {}".format(param_name, master_system_params.get(param_name))
+        return worker_launch_system_params
```
tcibinan: Could you please use `'` rather than `"` like in the most of the script?

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -627,19 +627,28 @@ def _launch_additional_worker(self, instance):
...
+        system_launch_params_string = self.api.retrieve_preference('launch.system.parameters', default_value="[]")
+        system_launch_params = json.loads(system_launch_params_string)
+        worker_launch_system_params = 'CP_CAP_SGE false ' \
+                                      'CP_CAP_AUTOSCALE false ' \
+                                      'CP_CAP_AUTOSCALE_WORKERS 0 ' \
+                                      'CP_DISABLE_RUN_ENDPOINTS true '
+        for launch_param in system_launch_params:
+            param_name = launch_param.get("name")
+            if launch_param.get("passToWorkers", False) and param_name in master_system_params:
+                worker_launch_system_params = worker_launch_system_params + \
```
tcibinan: There is a helpful `+=` operator for such appends.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -617,6 +614,9 @@ def scale_up(self, resource):
 
     def _launch_additional_worker(self, instance):
         Logger.info('Launch additional worker.')
+        parent_run = self.api.load_run(self.parent_run_id)
+        worker_launch_system_params = self.fetch_worker_launch_system_params(parent_run)
```
tcibinan: It seems like a good idea to move this initialization logic to the beginning of *main* method. It makes the whole autoscaling launch consistent independently of changes in the system preference.

tcibinan (approved):

mzueva (approved):

Issue #1653, opened 2020-12-14
Node Pools: Allow to filter nodes by labels


mzueva:
**Background**
At the moment it is possible to filter cluster nodes by `runId` and `ip` using `/cluster/node/filter` API method. It would be useful to allow to filter nodes by labels and values as well.

**Approach**
- Extend `FilterNodesVO` object with `labels` [`Map<String, String>`]
- `/cluster/node/filter` API method shall returns only nodes with matching `labels`

PR #1664, opened 2020-12-18
Issue #1653: Added labels filter to the filterNodes method

youkofan:
Issue #1653

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
@@ -59,11 +59,13 @@ import org.springframework.util.Assert;
 import javax.annotation.PostConstruct;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Predicate;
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
```diff
@@ -153,8 +153,14 @@ public void init() {
                         address.getAddress().equalsIgnoreCase(filterNodesVO.getAddress());
                 addressFilter = node ->
                         node.getAddresses() != null && node.getAddresses()
-                                .stream()
-                                .filter(addressEqualsPredicate).count() > 0;
+                                .stream().anyMatch(addressEqualsPredicate);
+            }
+            Predicate<NodeInstance> labelsFilter = node -> true;
```
mzueva: I'd suggest to reuse `labelsMap` (line 145) as it used for filtering by labels - just what we need here.

api/src/main/java/com/epam/pipeline/manager/cluster/NodesManager.java
```diff
@@ -153,8 +153,14 @@ public void init() {
...
+            if (filterNodesVO.getLabels() != null && !filterNodesVO.getLabels().isEmpty()) {
```
mzueva: You can use `MapUtils.isNotEmpty` fro such check

mzueva (changes requested):

mzueva (approved):

Issue #1652, opened 2020-12-14
Node Pools: VMMonitor shall handle pool nodes


mzueva:
**Background**
With implementation of node pools #841 cluster may have nodes with non-numeric `runid` label, e.g. `p-123`. `VMMonitor` service shall handle such nodes and verify their state accordingly.

**Approach**
If a node has non-numeric `runid` label `VMMonitor` service shall check `pool_id` label. If pool label is present and corresponding `NodePool` exists such node shall be considered valid, otherwise a notification shall be send.

PR #1665, opened 2020-12-18
Issue #1652: VMMonitor shall handle pool nodes

cryteq:
This PR is related to issue #1652 and enables monitor to validate nodes with non-numeric `runid` and existing `pool_id`

### DIFF
--- a/vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/vm/VMMonitor.java
+++ b/vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/vm/VMMonitor.java
@@ -48,24 +48,26 @@ import java.util.stream.Collectors;
 @Service
 public class VMMonitor {
 
-    private static final String POOL_ID_LABEL = "poolid";
     private final CloudPipelineAPIClient apiClient;
     private final VMNotifier notifier;
     private final Map<CloudProvider, VMMonitorService> services;
     private final List<String> requiredLabels;
     private final String runIdLabel;
+    private final String poolIdLabel;
 
     public VMMonitor(final CloudPipelineAPIClient apiClient,
                      final VMNotifier notifier,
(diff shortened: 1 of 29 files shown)


vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/vm/VMMonitor.java
```diff
@@ -102,7 +106,7 @@ private void checkVmState(final VirtualMachine vm) {
                 checkMatchingNodes(nodes, vm);
             } else {
                 log.debug("No matching nodes were found for VM {} {}.", vm.getInstanceId(), vm.getCloudProvider());
-                if (!matchingRunExists(vm)) {
+                if (!matchingRunExists(vm) && !poolIdExists(vm)) {
```
mzueva: Pool label exists only on kubernetes nodes, so we shall check it in `checkLabels` method, which operates kubernetes `Node` instead of AWS `VirtualMachine`

mzueva (changes requested):

vm-monitor/src/main/java/com/epam/pipeline/vmmonitor/service/vm/VMMonitor.java
```diff
@@ -124,6 +128,23 @@ private boolean matchingRunExists(final VirtualMachine vm) {
         return false;
     }
 
+    private boolean poolIdExists(final VirtualMachine vm) {
+        log.debug("Checking whether a node pool with corresponding pool id exists.");
```
mzueva, line 132: As I mentioned previously `VirtualMachine ` doesn't have `pool` label, we need to check  `NodeInstance ` tags instead.

mzueva (changes requested):

mzueva (approved):

Issue #1639, opened 2020-12-09
`pipe` shall allow to import users from CSV


sidoruka:
**Background**
At the moment, Cloud Pipeline allows creating users one-by-one via the GUI. If a number of users shall be created - it would be nice to have a batch creation option.

**Approach**
* Introduce a `pipe users import` command:
```
pipe users import --create-user \
                  --create-group \
                  --create-<METADATA_ITEM_NAME>
                  <PATH_TO_CSV>
```
* Command shall import users from the CSV file (specified by `PATH_TO_CSV`), according to the following rules:
* CSV format:
```
UserName,Groups,MetadataItem1,MetadataItem2,MetadataItemN
user1,ROLE_ADMIN|NGS_USERS,Value1,Value2,ValueN
user2,NGS_USERS,Value3,Value4,ValueN
user3,,Value3,Value4,ValueN
user4,NGS_USERS,,,
```
* If the user with `UserName` does not exist:
  * If `--create-user` is specified - the user shall be created
  * Otherwise - skip such user
* If the user is not a member of one of the groups, listed in `Groups` - the user shall be added to the group(s)
* If one of the group(s) does not exist:
  * If `--create-group` is specified - the group shall be created and a user shall be assigned to that group
  * Otherwise - skip the group
* For each of the `MetadataItemX` the following actions shall be performed:
  * If the user does not have an attribute with the name `MetadataItemX` and the attribute's name does not match any existing `SYSTEM DICTIONARY` - it shall be added with a corresponding value
  * If an attribute already exists, but the value is different - it shall be updated with a corresponding value (for the cases, when the attribute's name does not match any existing `SYSTEM DICTIONARY`)
  * If an attribute is an existing `SYSTEM DICTIONARY` and a value is linked to another dictionary - this link shall be handled in the same manner as GUI does (i.e. another attribute shall be added/updated, according to the link)
  * If an attribute is an existing `SYSTEM DICTIONARY` and a value does not exist in that dictionary:
    * If `--create-<METADATA_ITEM_NAME>` is specified (where `<METADATA_ITEM_NAME>` shall match the column name) - the value shall be added as a new dictionary entry
    * Otherwise - skip the metadata item
* If one of the values is empty (i.e. `Groups`, `MetadataItemX`) for a specific user - this field shall be skipped (like `user3` and `user4` in the example above)
* For each of the actions (user created, user assigned to a group, group created, metadata updated for a user, etc.) - a corresponding log string shall be written to `stdout`

[...]

PR #1667, opened 2020-12-18
`pipe` shall allow to import users from CSV

ekazachkova:
The current PR provides implementation for issue #1639 and contains the following changes:
- a new API method `POST users/import` was implemented. This method registers users, roles and metadata specified in input file. Supports optional query parameters:
  - `createUser` - true if user shall be created if not exists. Default: `false`
  - `createGroup` - true if group shall be created if not exists. Default: `false` 
  - `createMetadata` - the list of metadata keys that shall be created if not exists
- a new CLI command `pipe users import <path-to-file>` implemented. This command provides the following options:
  - `--create-user (-cu)` - true if user shall be created if not exists. Default: `false`
  - `--create-group (-cg)` - true if group shall be created if not exists. Default: `false` 
  - `--create-metadata (-cm)` - metadata keys that shall be created. Multiple options supported.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/user/UserImportManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/user/UserImportManager.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.manager.user;
+
+import com.epam.pipeline.dao.user.RoleDao;
+import com.epam.pipeline.entity.metadata.CategoricalAttribute;
+import com.epam.pipeline.entity.metadata.CategoricalAttributeValue;
+import com.epam.pipeline.entity.metadata.MetadataEntry;
+import com.epam.pipeline.entity.metadata.PipeConfValue;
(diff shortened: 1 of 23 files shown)


api/src/main/java/com/epam/pipeline/manager/user/UserImportManager.java
```diff
@@ -0,0 +1,241 @@
...
+                events.addAll(ListUtils.emptyIfNull(processUser(user, createUser, createGroup, categoricalAttributes)));
+            } catch (Exception e) {
+                log.error(String.format("Failed to process user '%s'", user.getUserName()), e);
+                events.add(PipelineUserEvent.error(user.getUserName(), e.getMessage()));
+            }
+        });
+        return events;
+    }
+
+    @Transactional(propagation = Propagation.REQUIRED)
```
mzueva, line 59: I don't think that `@Transactional` will work in this case as it is called from public non-transcation methods in the same class

api/src/main/java/com/epam/pipeline/manager/user/UserImportManager.java
```diff
@@ -0,0 +1,241 @@
...
+                                                       final List<String> attributesToCreate,
+                                                       final MultipartFile file) {
+        final List<PipelineUserEvent> events = new ArrayList<>();
+        final List<CategoricalAttribute> categoricalAttributes = ListUtils
+                .emptyIfNull(categoricalAttributeManager.loadAll());
+        final List<PipelineUserWithStoragePath> users =
+                new UserImporter(events, categoricalAttributes, attributesToCreate).importUsers(file);
+        categoricalAttributeManager.updateCategoricalAttributes(categoricalAttributes);
+
+        users.forEach(user -> {
```
mzueva: can be replaced with `map`

api/src/main/java/com/epam/pipeline/manager/user/UserImportManager.java
```diff
@@ -0,0 +1,241 @@
...
+            try {
+                events.addAll(ListUtils.emptyIfNull(processUser(user, createUser, createGroup, categoricalAttributes)));
+            } catch (Exception e) {
+                log.error(String.format("Failed to process user '%s'", user.getUserName()), e);
```
mzueva: I'd suggest to use `MessageHelper` for message formatting

mzueva (changes requested):

mzueva (approved):

Issue #1646, opened 2020-12-10
Allow to export node monitoring report in Excel format


sidoruka:
**Background**
Relates to #1628
At the moment, Cloud Pipeline allows exporting instance utilization report in the CSV format. Either from GUI or CLI (via #1628).
It would be great to have some charts in the exported file as well. To achieve this - we shall use Excel format with the built-in charts.

**Approach**
* Admin shall be able to specify a location of the Excel template (`application.properties`/`env vars` will do)
* Excel template shall have two sheets:
  * Charts - contains a set of charts, that mimic the Cloud Pipeline GUI and use the data from the `Data` sheet
  * Data - contains a table, similar to the existing CSV export
* User shall be able to choose which format to use for the export (via GUI and CLI): `CSV` or `XLS` (default)

sidoruka:
Please backport to `release/0.16`

rodichenko:
@sidoruka @NShaforostov @mzueva GUI part backported to `release/0.16` (26eaccd8b96e4334028277f8596299d5afaf573c)

NShaforostov:
Verified.
Works well.

Were tested:
- export node utilization data in `xls` via GUI (with different settings - period interval, start/end dates)
- export node utilization data in `xls` via `pipe` CLI (with different options - `--interval`, `--date-from`, `--date-to`, `--output`)
- comparison of reports received by different ways

NShaforostov:
**Docs** were updated via #1516.

PR #1677, opened 2020-12-22
Issue #1646 Add exported monitoring report type selection to CLI

Wedds:
This PR is related to issue #1646

It allows specifying the type of monitoring report, that a user wants to export via pipe-cli (`XLS` or `CSV`).

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -1623,8 +1623,8 @@ def cluster():
 @click.option('-p', '--interval', required=False, help='The time interval. This option shall have the following format:'
                                                        ' <N>m for minutes or <N>h for hours, where <N> is the required '
                                                        'number of minutes/hours. Default: 1m.')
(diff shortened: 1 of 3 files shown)


pipe-cli/pipe.py
```diff
@@ -1623,13 +1623,15 @@ def cluster():
 @click.option('-p', '--interval', required=False, help='The time interval. This option shall have the following format:'
                                                        ' <N>m for minutes or <N>h for hours, where <N> is the required '
                                                        'number of minutes/hours. Default: 1m.')
+@click.option('-rt', '--report-type', required=False, help='Exported report type (case insensitive).'
```
mzueva: Please leave `csv` as default type to keep current behaviour as default. Set value using `default='CSV'`.

mzueva (changes requested):

pipe-cli/pipe.py
Wedds: Done

Issue #1736, opened 2021-01-18
Unit tests: Refactor PipelineWithPermissionsTest


mzueva:
**Background**
At the moment `PipelineWithPermissionsTest` is annotated with `@DirtiesContext` and uses `TestApplicationWithAclSecurity` as context configuration. In order to reduce test startup time we need to remove `@DirtiesContext` annotation and rewrite test using inheritance from `AbstractAclTest`.

PR #1748, opened 2021-01-26
Issue #1736 Refactor PipelineWithPermissionTest

VolLol:
Remove @DirtiesContext, inheritance from AbstractAclTest and rewrite test methods

Issue #1736


api/src/test/java/com/epam/pipeline/test/acl/AbstractAclTest.java
```diff
@@ -29,19 +30,14 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 import java.util.stream.IntStream;
 
-import static org.mockito.Matchers.anyList;
-import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.*;
 import static org.mockito.Mockito.doReturn;
```
tcibinan: It is a good practice to not use asterisk imports. You can disable automated imports optimization in your IDE settings.

api/src/test/java/com/epam/pipeline/test/acl/AbstractAclTest.java
```diff
@@ -107,8 +108,36 @@ protected void initAclEntity(AbstractSecuredEntity entity, List<AbstractGrantPer
...
+                        .forEach(i -> {
+                            AbstractGrantPermission permission = permissions.get(i);
+                            acl.insertAce(i, permissionFactory.buildFromMask(permission.mask), permission.toSid(), true);
+                        });
+            }
+            doReturn(acl).when(aclService).readAclById(eq(objectIdentity), anyList());
+            doReturn(acl).when(aclService).getAcl(eq(entity));
+            doReturn(acl).when(aclService).getOrCreateObjectIdentity(eq(entity));
+            doReturn(acl).when(aclService).createAcl(eq(entity));
+            doReturn(acl).when(aclService).updateAcl(acl);
```
tcibinan: It looks like we can extract this code to a private method which returns an acl object. This way we can get rid of code duplication.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
```diff
@@ -71,80 +67,264 @@
...
-    @WithMockUser(username = TEST_OWNER1)
-    public void testMergePermissions() {
+    public void shouldReturnPermissionsWhenGrantedToPipelineCase1() {
         // key - pipelineId, value - <sid, mask>
-        Map<Long, Map<String, Integer>> expectedMap = new HashMap<>();
-        HashMap<String, Integer> permissionsForPipeline1 = new HashMap<>();
-        HashMap<String, Integer> permissionsForPipeline2 = new HashMap<>();
+        final Map<Long, Map<String, Integer>> expectedMap = new HashMap<>();
+        final Map<String, Integer> permissionsForPipeline1 = new HashMap<>();
+        final Map<String, Integer> permissionsForPipeline2 = new HashMap<>();
```
tcibinan: I suggest moving these declarations closer to the call site.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
```diff
@@ -71,80 +67,264 @@
...
+    @Test
+    public void shouldReturnPermissionsWhenGrantedToPipelineCase2() {
+        // key - pipelineId, value - <sid, mask>
+        final Map<Long, Map<String, Integer>> expectedMap = new HashMap<>();
+        final Map<String, Integer> permissionsForPipeline1 = new HashMap<>();
+        final Map<String, Integer> permissionsForPipeline2 = new HashMap<>();
+
+        final List<Pair<AbstractSecuredEntity, List<AbstractGrantPermission>>> pipelinesAndFolders = Arrays.asList(
+                new ImmutablePair<>(pipeline1, Collections.singletonList(new UserPermission(TEST_USER_1, AclPermission.WRITE.getMask()))),
+                new ImmutablePair<>(pipeline1, Collections.singletonList(new UserPermission(TEST_USER_2, AclPermission.WRITE.getMask()))),
```
tcibinan: I believe that you should specify permissions for a single entity using a single pair because new `initAclEntity` method just overrides already specified object entity permissions.

As a result in this test you give WRITE permissions for the first pipeline to the first user but in the `expectedMap` you specify that the first user should have READ permissions for the first pipeline.

```java
new ImmutablePair<>(pipeline1, Arrays.asList(
    new UserPermission(TEST_USER_1, AclPermission.WRITE.getMask()), 
    new UserPermission(TEST_USER_2, AclPermission.WRITE.getMask())))
```

Please check for similar cases in other tests too.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
```diff
@@ -71,80 +67,264 @@
...
+
+        final PipelineWithPermissions pipelineWithPermissions2 = new PipelineWithPermissions();
+        pipelineWithPermissions2.setId(pipeline2.getId());
+        pipelineWithPermissions2.setOwner(pipeline2.getOwner());
+        doReturn(pipelineWithPermissions2).when(mockPipelineWithPermissionsMapper).toPipelineWithPermissions(eq(pipeline2));
+
+        final PipelineWithPermissions pipelineWithPermissions3 = new PipelineWithPermissions();
+        pipelineWithPermissions3.setId(pipeline2.getId());
+        pipelineWithPermissions3.setOwner(pipeline2.getOwner());
+        doReturn(pipelineWithPermissions3).when(mockPipelineWithPermissionsMapper).toPipelineWithPermissions(eq(pipeline2));
```
tcibinan: Mocking of `pipelineWithPermissions2` is overriden by mocking of `pipelineWithPermissions3`. I would suggest to get rid of  `pipelineWithPermissions3` declaration at all because it doesn't really have any effect on the test.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
```diff
@@ -71,80 +67,264 @@
...
+
+        final PipelineWithPermissions pipelineWithPermissions1 = new PipelineWithPermissions();
+        pipelineWithPermissions1.setId(pipeline1.getId());
+        pipelineWithPermissions1.setOwner(pipeline1.getOwner());
+        doReturn(pipelineWithPermissions1).when(mockPipelineWithPermissionsMapper).toPipelineWithPermissions(eq(pipeline1));
+
+        final PipelineWithPermissions pipelineWithPermissions2 = new PipelineWithPermissions();
+        pipelineWithPermissions2.setId(pipeline2.getId());
+        pipelineWithPermissions2.setOwner(pipeline2.getOwner());
+        doReturn(pipelineWithPermissions2).when(mockPipelineWithPermissionsMapper).toPipelineWithPermissions(eq(pipeline2));
```
tcibinan: Feels like this part of the tests can be extracted to `mockPipelineAndFolders` method because it is common mocking and it is pretty transparent.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
tcibinan: I meant that we should just move these declarations to the end of each test. Something like this 

```java
final Map<String, Integer> permissionsForPipeline1 = Collections.singletonMap(TEST_USER_1.toUpperCase(), 5);
final Map<String, Integer> permissionsForPipeline2 = Collections.singletonMap(TEST_USER_1.toUpperCase(), 1);
final Map<Long, Map<String, Integer>> expectedMap = new HashMap<>();
expectedMap.put(pipeline1.getId(), permissionsForPipeline1);
expectedMap.put(pipeline2.getId(), permissionsForPipeline2);
assertPipelineWithPermissions(expectedMap);
```

or even this if applicable

```java
final Map<Long, Map<String, Integer>> expectedMap = new HashMap<>();
expectedMap.put(pipeline1.getId(), Collections.singletonMap(TEST_USER_1.toUpperCase(), 5));
expectedMap.put(pipeline2.getId(), Collections.singletonMap(TEST_USER_1.toUpperCase(), 1));
assertPipelineWithPermissions(expectedMap);
```

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
```diff
@@ -71,80 +78,159 @@
...
+                        new UserPermission(TEST_USER_1, AclPermission.WRITE.getMask()),
+                        new UserPermission(TEST_USER_2, AclPermission.WRITE.getMask()))),
+                new ImmutablePair<>(pipeline2, Collections.singletonList(new UserPermission(TEST_USER_1, AclPermission.READ.getMask()))),
+                new ImmutablePair<>(folder1, Collections.singletonList(new UserPermission(TEST_USER_1, AclPermission.READ.getMask()))),
+                new ImmutablePair<>(folder2, Collections.singletonList(new UserPermission(TEST_USER_1, AclPermission.READ.getMask()))));
+
+        mockPipelineAndFolders(pipelinesAndFolders);
+
+        Arrays.asList(new UserPermission(TEST_USER_1, AclPermission.WRITE.getMask()),
+                new UserPermission(TEST_USER_2, AclPermission.WRITE.getMask()));
```
tcibinan: Unused piece of code.

api/src/test/java/com/epam/pipeline/test/acl/AbstractAclTest.java
```diff
@@ -107,8 +120,35 @@ protected void initAclEntity(AbstractSecuredEntity entity, List<AbstractGrantPer
...
+        pairs.forEach(p -> {
+            final AbstractSecuredEntity entity = p.getLeft();
+            final List<AbstractGrantPermission> permissions = p.getRight();
+            initAclEntity(entity, permissions);
+            ObjectIdentityImpl objectIdentity = new ObjectIdentityImpl(entity);
+            AclImpl acl = prepareAcl(objectIdentity, entity, permissions);
+            acls.put(objectIdentity, acl);
+        });
+        doReturn(acls).when(aclService).getObjectIdentities(anySet());
+    }
```
tcibinan: I would suggest getting rid of an additional `prepareAcl` method but just return `AclImpl` from an existing `initAclEntity` method. Once such refactoring is applied then we can simplify new method like this:

```suggestion
    protected void initAclEntity(List<Pair<AbstractSecuredEntity, List<AbstractGrantPermission>>> pairs) {
        final Map<ObjectIdentity, Acl> acls = pairs.stream()
            .map(p -> initAclEntity(p.getLeft(), p.getRight()))
            .collect(Collectors.toMap(AclImpl::getObjectIdentity, Function.identity());
        doReturn(acls).when(aclService).getObjectIdentities(anySet());
    }
```

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/test/acl/AbstractAclTest.java
tcibinan: See https://github.com/epam/cloud-pipeline/pull/1748#discussion_r567904583.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
```diff
@@ -71,80 +78,159 @@
...
+    public void shouldReturnPermissionsWhenGrantedToPipelineCase4() {
+        final List<Pair<AbstractSecuredEntity, List<AbstractGrantPermission>>> pipelinesAndFolders = Arrays.asList(
+                new ImmutablePair<>(pipeline1, Arrays.asList(
+                        new UserPermission(TEST_USER_1, AclPermission.WRITE.getMask()),
+                        new UserPermission(TEST_USER_2, AclPermission.WRITE.getMask()))),
+                new ImmutablePair<>(pipeline2, Collections.singletonList(new UserPermission(TEST_USER_1, AclPermission.READ.getMask()))),
+                new ImmutablePair<>(folder1, Arrays.asList(
+                        new UserPermission(TEST_USER_1, AclPermission.READ.getMask()),
+                        new UserPermission(TEST_USER_2, AclPermission.READ.getMask()),
+                        new UserPermission(TEST_USER_2, AclPermission.NO_READ.getMask()))),
```
tcibinan: Why do you set both `READ` and `NO_READ` permissions? According to [the original test](https://github.com/epam/cloud-pipeline/blob/develop/api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java#L138) you should grant `NO_READ` permission for the first folder to second user and then check that resulting mask equals to `2`.

Could you please double check that you are testing exact the same cases as the original test does in all test methods.

tcibinan (changes requested):

tcibinan (changes requested):
Could you please also execute the following gradle tasks to check if automated verification passes: pmdMain, pmdTest, checkstyleMain, checkstyleTest. You can do it using your IDE or using a command line like this.

```bash
# linux
./gradlew api:pmdMain api:pmdTest api:checkstyleMain api:checkstyleTest

# windows
gradlew.bat api:pmdMain api:pmdTest api:checkstyleMain api:checkstyleTest
```

Usually before submitting pull request I build an entire api module to be sure that my changes don't break anything. You can do it using one of the following commands. Unfortunately it takes about 10 minutes right now but the build time is constantly decreasing due to pull requests like this.

```bash
# linux
./gradlew api:build

# windows
gradlew.bat api:build
```

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineWithPermissionsTest.java
VolLol: In the first test it will work, but in the other tests it will be replaced existing map in the expected map and tests will not passed.

tcibinan (approved):

Issue #1750, opened 2021-01-26
Introduce mount timeout for EFS and Lustre


kamyshova:
**Background**

At the moment, Cloud Pipeline sensitive runs don't have access to the EFS and the FSx for Lustre filesystems. 
Thus the sensitive run hangs at startup initialization if an NFS storage is mounted. 

**Approach**

* Introduce a `mount_timeout` option to NFS mounting.

mzueva:
Timeout approach doesn't work for all cases, so it would be helpful to completely skip mounting of NFS storage to sensitive runs. Let's introduce a new SystemPreference `storage.mounts.nfs.sensitive.policy` with the following supported values:
- `SKIP` -  NFS mounts shall be skipped for sensitive runs
- `TIMEOUT` - NFS mounts shall be mounted with retry & timeo options for sensitive runs
- `NONE` or missing value - handle NFS mount in common way (mount with options provided for storage)

rodichenko:
@sidoruka, @NShaforostov  GUI part implemented, backported to `release/0.16` (66bba6f7bdc2854556fd38f6a3175acd9f82b7fb)

PR #1769, opened 2021-01-29
Introduce the new system preference for NFS mount of sensitive runs

kamyshova:
The addition to issue #1750
`storage.mounts.nfs.sensitive.policy` system preference was added for NFS mount of sensitive runs.
The preference supports `SKIP`, `TIMEOUT`, `NONE` values.
-   `SKIP` - NFS mounts shall be skipped for sensitive runs
-   `TIMEOUT` - NFS mounts shall be mounted with retry & timeo options for sensitive runs
-   `NONE` or missing value - handle NFS mount in common way (mount with options provided for storage)

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
+++ b/api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 EPAM Systems, Inc. (https://www.epam.com/)
+ * Copyright 2017-2021 EPAM Systems, Inc. (https://www.epam.com/)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/epam/pipeline/manager/preference/PreferenceValidators.java
```diff
@@ -255,6 +255,9 @@
     public static final BiPredicate<String, Map<String, Preference>> isValidLongPauseRunAction =
             (pref, ignored) -> LongPausedRunAction.contains(pref);
 
+    public static final BiPredicate<String, Map<String, Preference>> isValidNFSMountSensitivePolicy = (pref, ignored) ->
```
mzueva: We already have `isValidEnum` preference validator, I think we can use it here

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -512,6 +514,7 @@ def _get_credentials(self, storage):
 
 class NFSMounter(StorageMounter):
     available = False
+    sensitive_policy = PipelineAPI(os.environ['API'], None).get_preference(SENSITIVE_POLICY_PREFERENCE)['value']
```
mzueva: I think it is better to pass policy in the constructor args instead of initialising it for each mounter

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -592,7 +596,8 @@ def build_mount_command(self, params):
         return command
 
     def append_timeout_options(self, mount_options):
-        if self.share_mount.mount_type == 'SMB':
+        if self.share_mount.mount_type == 'SMB' or \
+                (PermissionHelper.is_run_sensitive() and NFSMounter.sensitive_policy != "TIMEOUT"):
```
mzueva: Condition should be `mount_type == SMB  or not PermissionHelper.is_run_sensitive() or NFSMounter.sensitive_policy != "TIMEOUT"` (do not apply timeout for non-sensitive runs)

mzueva (changes requested):

Issue #1771, opened 2021-02-01
Pipeline Runs shall not be deleted with parent Pipeline


mzueva:
**Background**
At the moment when `Pipeline` entity is deleted all child `PipelineRuns` are deleted as well. This leads to incorrect billing results, as deleted `PipelineRuns` are not included into reports and prevents server from clearing resources associated with `PipelineRun`, e.g. LustreFs and DNS records.

**Approach**
`PipelineRuns` shall not be deleted with parent `Pipeline` entity.

rodichenko:
@mzueva @sidoruka GUI part implemented (#1854)

rodichenko:
@mzueva @sidoruka @NShaforostov backported to `release/0.16` (2678f60bef5a55d6ae647724b0d3e4fa7fed8c69)

PR #1784, opened 2021-02-05
Issue #1771: Pipeline Runs shall not be deleted with parent Pipeline

ekazachkova:
The current PR provides implementation for issue #1771 

The following changes were added to DB:
 - removed `pipeline_id` constraint from `pipeline_run` table
 - added a new field `pipeline_name` to `pipeline_run` table


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineManager.java
```diff
@@ -184,6 +186,8 @@ public Pipeline update(final PipelineVO pipelineVO) {
         dbPipeline.setParentFolderId(pipelineVO.getParentFolderId());
         setFolderIfPresent(dbPipeline);
         pipelineDao.updatePipeline(dbPipeline);
+        ListUtils.emptyIfNull(pipelineRunDao.loadAllRunsForPipeline(pipelineVO.getId()))
```
mzueva: I'd suggest to use one SQL method to update pipelineName for all runs at once

mzueva (changes requested):
I think we are missing logic for setting `pipelineName` for runs table when a new run is created?

Issue #1755, opened 2021-01-27
Unit tests: Rewrite RunScheduleManagerTest


mzueva:
**Background**
At the moment `RunScheduleManagerTest` is annotated with `@DirtiesContext` and uses TestApplicationWithAclSecurity as context configuration. In order to reduce test startup time we need to remove `@DirtiesContext` annotation and make `RunScheduleManagerTest` a unit test without spring context (dao layer & other beans shall be replaced by mocks).

PR #1785, opened 2021-02-05
Unit tests: Rewrite RunScheduleManagerTest #1755

VolLol:
Remove @DirtiesContext and re-write test methods.

### DIFF
--- a/api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
+++ b/api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
@@ -257,7 +257,6 @@ public class RunScheduleManagerTest {
         runScheduleManager.deleteSchedulesForRunByPipeline(testPipeline.getId());
 
         verify(mockRunScheduler, times(2)).unscheduleRunSchedule(any());
-        verify(mockRunScheduleDao).deleteRunSchedulesForRunByPipeline(anyLong());
     }
 
     @Test(expected = IllegalArgumentException.class)
--- a/api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
+++ b/api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
@@ -74,8 +74,10 @@ public class RunScheduleManagerTest {
             PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION2, TIME_ZONE);
     private final PipelineRunScheduleVO testRunScheduleVO3 =
             PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.RUN, CRON_EXPRESSION2, TIME_ZONE);
-    private final RunSchedule runPauseSchedule = createRunSchedule(PAUSE_SCHEDULE_ID, RunScheduledAction.PAUSE, CRON_EXPRESSION1);
-    private final RunSchedule runResumeSchedule = createRunSchedule(RESUME_SCHEDULE_ID, RunScheduledAction.RESUME, CRON_EXPRESSION2);
+    private final RunSchedule runPauseSchedule =
+            createRunSchedule(PAUSE_SCHEDULE_ID, RunScheduledAction.PAUSE, CRON_EXPRESSION1);
+    private final RunSchedule runResumeSchedule =
+            createRunSchedule(RESUME_SCHEDULE_ID, RunScheduledAction.RESUME, CRON_EXPRESSION2);
 
     @InjectMocks
     private RunScheduleManager runScheduleManager;
@@ -227,7 +229,8 @@ public class RunScheduleManagerTest {
         doReturn(Optional.of(runResumeSchedule))
                 .when(mockRunScheduleDao).loadRunSchedule(RESUME_SCHEDULE_ID);
 
-        runScheduleManager.deleteSchedules(RUN_ID, ScheduleType.PIPELINE_RUN, Arrays.asList(PAUSE_SCHEDULE_ID, RESUME_SCHEDULE_ID));
+        runScheduleManager.deleteSchedules(
+                RUN_ID, ScheduleType.PIPELINE_RUN, Arrays.asList(PAUSE_SCHEDULE_ID, RESUME_SCHEDULE_ID));
 
         verify(mockRunScheduler, times(2)).unscheduleRunSchedule(any());
     }
--- a/api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
+++ b/api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
@@ -18,64 +18,64 @@ package com.epam.pipeline.manager.pipeline;
 
 import com.epam.pipeline.common.MessageHelper;
 import com.epam.pipeline.controller.vo.PipelineRunScheduleVO;
-import com.epam.pipeline.controller.vo.PipelineUserVO;
 import com.epam.pipeline.dao.pipeline.RunScheduleDao;
 import com.epam.pipeline.entity.configuration.RunConfiguration;
-import com.epam.pipeline.entity.configuration.RunConfigurationEntry;
 import com.epam.pipeline.entity.pipeline.Pipeline;
-import com.epam.pipeline.entity.pipeline.PipelineRun;
-import com.epam.pipeline.entity.pipeline.TaskStatus;
 import com.epam.pipeline.entity.pipeline.run.RunSchedule;
 import com.epam.pipeline.entity.pipeline.run.ScheduleType;
 import com.epam.pipeline.entity.pipeline.run.RunScheduledAction;
-import com.epam.pipeline.entity.region.AbstractCloudRegion;
 import com.epam.pipeline.entity.user.PipelineUser;
 import com.epam.pipeline.manager.ObjectCreatorUtils;
+import com.epam.pipeline.manager.configuration.RunConfigurationManager;
 import com.epam.pipeline.manager.scheduling.RunScheduler;
 import com.epam.pipeline.manager.scheduling.ScheduleProviderManager;
 import com.epam.pipeline.manager.scheduling.provider.PipelineRunScheduleProvider;
 import com.epam.pipeline.manager.scheduling.provider.RunConfigurationScheduleProvider;
 import com.epam.pipeline.manager.security.AuthManager;
+import com.epam.pipeline.test.creator.configuration.ConfigurationCreatorUtils;
+import com.epam.pipeline.test.creator.pipeline.PipelineCreatorUtils;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
-import org.mockito.Mockito;
 
 import static org.mockito.Matchers.any;
 
 import org.mockito.MockitoAnnotations;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
-import java.util.List;
 import java.util.Optional;
 import java.util.TimeZone;
-import java.util.stream.Collectors;
 
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Matchers.anyList;
+import static org.mockito.Matchers.anyListOf;
 import static org.mockito.Matchers.anyLong;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.when;
-import static org.mockito.Matchers.eq;
 
 public class RunScheduleManagerTest {
 
     private static final Long RUN_ID = 1L;
+    private static final Long PAUSE_SCHEDULE_ID = 1L;
+    private static final Long RESUME_SCHEDULE_ID = 2L;
     private static final String CRON_EXPRESSION1 = "0 0 12 * * ?";
     private static final String CRON_EXPRESSION2 = "0 15 10 ? * *";
     private static final String TEST_NAME = "TEST";
-    private static final String TEST_REPOSITORY = "///";
-    private static final TimeZone TIME_ZONE = TimeZone.getTimeZone("UTC");
-    private static final String TEST_PIPELINE_REPO_SSH = "git@test";
-    private static final String USER_OWNER = "OWNER";
-    private static final String ENTRY_NAME = "entry";
+    private static final String TIME_ZONE = "Coordinated Universal Time";
+
+    private final RunConfiguration runConfiguration = ConfigurationCreatorUtils.getRunConfiguration(2L, TEST_NAME);
+    private final Pipeline testPipeline = PipelineCreatorUtils.getPipeline(RUN_ID, TEST_NAME);
+    private final PipelineRunScheduleVO testRunScheduleVO =
+            PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.PAUSE, CRON_EXPRESSION1, TIME_ZONE);
+    private final PipelineRunScheduleVO testRunScheduleVO2 =
+            PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION2, TIME_ZONE);
+    private final PipelineRunScheduleVO testRunScheduleVO3 =
+            PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.RUN, CRON_EXPRESSION2, TIME_ZONE);
+    private final RunSchedule runPauseSchedule = createRunSchedule(PAUSE_SCHEDULE_ID, RunScheduledAction.PAUSE, CRON_EXPRESSION1);
+    private final RunSchedule runResumeSchedule = createRunSchedule(RESUME_SCHEDULE_ID, RunScheduledAction.RESUME, CRON_EXPRESSION2);
 
     @InjectMocks
     private RunScheduleManager runScheduleManager;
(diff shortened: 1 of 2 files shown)


api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,179 +121,242 @@ public void setUp() {
...
     @Test
-    @WithMockUser(username = USER_OWNER)
     public void testCreateRunSchedules() {
-        final List<RunSchedule> runSchedules = runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
-                    Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
-        runSchedules.forEach(this::loadAndAssertSchedule);
+        when(mockRunScheduleDao.loadAllRunSchedules()).thenReturn(Collections.emptyList());
+        when(mockScheduleProviderManager.getProvider(ScheduleType.PIPELINE_RUN)).thenReturn(mockPipelineRunScheduleProvider);
+        when(mockScheduleProviderManager.getProvider(ScheduleType.RUN_CONFIGURATION)).thenReturn(mockRunConfigurationScheduleProvider);
+        when(mockAuthManager.getCurrentUser()).thenReturn(mockPipelineUser);
```
tcibinan: I would say that `mockScheduleProviderManager` and `mockAuthManager` mocking can be extracted to `setUp` method because it's the same for all the tests.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,179 +121,242 @@ public void setUp() {
...
         final PipelineRunScheduleVO runScheduleVO2 = getRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION1);
         runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
                 Arrays.asList(runScheduleVO, runScheduleVO2));
     }
 
     @Test(expected = IllegalArgumentException.class)
-    @WithMockUser(username = USER_OWNER)
-    public void  createSchedulesWithWrongAction() {
+    public void createSchedulesWithWrongAction() {
+        doThrow(IllegalArgumentException.class).when(mockScheduleProviderManager).getProvider(ScheduleType.RUN_CONFIGURATION);
```
tcibinan: The test is a bit confusing because it is not clear how does a wrong action and `getProvider` throwing `IllegalArgumentException` are connected. I think that `mockScheduleProviderManager.verifyScheduleAction` should throw the exception instead.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,179 +121,242 @@ public void setUp() {
...
+
     @Test(expected = IllegalArgumentException.class)
-    @WithMockUser(username = USER_OWNER)
     public void testCreateRunScheduleWithExistentCronExpression() {
+        when(mockScheduleProviderManager.getProvider(ScheduleType.PIPELINE_RUN)).thenReturn(mockPipelineRunScheduleProvider);
+        when(mockAuthManager.getCurrentUser()).thenReturn(mockPipelineUser);
+
         runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
                 Collections.singletonList(testRunScheduleVO));
         testRunScheduleVO.setAction(RunScheduledAction.RESUME);
```
tcibinan: If you delete these lines the test still passes. You see, the problem here is that previously the test used real data base and some additional actions were required. As long as we mock everything now we can skip most of the preconfiguration and just test exactly what we want to verify.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,179 +121,242 @@ public void setUp() {
...
+
+        doReturn(Collections.singletonList(createRunSchedule())).
```
tcibinan: Could you please move the dot to the next line.

```java
// bad
doReturn().
    when();

// good
doReturn
    .when();
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,179 +121,242 @@ public void setUp() {
...
         testRunScheduleVO = getRunScheduleVO(RunScheduledAction.PAUSE, CRON_EXPRESSION1);
         testRunScheduleVO2 = getRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION2);
         testRunScheduleVO3 = getRunScheduleVO(RunScheduledAction.RUN, CRON_EXPRESSION2);
 
+        MockitoAnnotations.initMocks(this);
     }
 
     @Test
-    @WithMockUser(username = USER_OWNER)
     public void testCreateRunSchedules() {
```
tcibinan: Could you please break the test into two tests: `testCreatePipelineRunSchedules` and `testCreateRunConfigurationSchedules`.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,179 +121,242 @@ public void setUp() {
...
+        when(mockAuthManager.getCurrentUser()).thenReturn(mockPipelineUser);
+
         final List<RunSchedule> schedules =
-            runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
-                    Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
+                runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
+                        Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
+
+        schedules.get(0).setId(1L);
+        schedules.get(1).setId(2L);
```
tcibinan: Again, as long as we write a unit test  then there is no actual need to create schedules first using `runScheduleManager`. Just mock everything you need and then test only a single method - `updateSchedules`.

Same for all update and delete tests in the class.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
```
tcibinan: Could you please add missing `final` keywords and also replace mocking by the following pattern. It tends to be more common syntax and we try to use it in all the tests right now.

```
# bad
when().doSomething().thenReturn();

# good
doReturn().when().doSomething();
```

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +115,240 @@ public void setUp() {
...
         final PipelineRunScheduleVO runScheduleVO2 = getRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION1);
         runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
                 Arrays.asList(runScheduleVO, runScheduleVO2));
     }
 
     @Test(expected = IllegalArgumentException.class)
-    @WithMockUser(username = USER_OWNER)
-    public void  createSchedulesWithWrongAction() {
+    public void createSchedulesWithWrongAction() {
+        doThrow(IllegalArgumentException.class).when(mockScheduleProviderManager).getProvider(ScheduleType.RUN_CONFIGURATION);
```
tcibinan: Please remove this mocking because you already have throwing one.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +115,240 @@ public void setUp() {
...
+                .when(mockRunScheduleDao).loadAllRunSchedulesBySchedulableIdAndType(RUN_ID, ScheduleType.PIPELINE_RUN);
+
+        runScheduleManager.updateSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
+                Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
+
+        verify(mockRunScheduler, times(2)).scheduleRunSchedule(any());
+        verify(mockRunScheduler, times(2)).unscheduleRunSchedule(any());
+        verify(mockRunScheduleDao, times(1)).updateRunSchedules(anyList());
+        verify(mockRunScheduleDao, times(1)).loadAllRunSchedulesBySchedulableIdAndType(anyLong(), eq(ScheduleType.PIPELINE_RUN));
+        verify(mockScheduleProviderManager, times(3)).getProvider(ScheduleType.PIPELINE_RUN);
```
tcibinan: I suppose we can just check only the core calls skipping all the rest.

```java
verify(mockRunScheduler, times(2)).scheduleRunSchedule(any());
verify(mockRunScheduler, times(2)).unscheduleRunSchedule(any());
verify(mockRunScheduleDao).updateRunSchedules(anyListOf(RunSchedule.class));
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +115,240 @@ public void setUp() {
...
-        final List<RunSchedule> runSchedules = runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
-                    Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
-        runSchedules.forEach(this::loadAndAssertSchedule);
+    public void testCreatePipelineRunSchedules() {
+        runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
+                Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
+
+        verify(mockRunScheduler, times(2)).scheduleRunSchedule(any());
+        verify(mockRunScheduleDao, times(1)).createRunSchedules(anyList());
+        verify(mockScheduleProviderManager, times(3)).getProvider(ScheduleType.PIPELINE_RUN);
```
tcibinan: I suppose we can skip this verification in all tests. It doesn't really check anything specific and strongly depends on the implementation details.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +115,240 @@ public void setUp() {
...
-        final List<RunSchedule> loadRunSchedule = runScheduleManager
-                .loadAllSchedulesBySchedulableId(RUN_ID, ScheduleType.PIPELINE_RUN);
-        assertEquals(0, loadRunSchedule.size());
-        Mockito.verify(runScheduler, Mockito.times(2)).unscheduleRunSchedule(any());
+        runScheduleManager.deleteSchedulesForRunByPipeline(testPipeline.getId());
 
+        verify(mockPipelineRunManager, times(1)).loadAllRunsByPipeline(anyLong());
+        verify(mockRunScheduleDao, times(1)).loadAllRunSchedulesBySchedulableIdAndType(anyLong(), eq(ScheduleType.PIPELINE_RUN));
+        verify(mockRunScheduler, times(2)).unscheduleRunSchedule(any());
+        verify(mockRunScheduleDao, times(1)).deleteRunSchedulesForRunByPipeline(anyLong());
```
tcibinan: Let's simplify the verification to just the following

```java
verify(mockRunScheduler, times(2)).unscheduleRunSchedule(any());
verify(mockRunScheduleDao).deleteRunSchedulesForRunByPipeline(anyLong());
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +115,240 @@ public void setUp() {
...
     @Test(expected = IllegalArgumentException.class)
-    @WithMockUser(username = USER_OWNER)
     public void loadRunScheduleByWrongId() {
-        final RunSchedule runSchedule =
-            runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
-                    Collections.singletonList(testRunScheduleVO)).get(0);
-        runScheduleManager.loadSchedule(runSchedule.getSchedulableId());
+        doThrow(IllegalArgumentException.class).when(mockRunScheduleDao).loadRunSchedule(1L);
+
+        runScheduleManager.loadSchedule(1L);
```
tcibinan: You can see that according to `mockRunScheduleDao::loadRunSchedule` implementation it doesn't throw errors on missing run schedules but rather return an empty optional.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
tcibinan: Could you please remove similar verifications from all the tests?

```
verify(mockScheduleProviderManager, times(3)).getProvider(ScheduleType.PIPELINE_RUN);
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -68,27 +69,40 @@
...
 
-    @Autowired
-    private RunConfigurationDao configurationDao;
+    @Mock
+    private PipelineUser mockPipelineUser;
 
-    @Autowired
-    private UserManager userManager;
+    @InjectMocks
+    private RunConfigurationScheduleProvider mockRunConfigurationScheduleProvider;
```
tcibinan, line 104: No need to use spy here. Let's use a regular mock for the class.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
tcibinan: The only actual mocking in the test should be the following.

```
        doThrow(AssertionError.class).when(mockRunConfigurationScheduleProvider)
                .verifyScheduleAction(RUN_ID, eq(any()));
```

tcibinan: The actual mocking should be something like this

```java
doThrow(IllegalArgumentException.class).when(mockRunConfigurationScheduleProvider)
        .verifyScheduleAction(RUN_ID, RunScheduledAction.RESUME);
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
-        assertEquals(schedule.getCronExpression(), loadRunSchedule.getCronExpression());
-        assertEquals(schedule.getAction(), loadRunSchedule.getAction());
-        assertEquals(schedule.getUser(), loadRunSchedule.getUser());
+    private RunSchedule createRunSchedule(final Long id, final Long schedulableId,
+                                          final RunScheduledAction action, final String cronExpression) {
+        final RunSchedule runSchedule = new RunSchedule();
+        runSchedule.setUser(USER_OWNER);
+        runSchedule.setAction(action);
+        runSchedule.setId(id);
+        runSchedule.setSchedulableId(schedulableId);
```
tcibinan: You can inline `RUN_ID` here removing extra method parameter `schedulableId`.

```
runSchedule.setSchedulableId(RUN_ID);
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
-        runSchedules.forEach(this::loadAndAssertSchedule);
+    public void testCreatePipelineRunSchedules() {
+        runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
+                Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
 
-        List<RunSchedule> configSchedules = runScheduleManager.createSchedules(runConfiguration.getId(),
-                ScheduleType.RUN_CONFIGURATION, Collections.singletonList(testRunScheduleVO3));
-        configSchedules.forEach(this::loadAndAssertSchedule);
+        verify(mockRunScheduler, times(2)).scheduleRunSchedule(any());
+        verify(mockRunScheduleDao, times(1)).createRunSchedules(anyList());
```
tcibinan: Could you please use typed version of `anyList` matcher called `anyListOf` like this:

```java
verify(mockRunScheduleDao).createRunSchedules(anyListOf(RunSchedule.class));
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
+    @Test
+    public void testCreateRunConfigurationSchedules() {
+        doReturn(runConfiguration).when(mockRunConfigurationManager).load(anyLong());
 
-        Mockito.verify(runScheduler, Mockito.times(3)).scheduleRunSchedule(any());
+        runScheduleManager.createSchedules(runConfiguration.getId(),
+                ScheduleType.RUN_CONFIGURATION, Collections.singletonList(testRunScheduleVO3));
 
+        verify(mockRunScheduler, times(1)).scheduleRunSchedule(any());
+        verify(mockRunScheduleDao, times(1)).createRunSchedules(anyList());
```
tcibinan: It's a nice idea to skip extra `times(1)` argument because `verify(something, times(1))` is equivalent to `verify(something)`. You can check the verify method's implementation to be sure about that.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
 
-        RunConfiguration configuration =
+        final RunConfiguration configuration =
                 ObjectCreatorUtils.createConfiguration(TEST_NAME, null, null,
                         USER_OWNER, Collections.singletonList(entry));
-        return configurationDao.create(configuration);
+        return configuration;
     }
 
+    private RunSchedule createRunSchedule() {
```
tcibinan: It seems like we can get rid of this method in advance to the overloaded version of the same method with several parameters.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
     }
 
     @Test(expected = IllegalArgumentException.class)
-    @WithMockUser(username = USER_OWNER)
     public void testUpdateRunSchedulesFailsWithoutScheduleId() {
         runScheduleManager.createSchedules(RUN_ID, ScheduleType.PIPELINE_RUN,
-                        Collections.singletonList(testRunScheduleVO));
+                Collections.singletonList(testRunScheduleVO));
 
         testRunScheduleVO.setCronExpression(CRON_EXPRESSION2);
```
tcibinan: Please do not try to create schedules before updating them. Please remember that all the dependent classes are mocks now so calling `createSchedules` basically does nothing.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
         testPipeline.setRepository(TEST_REPOSITORY);
         testPipeline.setOwner(TEST_NAME);
         testPipeline.setRepositorySsh(TEST_PIPELINE_REPO_SSH);
-        pipelineDao.createPipeline(testPipeline);
+        testPipeline.setId(RUN_ID);
 
         createPipelineRun(RUN_ID, testPipeline.getId());
```
tcibinan: You can get rid of this call. It doesn't have any effect now because everything is mock.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
+
         runConfiguration = createRunConfiguration();
+        runConfiguration.setId(2L);
 
         PipelineUserVO userVO = new PipelineUserVO();
         userVO.setUserName(USER_OWNER);
```
tcibinan: Same here. Just remove the unused code please.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -103,214 +117,238 @@ public void setUp() {
...
+
+        mockRunConfigurationScheduleProvider =
+                Mockito.spy(new RunConfigurationScheduleProvider(mockMessageHelper, mockRunConfigurationManager));
+
+        doReturn(mockPipelineRunScheduleProvider)
+                .when(mockScheduleProviderManager).getProvider(ScheduleType.PIPELINE_RUN);
+        doReturn(mockRunConfigurationScheduleProvider)
+                .when(mockScheduleProviderManager).getProvider(ScheduleType.RUN_CONFIGURATION);
+        doReturn(mockPipelineUser).when(mockAuthManager).getCurrentUser();
     }
```
tcibinan, line 135: The whole method can be greatly simplified by using existing `CreatorUtils` classes. Additionally it would be nice to replace existing private creator methods with the usage of existing `CreatorUtils` methods. Of course missing methods can be implemented.

```java 
private final RunConfiguration runConfiguration = ConfigurationCreatorUtils.getRunConfiguration(2L, TEST_NAME);
private final Pipeline testPipeline = PipelineCreatorUtils.getPipeline(RUN_ID, TEST_NAME);
private final PipelineRunScheduleVO testRunScheduleVO = getRunScheduleVO(RunScheduledAction.PAUSE, CRON_EXPRESSION1);
private final PipelineRunScheduleVO testRunScheduleVO2 = getRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION2);
private final PipelineRunScheduleVO testRunScheduleVO3 = getRunScheduleVO(RunScheduledAction.RUN, CRON_EXPRESSION2);

@Before
public void setUp() {
    MockitoAnnotations.initMocks(this);
    doReturn(mockPipelineRunScheduleProvider)
            .when(mockScheduleProviderManager).getProvider(ScheduleType.PIPELINE_RUN);
    doReturn(mockRunConfigurationScheduleProvider)
            .when(mockScheduleProviderManager).getProvider(ScheduleType.RUN_CONFIGURATION);
    doReturn(mockPipelineUser).when(mockAuthManager).getCurrentUser();
}
```

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
tcibinan: Do you use Intellij IDEA? It seems like such an issue should be highlighted.
image

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -16,301 +16,282 @@
...
+
+    private PipelineRun createPipelineRun(final Long runId, final Long pipelineId) {
         final AbstractCloudRegion cloudRegion = ObjectCreatorUtils.getDefaultAwsRegion();
-        regionDao.create(cloudRegion);
         PipelineRun run = ObjectCreatorUtils.createPipelineRun(runId, pipelineId, null, cloudRegion.getId());
         run.setStatus(TaskStatus.RUNNING);
         run.getInstance().setSpot(false);
-        pipelineRunDao.createPipelineRun(run);
+        return run;
     }
```
tcibinan: As long as there is only a single call of the method let's replace the method with a single class field. I think something like this should be enough.

```java
private final PipelineRun pipelineRun = ObjectCreatorUtils.createPipelineRun(runId, pipelineId, null, CommonCreatorConstants.ID);
```

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -16,301 +16,282 @@
...
-    private RunConfiguration createRunConfiguration() {
-        //create
-        RunConfigurationEntry entry =
-                ObjectCreatorUtils.createConfigEntry("entry", true, null);
-
-        RunConfiguration configuration =
-                ObjectCreatorUtils.createConfiguration(TEST_NAME, null, null,
-                        USER_OWNER, Collections.singletonList(entry));
-        return configurationDao.create(configuration);
+    private RunSchedule createRunSchedule(final Long id, final Long schedulableId,
```
tcibinan: It seems like there is only two repeating invocations of the method. Can we extract two class fields out of this and use them in all the tests?

image

Something like

```java
private final RunSchedule runPauseSchedule = ...
private final RunSchedule runResumeSchedule = ...

File: api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java [hunk 24]
```diff
@@ -16,301 +16,282 @@
...
-                Arrays.asList(testRunScheduleVO, testRunScheduleVO2));
-        Mockito.verify(runScheduler, Mockito.times(2)).scheduleRunSchedule(any());
+        doReturn(
+                Arrays.asList(
+                        createRunSchedule(1L, RUN_ID, RunScheduledAction.PAUSE, CRON_EXPRESSION1),
+                        createRunSchedule(2L, RUN_ID, RunScheduledAction.RESUME, CRON_EXPRESSION2)))
+                .when(mockRunScheduleDao).loadAllRunSchedulesBySchedulableIdAndType(RUN_ID, ScheduleType.PIPELINE_RUN);
 
-        runScheduleManager.deleteSchedulesForRunByPipeline(testPipeline.getId());
+        doReturn(Arrays.asList(createPipelineRun(RUN_ID, testPipeline.getId())))
```
tcibinan: It's a nice idea to use `Collections.singletonList` rather than `Arrays.asList` for singleton lists.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
tcibinan: I meant that we could just remove whole method `createPipelineRun` and replace its usage with `ObjectCreatorUtils.createPipelineRun`. It should work I think.

api/src/test/java/com/epam/pipeline/manager/pipeline/RunScheduleManagerTest.java
```diff
@@ -16,301 +16,276 @@
...
+    private final RunConfiguration runConfiguration = ConfigurationCreatorUtils.getRunConfiguration(2L, TEST_NAME);
+    private final Pipeline testPipeline = PipelineCreatorUtils.getPipeline(RUN_ID, TEST_NAME);
+    private final PipelineRunScheduleVO testRunScheduleVO =
+            PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.PAUSE, CRON_EXPRESSION1, TIME_ZONE);
+    private final PipelineRunScheduleVO testRunScheduleVO2 =
+            PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.RESUME, CRON_EXPRESSION2, TIME_ZONE);
+    private final PipelineRunScheduleVO testRunScheduleVO3 =
+            PipelineCreatorUtils.getPipelineRunScheduleVO(RunScheduledAction.RUN, CRON_EXPRESSION2, TIME_ZONE);
+    private final RunSchedule runPauseSchedule = createRunSchedule(RUN_ID, RunScheduledAction.PAUSE, CRON_EXPRESSION1);
+    private final RunSchedule runResumeSchedule = createRunSchedule(2L, RunScheduledAction.RESUME, CRON_EXPRESSION2);
```
tcibinan: It is very confusing to see `RUN_ID` used as run schedule id. Let's add two more constants for run schedule ids and use it in the class rather than using `1L` and `2L` values.

```java
private static final Long RUN_ID = 1L;
private static final Long PAUSE_SCHEDULE_ID = 1L;
private static final Long RESUME_SCHEDULE_ID = 2L;

private final RunSchedule runPauseSchedule = createRunSchedule(PAUSE_SCHEDULE_ID, RunScheduledAction.PAUSE, CRON_EXPRESSION1);
private final RunSchedule runResumeSchedule = createRunSchedule(RESUME_SCHEDULE_ID, RunScheduledAction.RESUME, CRON_EXPRESSION2);

...

testRunScheduleVO.setScheduleId(PAUSE_SCHEDULE_ID);

...

doReturn(Optional.of(runPauseSchedule))
        .when(mockRunScheduleDao).loadRunSchedule(PAUSE_SCHEDULE_ID);

# and etc
```

tcibinan (changes requested):

tcibinan (approved):

Issue #1776, opened 2021-02-03
Unit tests: Refactor PipelineRunManagerTest


mzueva:
**Background**
At the moment `PipelineRunManagerTest` is annotated with `@DirtiesContext` and uses `TestApplicationWithAclSecurity` as context configuration. In order to reduce test startup time we need to remove `@DirtiesContext` annotation and split test into minor ones unit or integration tests.

**Approach**
- Extract all `instanceType` and `priceType` related tests to a separate unit test
- Extract all `region` related tests to a separate unit test

mzueva:
Further refactoring:
- for now leave tests `testRunCmdFailed` and `testAdminRunForce` in `PipelineRunManagerTest` these tests shall be rewritten when test context with enabled aspects is ready
- move test `testEnvVarsReplacement` to `PipelineRunManagerUnitTest` and split it into separate tests by cases
- move tests `testResolveProjectFiltering`, `shouldThrowExceptionOnInexistentRunTagUpdate` to `PipelineRunManagerUnitTest`
- remove test `testUpdateRunTags`
- move test `testLoadRunsActivityStats` to `PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest` (let's remove this test into `PipelineRunManagerLaunchTest`
- rename all refactored tests using `shouldDoSmth` notation

PR #1742, opened 2021-01-21
Issue-1735: Unit tests: Rewrite FolderTemplateManagerTest.

Sergeev-Ilya:
Issue-1735: Refactored FolderTemplateManagerTest. Removed @DirtiesContextAnnotation, inherited class from AbstractAclTest, reduced configuration.

Link to issue: #1735

PR #1793, opened 2021-02-08
Issue 1776: Unit Tests: Refactor PipelineRunManagerTest

Sergeev-Ilya:
Issue: https://github.com/epam/cloud-pipeline/issues/1776


api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java
```diff
@@ -91,12 +86,11 @@
...
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-@DirtiesContext //TODO: find a better workaround, this may make tests slower. Maybe, create a special package for
-                // integration tests, so they will be executed one after another and the context will remain?
+// TODO: find a better workaround, this may make tests slower. Maybe, create a special package for
+// integration tests, so they will be executed one after another and the context will remain?
```
tcibinan: Let's remove the comment along with the annotation because we've resolved the corresponding todo.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,707 @@
...
+        doReturn(DEFAULT_COMMAND).when(pipelineLauncher).launch(
+                any(PipelineRun.class), any(), any(), anyString(), anyString());
+        doReturn(ToolVersion.builder().size(1L).build())
+                .when(toolVersionManager).loadToolVersion(anyLong(), anyString());
+        doReturn(true).when(permissionHelper).isAllowed(any(), any());
+        doReturn(TEST_USER).when(securityManager).getAuthorizedUser();
+    }
+
+    @Test
+    public void shouldLaunchPipelineValidatesToolInstanceTypeAndPriceType() {
```
tcibinan: The idea to have long test case names is to be descriptive. For this test and all the others I would propose using names which are almost english sentences like `launchToolShouldValidateToolInstanceTypeAndPriceType`.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,707 @@
...
+        verify(toolManager).getCurrentImageSize(eq(TEST_IMAGE));
+        verifyPreferenceManager(1);
+
+        verify(instanceOfferManager).getInstanceEstimatedPrice(eq(null), eq(1), eq(true), eq(ID));
+        verify(securityManager).getAuthorizedUser();
+        verify(pipelineLauncher).launch(argThat(matches(Predicates.forPipelineRun())),
+                argThat(matches(Predicates.forConfiguration())),
+                eq(null), eq("0"), eq(null));
+
+        verify(dataStorageManager).analyzePipelineRunsParameters(anyListOf(PipelineRun.class));
```
tcibinan: It seems to be a little misunderstanding after the review in #1742.

I asked you to add a lot of verification to the test because the original test was testing all kind of stuff and we wanted to keep the original test behavior as much as possible. At the same time it doesn't mean that it is a good practice and definitely `FolderTemplateManagerTest` class is not a good example of a test because it has only a single god-like method that tests everything at once.

Regarding the tests in `PipelineRunManagerTest` that you moved here I would say that they are much better tests because they tested a single specification at once. If we write new or update similar tests it's a good idea to keep them in that notation: _a single test verifies a single aspect of a testing behavior_. 

In specific I would simplify this test to just the following two tests as it was originally.

```java
    @Test
    public void launchPipelineShouldValidatePriceType() {
        launchPipeline(configuration, INSTANCE_TYPE);

        verify(instanceOfferManager).isPriceTypeAllowed(eq(SPOT), any(), anyBoolean());
    }

    @Test
    public void launchPipelineShouldValidatePipelineInstanceType() {
        launchPipeline(configuration, INSTANCE_TYPE);

        verify(instanceOfferManager).isInstanceAllowed(eq(INSTANCE_TYPE), eq(ID), eq(true));
    }
```

Could I ask you to redesign tests in this class in the same manner according to the original tests?

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,707 @@
...
+        mockToolManager();
+
+        doReturn(parentRun).when(pipelineRunDao).loadPipelineRun(eq(PARENT_RUN_ID));
+        doReturn(false).when(permissionManager).isRunSshAllowed(eq(parentRun));
+        doReturn(DEFAULT_COMMAND).when(pipelineLauncher).launch(
+                any(PipelineRun.class), any(), any(), anyString(), anyString());
+        doReturn(ToolVersion.builder().size(1L).build())
+                .when(toolVersionManager).loadToolVersion(anyLong(), anyString());
+        doReturn(true).when(permissionHelper).isAllowed(any(), any());
+        doReturn(TEST_USER).when(securityManager).getAuthorizedUser();
```
tcibinan: It feels like a big part of this mocking can be omitted or at least inlined to specific tests. For example there is no actual need to mock `securityManager.getAuthorizedUser` because there are no tests that are actual testing this functionality.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,707 @@
...
+
+        tool = getTool(TEST_IMAGE, DEFAULT_COMMAND);
+        defaultAwsRegion = getDefaultAwsRegion(ID);
+        nonAllowedAwsRegion = getNonDefaultAwsRegion(ID_2);
+        notDefaultAwsRegion = getNonDefaultAwsRegion(ID_3);
+        configuration = getPipelineConfiguration(TEST_IMAGE, INSTANCE_DISK, true, defaultAwsRegion.getId());
+        configurationWithoutRegion = getPipelineConfiguration(TEST_IMAGE, INSTANCE_DISK);
+        price = new InstancePrice(configuration.getInstanceType(), parseInt(configuration.getInstanceDisk()),
+                PRICE_PER_HOUR, COMPUTE_PRICE_PER_HOUR, DISK_PRICE_PER_HOUR);
+        parentRun = PipelineCreatorUtils.getPipelineRunWithInstance(PARENT_RUN_ID, TEST_USER, ID_3);
```
tcibinan: Could you please join field declarations and assignments?

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Omitted 3 mocks. Also some mocks are grouped and replaced to a special method. (mockParentRun())

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: All tests are significantly redesigned.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java
Sergeev-Ilya: Done!

Sergeev-Ilya:
Solved all the comments!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    @Test
+    public void launchPipelineShouldValidateToolInstanceTypeAndPriceType() {
+        launchTool(configuration, INSTANCE_TYPE);
+
+        verify(toolManager).loadByNameOrId(eq(IMAGE));
+        verify(instanceOfferManager).isToolInstanceAllowed(eq(INSTANCE_TYPE), eq(getResource()), eq(ID), eq(true));
+        verify(instanceOfferManager).isPriceTypeAllowed(eq(SPOT), eq(getResource()), eq(false));
+    }
```
tcibinan: Could you please split this test into two according to the original tests: [first](https://github.com/epam/cloud-pipeline/blob/9ce7c2f068113bedf418bde38bbc8843382caddc/api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java#L365-L371) and [second](https://github.com/epam/cloud-pipeline/blob/9ce7c2f068113bedf418bde38bbc8843382caddc/api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java#L441-L447)?

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+
+        verify(pipelineRunDao).loadPipelineRun(eq(PARENT_RUN_ID));
+        verify(permissionManager).isRunSshAllowed(eq(parentRun));
+        verify(cloudRegionManager).load(eq(ID_3));
+    }
+
+    private void mockToolManager() {
+        doReturn(tool).when(toolManager).loadByNameOrId(eq(IMAGE));
+        doReturn(tool).when(toolManager).resolveSymlinks(eq(IMAGE));
+        doReturn(Optional.empty()).when(toolManager).loadToolVersionScan(eq(tool.getId()), eq(null));
```
tcibinan: There is no such call in `PipelineRunManager` so we can skip it.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+                .when(preferenceManager).getPreference(eq(CLUSTER_INSTANCE_HDD_EXTRA_MULTI));
+        doReturn(CLUSTER_SPOT.getDefaultValue()).when(preferenceManager).getPreference(eq(CLUSTER_SPOT));
+        doReturn(COMMIT_TIMEOUT.getDefaultValue()).when(preferenceManager).getPreference(eq(COMMIT_TIMEOUT));
+    }
+
+    private void mockInstanceOfferManager() {
+        doReturn(true).when(instanceOfferManager).isInstanceAllowed(anyString(), eq(ID), eq(true));
+        doReturn(true).when(instanceOfferManager).isInstanceAllowed(anyString(), eq(ID), eq(false));
+        doReturn(true).when(instanceOfferManager).isToolInstanceAllowed(anyString(), any(), eq(ID), eq(true));
+        doReturn(true).when(instanceOfferManager).isToolInstanceAllowed(anyString(), any(), eq(ID), eq(false));
```
tcibinan: Just
```
doReturn(true).when(instanceOfferManager).isInstanceAllowed(anyString(), eq(ID), anyBoolean());
doReturn(true).when(instanceOfferManager).isToolInstanceAllowed(anyString(), any(), eq(ID), anyBoolean());
```

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+        doReturn(false).when(instanceOfferManager).isInstanceAllowed(anyString(), eq(ID_2), eq(true));
+        doReturn(true).when(instanceOfferManager).isInstanceAllowed(anyString(), eq(ID_3), eq(false));
+        doReturn(true).when(instanceOfferManager).isPriceTypeAllowed(anyString(), any(), anyBoolean());
+        doReturn(BehaviorSubject.create()).when(instanceOfferManager).getAllInstanceTypesObservable();
```
tcibinan: Another unnecessary mock.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+        doReturn(price).when(instanceOfferManager)
+                .getInstanceEstimatedPrice(anyString(), anyInt(), anyBoolean(), anyLong());
+    }
+
+    private void mockParentRun() {
+        doReturn(parentRun).when(pipelineRunDao).loadPipelineRun(eq(PARENT_RUN_ID));
+        doReturn(false).when(permissionManager).isRunSshAllowed(eq(parentRun));
```
tcibinan: This mock can be skipped as well.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+        doReturn(toolConfiguration).when(pipelineConfigurationManager)
+                .getConfigurationForTool(eq(tool), eq(configuration));
+        doReturn(NOT_ALLOWED).when(messageHelper).getMessage(eq(ERROR_TOOL_CLOUD_REGION_NOT_ALLOWED), any(), any());
+
+        assertThrows(e -> e.getMessage().contains(NOT_ALLOWED), () -> launchPipeline(configuration, null));
+        assertThrows(e -> e.getMessage().contains(NOT_ALLOWED), () -> launchTool(configuration, null));
+
+        verify(cloudRegionManager, times(2)).load(eq(ID));
+        verify(permissionHelper, times(2)).isAdmin();
+        verify(cloudRegionManager, times(2)).load(eq(ID_3));
```
tcibinan: I suppose such verification calls can be skipped because they don't really check anything related to the test description.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+    @Test
+    public void launchPipelineShouldFailIfToolCloudRegionIsConfiguredAndItDiffersFromRunConfigurationOne() {
+        final PipelineConfiguration toolConfiguration = new PipelineConfiguration();
+        toolConfiguration.setCloudRegionId(ID_3);
+        doReturn(toolConfiguration).when(pipelineConfigurationManager)
+                .getConfigurationForTool(eq(tool), eq(configuration));
+        doReturn(NOT_ALLOWED).when(messageHelper).getMessage(eq(ERROR_TOOL_CLOUD_REGION_NOT_ALLOWED), any(), any());
+
+        assertThrows(e -> e.getMessage().contains(NOT_ALLOWED), () -> launchPipeline(configuration, null));
+        assertThrows(e -> e.getMessage().contains(NOT_ALLOWED), () -> launchTool(configuration, null));
```
tcibinan: Let's skip mocking and checking the error message. It made sense in the original test but now it doesn't.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+
+    @Test
+    public void launchPipelineShouldFailOnNotAllowedPriceType() {
+        doReturn(false).when(instanceOfferManager).isPriceTypeAllowed(eq(SPOT), eq(null), eq(false));
+        doReturn(NOT_ALLOWED).when(messageHelper).getMessage(eq(ERROR_PRICE_TYPE_IS_NOT_ALLOWED), eq(PriceType.SPOT));
+
+        final Runnable result = () -> launchPipeline(configuration, INSTANCE_TYPE);
+        assertThrows(e -> e.getMessage().contains(NOT_ALLOWED), result);
+
+        verify(instanceOfferManager).isInstanceAllowed(eq(INSTANCE_TYPE), eq(ID), eq(true));
```
tcibinan: We test price types here so let's omit non price type checks.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
```diff
@@ -0,0 +1,428 @@
...
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@SuppressWarnings("PMD.UnusedPrivateField")
+public class PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest {
```
tcibinan: What do you think about renaming the test class to `PipelineRunManagerInstanceSettingsTest`? It seems like instance type, price type and cloud region can be referred as instance settings.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: I have already renamed it to "PipelineRunManagerLaunchTest" in this branch: https://github.com/epam/cloud-pipeline/pull/1797

according to this issue and Maria's comment : https://github.com/epam/cloud-pipeline/issues/1776#issuecomment-776663434

So I will rename it to PipelineRunManagerLaunchTest there in order to avoid merge conflict.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done! Omit such case in some other tests, where we don't have to verify some mocks according to test's name.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done for every case where we have assertThrows.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerInstanceTypePriceTypeAndCloudRegionTest.java
Sergeev-Ilya: Done!

Sergeev-Ilya:
Solved the comments above

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+        doReturn(true).when(permissionHelper).isAllowed(any(), any());
+    }
+
+    @Test
+    public void launchPipelineShouldValidateToolInstanceType() {
+        launchTool(configuration, INSTANCE_TYPE);
+
+        verify(toolManager).loadByNameOrId(eq(IMAGE));
+        verify(instanceOfferManager).isToolInstanceAllowed(eq(INSTANCE_TYPE), eq(getResource()), eq(ID), eq(true));
+    }
```
tcibinan: Let's omit `verify(toolManager).loadByNameOrId(eq(IMAGE))` calls as well in most of the test because it also seems like extra verifactions non related to the specific tests.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+
+    @Test
+    public void launchPipelineShouldNotValidateToolInstanceTypeIfItIsNotSpecified() {
+        launchTool(configuration, null);
+
+        verify(toolManager).loadByNameOrId(eq(IMAGE));
+        verify(instanceOfferManager, times(0)).isToolInstanceAllowed(any(), any(), eq(ID), eq(true));
+        verify(instanceOfferManager).isPriceTypeAllowed(eq(SPOT), eq(getResource()), eq(false));
```
tcibinan: Let's better check only instance type allowance using the verification from the original [test](https://github.com/epam/cloud-pipeline/blob/9ce7c2f068113bedf418bde38bbc8843382caddc/api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java#L389-L390).

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+        verify(instanceOfferManager).isToolInstanceAllowed(eq(INSTANCE_TYPE), eq(getResource()), eq(ID_2), eq(true));
+    }
+
+    @Test
+    public void launchPipelineShouldValidatePipelineInstanceType() {
+        launchPipeline(configuration, INSTANCE_TYPE);
+
+        verify(instanceOfferManager).isInstanceAllowed(eq(INSTANCE_TYPE), eq(ID), eq(true));
+        verify(instanceOfferManager).isPriceTypeAllowed(eq(SPOT), eq(null), eq(false));
+    }
```
tcibinan: Price type check can be omitted as in the original [test](https://github.com/epam/cloud-pipeline/blob/9ce7c2f068113bedf418bde38bbc8843382caddc/api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java#L409). Same for all the similar checks in the class.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+
+    @Test
+    public void workerRunShouldUseCloudRegionFromConfiguration() {
+        final PipelineRun pipelineRun = launchPipelineWithParentId(configuration);
+
+        assertThat(pipelineRun.getInstance().getCloudRegionId(), is(ID));
+
+        verify(pipelineRunDao).loadPipelineRun(eq(PARENT_RUN_ID));
+        verify(permissionManager).isRunSshAllowed(eq(parentRun));
+        verify(cloudRegionManager).load(eq(ID));
```
tcibinan: It feels like this all the `verify` calls can be omitted without making test worse. Same for all similar checks in the class.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+        return launchPipeline(configuration, new Pipeline(), INSTANCE_TYPE, PARENT_RUN_ID);
+    }
+
+    private PipelineRun launchPipeline(final PipelineConfiguration configuration, final Pipeline pipeline,
+                                       final String instanceType, final Long parentRunId) {
+        return pipelineRunManager.launchPipeline(configuration, pipeline, null, instanceType, null,
+                null, null, parentRunId, null, null, null);
+    }
+
+    private ContextualPreferenceExternalResource getResource() {
```
tcibinan: Let's inline this resource to the class field.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+    private PipelineRunDao pipelineRunDao;
+
+    @Mock
+    private PreferenceManager preferenceManager;
+
+    private static final String IMAGE = "testImage";
+    private final Tool tool = getTool(IMAGE, DEFAULT_COMMAND);
+    private final AwsRegion defaultAwsRegion = getDefaultAwsRegion(ID);
+    private final AwsRegion nonAllowedAwsRegion = getNonDefaultAwsRegion(ID_2);
+    private final AwsRegion notDefaultAwsRegion = getNonDefaultAwsRegion(ID_3);
```
tcibinan: What do you think about using the same naming notation for region ids as in the original [test](https://github.com/epam/cloud-pipeline/blob/9ce7c2f068113bedf418bde38bbc8843382caddc/api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerTest.java#L113-L115)? It feels more descriptive than just number indexes.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -0,0 +1,401 @@
...
+    @Mock
+    private CheckPermissionHelper permissionHelper;
+
+    @Mock
+    private PipelineRunDao pipelineRunDao;
+
+    @Mock
+    private PreferenceManager preferenceManager;
+
+    private static final String IMAGE = "testImage";
```
tcibinan: Could you please move this static field declaration closer to other static ones?

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: I used CommonCreatorConstants to avoid extra class fields, but I can make them if there is a necessity. Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done! Finally understood the principal of asserting and verifying the minimum :)

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

Sergeev-Ilya:
Fixed all comments + start using new mocking approach.

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -320,48 +322,56 @@ public void workerRunShouldUseParentRunCloudRegionWithParentRunIdPassedAsParamet
         assertThat(pipelineRun.getInstance().getCloudRegionId(), is(NON_DEFAULT_REGION_ID));
     }
 
-    private void mockToolManager() {
-        doReturn(tool).when(toolManager).loadByNameOrId(eq(IMAGE));
-        doReturn(tool).when(toolManager).resolveSymlinks(eq(IMAGE));
+    private void mockCloudRegions() {
+        mock(defaultAwsRegion);
+        mock(nonAllowedAwsRegion, false);
+        mock(notDefaultAwsRegion);
```
tcibinan: For simple cases it is better to have two methods instead of one if you have a single boolean parameter. It helps code readability a bit because seeing `true` or `false` in the parameters doesn't really tell you anything.

```
mock(allowedRegion);
mock(notDefaultAwsRegion);
mockRestricted(notAllowedRegion);
```

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
```diff
@@ -142,13 +143,14 @@
...
         MockitoAnnotations.initMocks(this);
 
-        mockCloudRegionManager();
-        mockInstanceOfferManager();
-        mockPreferenceManager();
-        mockPipelineConfigurationManager();
-        mockToolManager();
-        mockParentRun();
+        mockCloudRegions();
+        mockPreferences();
```
tcibinan: I think it's ok to inline these grouping mock methods. It'll show exactly what we mock in the test class right in place.

tcibinan (changes requested):

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

api/src/test/java/com/epam/pipeline/manager/pipeline/PipelineRunManagerLaunchTest.java
Sergeev-Ilya: Done!

tcibinan (approved):

Issue #1788, opened 2021-02-08
Unit tests: fix UserManagerTests


mzueva:
**Background**
At the moment `UserManagerTests` fails if AWS credentials are not present on the instance running tests since cloud-related S3 storage code is not fully mocked.

**Approach**
Mock any methods using actual S3 calls.

PR #1794, opened 2021-02-09
Issue #1788: fixed failing tests

youkofan:
Issue #1788


api/src/test/java/com/epam/pipeline/manager/user/UserManagerTest.java
```diff
@@ -109,20 +121,33 @@
...
+    private StorageProviderManager storageProviderManager;
+
+    @MockBean
+    private DataStorageValidator storageValidator;
 
     @Before
     public void setUpPreferenceManager() {
+        final StorageProvider<S3bucketDataStorage> storageProvider = new S3StorageProvider(authManager, messageHelper,
+                                                      cloudRegionManager, preferenceManager, stsCredentialsGenerator);
+        doReturn(storageProvider).when(storageProviderManager).getStorageProvider(any());
```
mzueva: Can't we return here `mock(S3StorageProvider.class)` instead of creating an instance manually?

Issue #1717, opened 2021-01-11
Store the Object Storages tags in the DB


sidoruka:
**Background**
At the moment, Cloud Pipeline stores the tags of the objects, stored in S3/GCS/BLOB, using the Cloud-native approach. This makes it hard (near to impossible) to include this information into the search and billing machinery.

**Approach**
* Tags shall be now stored in the Cloud Pipeline's DB (read/added/deleted/updated using the existing API)
* This feature shall be implemented for all the supported Cloud Providers (AWS/GCP/Azure)

mzueva:
@tcibinan @sidoruka 
Some points to consider:
- NFS files shall be tagged as well
- Tags shall be supported for files only
- Tags shall be deleted when a file deleted from storage using API or pipe CLI
- Init scripts to transfer tags from Cloud provider to DB shall be provided

PR #1798, opened 2021-02-11
Implement data storage object tags management

tcibinan:
Relates to #1717.

The pull request brings support for data storage object tags management as a replacement for original cloud-based tags management. From now on all the data storage operations both in cli and api cause explicit data storage object tags update.


api/src/main/resources/db/migration/v2021.03.16_16.00__issue_1717_data_storage_tags.sql
```diff
@@ -0,0 +1,90 @@
...
+        VALUES (get_nfs_datastorage_root(path))
+        ON CONFLICT (datastorage_root_path)
+        DO NOTHING;
+        RETURN new;
+    END IF;
+END;
+$BODY$
+language plpgsql;
+
+CREATE TRIGGER create_datastorage_root_trigger
```
mzueva: I'd suggest not to use triggers and move this logic to Java code

pipe-cli/src/api/data_storage.py
```diff
@@ -336,6 +336,81 @@ def delete_object_tags(cls, identifier, path, tags, version):
         else:
             raise RuntimeError("Failed to update tags for object {}.".format(path))
 
+    @classmethod
+    def batch_insert_object_tags(cls, identifier, requests):
+        api = cls.instance()
```
mzueva, line 315: Let's refactor this code and extracts duplicates

api/src/main/java/com/epam/pipeline/dao/datastorage/tags/DataStorageTagDao.java
```diff
@@ -0,0 +1,192 @@
+package com.epam.pipeline.dao.datastorage.tags;
```
mzueva, line 1: Can we consider using JPA for tag requests?

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -940,10 +1127,47 @@ private DataStorageFolder createDataStorageFolder(final AbstractDataStorage stor
...
+        final String rootPath = rootAndRelativePaths.getLeft();
+        final String relativePath = rootAndRelativePaths.getRight();
+        final Map<String, String> defaultTags = Collections.singletonMap(ProviderUtils.OWNER_TAG_KEY, authorizedUser);
+        tagManager.insert(rootPath, new DataStorageObject(relativePath, null), defaultTags);
+        if (storage.isVersioningEnabled()) {
+            tagManager.insert(rootPath, new DataStorageObject(relativePath, version), defaultTags);
+        }
+    }
+
+    private Pair<String, String> getRootAndRelativePaths(final AbstractDataStorage storage, final String path) {
```
mzueva: It would be nice to introduce a class instead of using Pair

api/src/main/resources/dao/datastorage-tag-dao.xml
```diff
@@ -0,0 +1,208 @@
...
+                <!--deleteAllTagsByPathPatternQuery-->
+                <![CDATA[
+                    DELETE
+                    FROM pipeline.datastorage_tag t
+                    WHERE t.datastorage_root_id IN (
+                        SELECT r.datastorage_root_id
+                        FROM pipeline.datastorage_root r
+                        WHERE r.datastorage_root_path = :DATASTORAGE_ROOT_PATH
+                    )
+                    AND t.datastorage_path LIKE :DATASTORAGE_PATH
```
mzueva, line 175: We need to check `LIKE` performance

pipe-cli/src/utilities/storage/s3.py
```diff
@@ -225,17 +243,7 @@ def transfer(self, source_wrapper, destination_wrapper, path=None, relative_path
...
-            ObjectTaggingManager(self.session, source_bucket, source_region), path)
-        if not tags and object_tags:
-            tags = self.convert_object_tags(object_tags)
-        if not self.has_required_tag(tags, 'CP_SOURCE'):
-            tags += ('CP_SOURCE=s3://{}/{}'.format(source_bucket, path),)
-        if not self.has_required_tag(tags, 'CP_OWNER'):
-            tags += ('CP_OWNER={}'.format(self._get_user()),)
-        TransferManager.ALLOWED_COPY_ARGS.append('Tagging')
         extra_args = {
-            'Tagging': self._convert_tags_to_url_string(tags),
```
mzueva, line 238: `TransferFromHttpOrFtpToS3Manager` and `UploadManager` still pass tags to s3 API

mzueva (changes requested):

pipe-cli/src/utilities/storage/s3.py
tcibinan: Both cli and api still pass tags whenever possible to cloud provider APIs because it provides possible data storage object tags backup source and it doesn't have any performance impact on the operations.

api/src/main/resources/dao/datastorage-tag-dao.xml
tcibinan: Discussed and decided to keep it as is for now. The operations which depends on such sql calls are not critical and the `LIKE` performance is tolerable at least for now.

api/src/main/java/com/epam/pipeline/dao/datastorage/tags/DataStorageTagDao.java
tcibinan: Discussed and decided to keep the current dao implementation due to the number of queues, their complexity and possible optimizations which may increase the complexity even more.

api/src/main/resources/db/migration/v2021.03.16_16.00__issue_1717_data_storage_tags.sql
tcibinan: Done

pipe-cli/src/api/data_storage.py
tcibinan: Added a handy method `api.base.API.retryable_call` which helped to refactor this changes and also can be useful in other places in pipe cli as a replacement for old style `api.base.API.call`.

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
tcibinan: The root and relative paths management was incorporated to `AbstractDataStorage` class. It simplified the call site and removed the usage of any pair-like instance at all.

mzueva (approved):

Issue #1802, opened 2021-02-11
Add pause/resume operations support to the `pipe`


sidoruka:
**Background**
At the moment, users may automate the `pause` and `resume` operation only via the API calls. While this is possible, it's not that nice as could be.
It would be great to expose `pipe pause` and `pipe resume` operations to the CLI.

**Approach**
* Add `pipe pause [RUN_ID] --check-size --sync/-s` and `pipe resume [RUN_ID] --sync/-s` commands
* In a `--sync` mode - we shall wait for the operation to complete. If the command is canceled - we shall inform the user, that the pause/resume operations can't be stopped and it will continue to run in the background

sidoruka:
Please backport to `release/0.16`

Wedds:
<h1> E2E test cases description</h1>

<h2>`pipe pause`</h2>

**Requirements:**
1. Active on-demand run:

    1.1. Start new run using UI or CLI 

    1.2. Wait till the run is initialized (has status 'RUNNING')

For the cases description, assume, that the active run has RunID=*1234* and its' image is *centos:latest*.
<hr>
<h3>CLI-PAUSE-TC-1</h3>

Run is paused in the non-sync mode

**Commands to run:**
1. > pipe pause 1234

**Expected result:**

1. CLI output:
    ```
    Pausing RunID 1234 of "centos:latest@None"
    ```
2. Run state is changed to 'PAUSED' in a reasonable time.

<hr>
<h3>CLI-PAUSE-TC-2</h3>

Run is paused in sync mode

**Commands to run:**
1. > pipe pause --sync 1234

**Expected result:**

1. CLI output:
    ```
    Pausing RunID 1234 of "centos:latest@None"
    ```
2. Run state is changed to 'PAUSED' in a reasonable time. 
3. One more string appears in CLI output:
    ```
    Pausing RunID 1234 of "centos:latest@None"
    RunID 1234 of "centos:latest@None" is paused
    ```
<hr>
<h3>CLI-PAUSE-TC-3</h3>

Run is paused in sync mode, execution is aborted before the operation starts at the server

**Commands to run:**
1. > pipe pause --sync 1234
2. Cancel execution using keyboard interruption (press Ctrl+C) before pausing called at the server-side ('Pausing RunID ...' message didn't appear in CLI output)


**Expected result:**
1. CLI output contains interruption character
    ```
    ^C
    ```
2. Run state didn't change from 'RUNNING'

<hr>
[...]

PR #1814, opened 2021-02-17
Issue #1802 Add pause/resume operations to the pipe CLI

Wedds:
This PR is related to issue #1802

It brings `pause` and `resume` operations to the pipe CLI.

Besides, it is allowing specifying commands, which abortion will be indicated with a custom message.


pipe-cli/src/utilities/pipeline_run_operations.py
```diff
@@ -290,6 +283,52 @@ def run(cls, pipeline, config, parameters, yes, run_params, instance_disk, insta
...
+            pipeline_name = cls.extract_pipeline_name(pipeline_run_model)
+            pipeline_version = pipeline_run_model.version
+            click.echo('Pausing RunID {} of "{}@{}"'.format(run_id, pipeline_name, pipeline_version))
+            if sync:
+                status = cls.get_pausing_pipeline_status(run_id)
+                if status == 'PAUSED':
+                    click.echo('RunID {} of "{}@{}" is paused'.format(run_id, pipeline_name, pipeline_version))
+                    sys.exit(1)
+                else:
+                    click.echo('Failed pausing RunID {} of "{}@{}"'.format(run_id, pipeline_name, pipeline_version))
```
ekazachkova: Let's add `err=True` since this is the error message

pipe-cli/pipe.py
```diff
@@ -698,6 +701,27 @@ def stop(run_id, yes):
...
+    """
+    PipelineRunOperations.pause(run_id, check_size, sync)
+
+
+@cli.command(name='resume')
+@click.argument('run-id', required=True, type=int)
+@click.option('-s', '--sync', is_flag=True, help=SYNC_FLAG_DESCRIPTION)
+@Config.validate_access_token
+def resume(run_id, sync):
+    """Resumes a running pipeline
```
ekazachkova: Maybe `paused` would be more clear instead of `running`?

ekazachkova (changes requested):

pipe-cli/src/utilities/pipeline_run_operations.py
Wedds: Done

pipe-cli/pipe.py
Wedds: Done

ekazachkova (approved):

Issue #1808, opened 2021-02-15
Search shall handle queries with special symbols


mzueva:
**Describe the bug**
At the moment global search requests fail when query contains special symbols, e.g. `folder/file.txt`. Such symbols shall be escaped in order to process these queries.

PR #1819, opened 2021-02-18
Issue #1808 Escape unsupported special characters in search queries

Wedds:
This PR is related to issue #1808.

The user is aware of 'Special expressions' of the Global Search and [characters](https://cloud-pipeline.com/release_notes/v.0.14/v.0.14_-_Release_notes/#global-search), which function as operators in a query.
But since not all the symbols are declared as supported, it would be convenient to automatically escape them to avoid search failures and side-effects.
From now, all the special symbols of the `QueryString` query, which are not described as an operator for Global Search, are escaped automatically.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/search/SearchRequestBuilder.java
+++ b/api/src/main/java/com/epam/pipeline/manager/search/SearchRequestBuilder.java
@@ -69,6 +69,8 @@ public class SearchRequestBuilder {
     private static final String SIZE_FIELD = "size";
     private static final String NAME_FIELD = "id";
     private static final String ES_FILE_INDEX_PATTERN = "cp-%s-file-%d";
+    private static final Set<Character> UNSUPPORTED_SPECIAL_ES_QUERY_CHARACTERS =
+        new HashSet<>(Arrays.asList('\\', '!', ':', '^', '[', ']', '{', '}', '?', '&', '/'));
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/epam/pipeline/manager/search/SearchRequestBuilder.java
```diff
@@ -183,4 +184,17 @@ private void addAclFilters(final BoolQueryBuilder queryBuilder, final PipelineUs
                         .orElseThrow(() -> new SearchException("Missing index name for type: " + type)))
                 .toArray(String[]::new);
     }
+
+    private String escapeUnsupportedSpecialCharacters(final String string) {
+        final StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < string.length(); i++) {
+            final char c = string.charAt(i);
+            if (c == '\\' || c == '!' || c == ':' || c == '^' || c == '[' || c == ']' || c == '{' || c == '}'
```
mzueva: Could you please add all special characters to a set? E.g. `Set<Character> special = new HashSet<>(Arrays.asList('\\', '!')); if (special.contains(c)) ......;`

Wedds: Done

mzueva (approved):

Issue #1840, opened 2021-03-05
Configure idle timeout for autoscaled SGE workers


mzueva:
**Background**
At the moment autoscaled SGE workers are stopped after 30 seconds idle timeout. It would be helpful to allow users to configure this idle timeout via run parameters.

PR #1846, opened 2021-03-09
Issue #1840 SGE workers idle timeout

Wedds:
This PR is related to issue #1840.

It allows specifying an idle timeout for the cluster workers, using `CP_CAP_AUTOSCALE_IDLE_TIMEOUT` run parameter.
That means, that if a node was active (performing a task) it will be scaled down not exactly after 'scaling down' timeout since the last active cluster task, but after passing the idle timeout of activity for that specific node.

### DIFF
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -1421,9 +1421,13 @@ if __name__ == '__main__':
                                          storage_file=os.path.join(shared_work_dir, '.autoscaler.storage'))
     scale_up_timeout = int(api.retrieve_preference('ge.autoscaling.scale.up.timeout', default_value=30))
     scale_down_timeout = int(api.retrieve_preference('ge.autoscaling.scale.down.timeout', default_value=30))
-    run_custom_scale_down_timeout = os.getenv('CP_CAP_AUTOSCALE_IDLE_TIMEOUT', None)
+    run_custom_scale_down_timeout = os.getenv('CP_CAP_AUTOSCALE_IDLE_TIMEOUT')
     if run_custom_scale_down_timeout:
-        scale_down_timeout = int(run_custom_scale_down_timeout)
+        if run_custom_scale_down_timeout.isdigit():
+            scale_down_timeout = int(run_custom_scale_down_timeout)
+        else:
+            Logger.warn('Idle timeout [%s], specified for workers via run parameter, is illegal. '
+                        'Using global value [%s] instead.' % (run_custom_scale_down_timeout, scale_down_timeout))
     scale_up_polling_timeout = int(api.retrieve_preference('ge.autoscaling.scale.up.polling.timeout',
                                                            default_value=900))
     scale_up_polling_delay = int(os.getenv('CP_CAP_AUTOSCALE_SCALE_UP_POLLING_DELAY', 10))
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -813,9 +813,10 @@ class MemoryHostStorage:
         self._validate_existence(host)
         self._storage.pop(host)
 
-    def update_host_activity(self, host, timestamp):
-        self._validate_existence(host)
-        self._storage[host] = timestamp
+    def update_hosts_activity(self, hosts, timestamp):
+        for host in hosts:
+            self._validate_existence(host)
+            self._storage[host] = timestamp
 
     def get_host_activity(self, host):
         self._validate_existence(host)
@@ -857,20 +858,21 @@ class FileSystemHostStorage:
 
         :param host: Additional host name.
         """
-        hosts = self._load_hosts_dict()
+        hosts = self._load_hosts_stats()
         if host in hosts:
             raise ScalingError('Host with name \'%s\' is already in the host storage' % host)
         hosts[host] = None
         self._update_storage_file(hosts)
 
-    def update_host_activity(self, host, timestamp):
-        hosts = self._load_hosts_dict()
-        self._validate_existence(host, hosts)
-        hosts[host] = timestamp
-        self._update_storage_file(hosts)
+    def update_hosts_activity(self, hosts, timestamp):
+        latest_hosts_stats = self._load_hosts_stats()
+        for host in hosts:
+            self._validate_existence(host, latest_hosts_stats)
+            latest_hosts_stats[host] = timestamp
+        self._update_storage_file(latest_hosts_stats)
 
     def get_host_activity(self, host):
-        hosts = self._load_hosts_dict()
+        hosts = self._load_hosts_stats()
         self._validate_existence(host, hosts)
         return hosts[host]
 
@@ -880,7 +882,7 @@ class FileSystemHostStorage:
 
         :param host: Additional host name.
         """
-        hosts = self._load_hosts_dict()
+        hosts = self._load_hosts_stats()
         self._validate_existence(host, hosts)
         hosts.pop(host)
         self._update_storage_file(hosts)
@@ -894,9 +896,9 @@ class FileSystemHostStorage:
                                                                      'file': self.storage_file})
 
     def load_hosts(self):
-        return list(self._load_hosts_dict().keys())
+        return list(self._load_hosts_stats().keys())
 
-    def _load_hosts_dict(self):
+    def _load_hosts_stats(self):
         """
         Load all additional hosts from storage.
 
@@ -1018,9 +1020,12 @@ class GridEngineAutoscaler:
         post_scale_additional_hosts = self.host_storage.load_hosts()
         Logger.info('There are %s additional pipelines.' % len(post_scale_additional_hosts))
 
-    def _update_hosts_activity(self, now, running_jobs):
+    def _update_hosts_activity(self, scaling_step_start, running_jobs):
+        active_hosts = set()
         for job in running_jobs:
-            self.host_storage.update_host_activity(job.hosts, now)
+            active_hosts.add(job.hosts)
+        if active_hosts:
+            self.host_storage.update_hosts_activity(active_hosts, scaling_step_start)
 
     def _filter_pending_job(self, updated_jobs):
         # kill jobs that are pending and can't be satisfied with requested resource
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -932,7 +932,7 @@ class FileSystemHostStorage:
 class GridEngineAutoscaler:
 
     def __init__(self, grid_engine, cmd_executor, scale_up_handler, scale_down_handler, host_storage, scale_up_timeout,
-                 scale_down_timeout, max_additional_hosts, idle_timeout=None, clock=Clock()):
+                 scale_down_timeout, max_additional_hosts, idle_timeout=30, clock=Clock()):
         """
         Grid engine autoscaler.
 
@@ -964,7 +964,7 @@ class GridEngineAutoscaler:
         self.clock = clock
 
         self.latest_running_job = None
-        self.idle_timeout = timedelta(seconds=idle_timeout) if idle_timeout else None
+        self.idle_timeout = timedelta(seconds=idle_timeout)
 
     def scale(self):
         now = self.clock.now()
@@ -974,8 +974,7 @@ class GridEngineAutoscaler:
         updated_jobs = self.grid_engine.get_jobs()
         running_jobs = [job for job in updated_jobs if job.state == GridEngineJobState.RUNNING]
         pending_jobs = self._filter_pending_job(updated_jobs)
-        if self.idle_timeout:
-            self._update_hosts_activity(now, running_jobs)
+        self._update_hosts_activity(now, running_jobs)
         if running_jobs:
             self.latest_running_job = sorted(running_jobs, key=lambda job: job.datetime, reverse=True)[0]
         if pending_jobs:
@@ -1058,7 +1057,7 @@ class GridEngineAutoscaler:
     def _scale_down(self, running_jobs, additional_hosts, scaling_period_start=None):
         active_hosts = set([host for job in running_jobs for host in job.hosts])
         inactive_additional_hosts = [host for host in additional_hosts if host not in active_hosts]
-        if self.idle_timeout and scaling_period_start:
+        if scaling_period_start:
             inactive_additional_hosts = self._filter_valid_idle_hosts(inactive_additional_hosts, scaling_period_start)
         if inactive_additional_hosts:
             Logger.info('There are %s inactive additional child pipelines. '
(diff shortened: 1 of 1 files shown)


workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -799,36 +799,50 @@ class MemoryHostStorage:
...
 
     def load_hosts(self):
-        return list(self._storage)
+        return list(self._storage.keys())
 
     def clear(self):
-        self._storage = list()
+        self._storage = dict()
+
+    def validate_existence(self, host):
```
tcibinan: Let's rename method to `_validate_existence` as long as it is a private method.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -843,53 +857,79 @@ def add_host(self, host):
...
                 return hosts
         else:
-            return []
+            return {}
 
     def clear(self):
-        self._update_storage_file([])
+        self._update_storage_file({})
+
+    def validate_existence(self, host, hosts_dict):
```
tcibinan: Private method too.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -843,53 +857,79 @@ def add_host(self, host):
...
+        if host not in hosts_dict:
+            raise ScalingError('Host with name \'%s\' doesn\'t exist in the host storage' % host)
 
 
 class GridEngineAutoscaler:
 
     def __init__(self, grid_engine, cmd_executor, scale_up_handler, scale_down_handler, host_storage, scale_up_timeout,
-                 scale_down_timeout, max_additional_hosts, clock=Clock()):
+                 scale_down_timeout, max_additional_hosts, idle_timeout=None, clock=Clock()):
```
tcibinan: Could you please add a small doc in the method's pydoc about the new argument starting with `:param` keyword.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -963,17 +1005,21 @@ def scale(self):
...
                 Logger.info('There are no previously running jobs. Scaling down is required.')
-                self._scale_down(running_jobs, additional_hosts)
+                self._scale_down(running_jobs, additional_hosts, now)
         Logger.info('Finish scaling step at %s.' % self.clock.now())
         post_scale_additional_hosts = self.host_storage.load_hosts()
         Logger.info('There are %s additional pipelines.' % len(post_scale_additional_hosts))
 
+    def update_hosts_activity(self, now, running_jobs):
+        for job in running_jobs:
+            self.host_storage.update_host_activity(job.hosts, now)
```
tcibinan: I believe that such operation can be very inefficient for huge autoscaled runs with thousands of jobs. For each of the running job the host storage file will be updated.

Probably a more efficient approach would be to collect all unique active hosts from all the running jobs and update host storage file only once using a single `self.host_storage.update_host_activity` call.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -963,17 +1005,21 @@ def scale(self):
...
                                 'Scaling down is not required.' % self.scale_down_timeout.seconds)
             else:
                 Logger.info('There are no previously running jobs. Scaling down is required.')
-                self._scale_down(running_jobs, additional_hosts)
+                self._scale_down(running_jobs, additional_hosts, now)
         Logger.info('Finish scaling step at %s.' % self.clock.now())
         post_scale_additional_hosts = self.host_storage.load_hosts()
         Logger.info('There are %s additional pipelines.' % len(post_scale_additional_hosts))
 
+    def update_hosts_activity(self, now, running_jobs):
```
tcibinan: Please add `_` prefix to all new private methods here as well.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1415,6 +1482,7 @@ def fetch_worker_launch_system_params(api, master_run_id):
                                          storage_file=os.path.join(shared_work_dir, '.autoscaler.storage'))
     scale_up_timeout = int(api.retrieve_preference('ge.autoscaling.scale.up.timeout', default_value=30))
     scale_down_timeout = int(api.retrieve_preference('ge.autoscaling.scale.down.timeout', default_value=30))
+    idle_timeout = get_idle_timeout()
```
tcibinan: I believe it's okay to fail fast here in case specified value is invalid. Also it is a nice feature and I believe we can use it by default with some fixed value like 30 seconds.

```
idle_timeout = int(os.getenv('CP_CAP_AUTOSCALE_IDLE_TIMEOUT', 30))
```

As a result we can handle this value as non null everywhere.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1002,9 +1048,11 @@ def scale_up(self, resource):
         Logger.info('Start grid engine SCALING UP.')
         return self.scale_up_handler.scale_up(resource)
 
-    def _scale_down(self, running_jobs, additional_hosts):
+    def _scale_down(self, running_jobs, additional_hosts, scaling_period_start=None):
```
tcibinan, line 1081: There is a single call of `self._scale_down` without the last argument `scaling_period_start`. Let's always pass the  argument and handle it as non nullable in the method below.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1002,9 +1048,11 @@ def scale_up(self, resource):
...
         active_hosts = set([host for job in running_jobs for host in job.hosts])
         inactive_additional_hosts = [host for host in additional_hosts if host not in active_hosts]
+        if self.idle_timeout and scaling_period_start:
+            inactive_additional_hosts = self.filter_valid_idle_hosts(inactive_additional_hosts, scaling_period_start)
```
tcibinan: Let's use another variable like `idle_additional_hosts` and log both types of hosts. I think it may give us a little bit of an insight on what's happening during troubleshooting.

```python
inactive_additional_hosts = ...
if inactive_additional_hosts:
    Logger.info('There are %s inactive additional child pipelines. '
                'Scaling down will be performed.' % len(inactive_additional_hosts))

idle_additional_hosts = ...
if idle_additional_hosts:
    Logger.info('There are %s idle additional child pipelines. '
                'Scaling down will be performed.' % len(idle_additional_hosts))
```

tcibinan (changes requested):

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done, updating and loading activity details in a batch manner

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -963,17 +1016,24 @@ def scale(self):
...
-                self._scale_down(running_jobs, additional_hosts)
+                self._scale_down(running_jobs, additional_hosts, now)
         Logger.info('Finish scaling step at %s.' % self.clock.now())
         post_scale_additional_hosts = self.host_storage.load_hosts()
         Logger.info('There are %s additional pipelines.' % len(post_scale_additional_hosts))
 
+    def _update_hosts_activity(self, scaling_step_start, running_jobs):
+        active_hosts = set()
+        for job in running_jobs:
+            active_hosts.add(job.hosts)
```
tcibinan: It doesn't seem like you can add `list` to a set using `add` method. It produces the following error: `unhashable type: 'list'`.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -799,97 +799,148 @@ class MemoryHostStorage:
...
                     stripped_line = line.strip().strip(FileSystemHostStorage._LINE_BREAKER)
                     if stripped_line:
-                        hosts.append(stripped_line)
+                        host_stats = stripped_line.strip().split(FileSystemHostStorage._VALUE_BREAKER)
+                        if host_stats:
+                            hostname = host_stats[0]
+                            last_activity = datetime.strptime(host_stats[1], FileSystemHostStorage._DATETIME_FORMAT) \
+                                if len(host_stats) > 1 \
+                                else None
+                            hosts[hostname] = last_activity
```
tcibinan: I'm kinda worried that we use space as a delimiter and also have space in the date time formatting pattern. Are you sure that we won't parse only date without time? Probably we can use different delimiter such as tab to differentiate columns.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -799,97 +799,148 @@ class MemoryHostStorage:
...
+        hosts = self._load_hosts_stats()
+        self._validate_existence(host, hosts)
+        hosts.pop(host)
         self._update_storage_file(hosts)
 
     def _update_storage_file(self, hosts):
-        self.executor.execute(FileSystemHostStorage._REPLACE_FILE % {'content': '\n'.join(hosts),
+        hosts_summary_table = []
+        for host, last_activity in hosts.items():
+            formatted_activity = last_activity.strftime(FileSystemHostStorage._DATETIME_FORMAT) if last_activity else ''
```
tcibinan: As long as we can't have empty `last_activity` let's handle it as non nullable.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1002,20 +1062,33 @@ def scale_up(self, resource):
...
                 self.host_storage.remove_host(inactive_additional_host)
         else:
-            Logger.info('There are no inactive additional child pipelines. Scaling down will not be performed.')
+            Logger.info('There are no additional child pipelines for scaling down.')
+
+    def _filter_valid_idle_hosts(self, inactive_host_candidates, scaling_period_start):
+        inactive_hosts = []
+        hosts_activity = self.host_storage.get_hosts_activity(inactive_host_candidates)
+        for host, last_activity in hosts_activity.items():
+            if not last_activity or scaling_period_start > last_activity + self.idle_timeout:
```
tcibinan: Let's simplify non nullable `last_activity` usage here as well.

```python
if scaling_period_start > last_activity + self.idle_timeout:
```

tcibinan (changes requested):

workflows/pipe-common/scripts/autoscale_sge.py
tcibinan: It's better to keep it the existing way. This single call performs better without `scaling_period_start` being specified because it only applies to deadlock scenarios.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -963,17 +1038,25 @@ def scale(self):
...
             else:
                 Logger.info('There are no previously running jobs. Scaling down is required.')
-                self._scale_down(running_jobs, additional_hosts)
+                self._scale_down(running_jobs, additional_hosts, now)
         Logger.info('Finish scaling step at %s.' % self.clock.now())
         post_scale_additional_hosts = self.host_storage.load_hosts()
         Logger.info('There are %s additional pipelines.' % len(post_scale_additional_hosts))
 
+    @staticmethod
+    def update_running_jobs_host_activity(running_jobs, host_storage, scaling_step_start):
```
tcibinan: It feels like this method should be a part of `host_storage` implementations.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -963,17 +1038,25 @@ def scale(self):
...
+        active_hosts = set()
+        for job in running_jobs:
+            active_hosts.update(job.hosts)
```
tcibinan: We need to update activity for only the existing hosts in host storage. This way we manager only additional hosts and ignore master and static workers.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -799,97 +813,156 @@ class MemoryHostStorage:
...
             raise ScalingError('Host with name \'%s\' is already in the host storage' % host)
-        hosts.append(host)
+        hosts[host] = self.clock.now()
         self._update_storage_file(hosts)
 
+    def update_hosts_activity(self, hosts, timestamp):
+        latest_hosts_stats = self._load_hosts_stats()
+        for host in hosts:
+            if host != self.master_host:
+                self._validate_existence(host, latest_hosts_stats)
```
tcibinan: We can get rid of this and all other additional `master_host` management since master host won't be written to the hosts  storage in the first place.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -249,6 +249,9 @@ def get_jobs(self):
                 current_host = None
         return jobs.values()
 
+    def filter_running_job(self, jobs):
+        return [job for job in jobs if job.state == GridEngineJobState.RUNNING]
+
```
tcibinan: I believe it's okay to inline this method since it really depends on the call site. Otherwise we can introduce `get_running_jobs` method which can be optimized later on to retrieve only running jobs from the qstat. But for now I would just inline the method.

tcibinan (changes requested):

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done - ignoring hosts, which are not presented in host_storage

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

workflows/pipe-common/scripts/autoscale_sge.py
Wedds: Done

tcibinan (approved):

Issue #1838, opened 2021-03-03
Faceted Filter search results (GUI)


rodichenko:
**Background**
To introduce a new **Faceted Filter** feature (#1829), we must implement a paged list of search results with previews.

**Approach**
* Each modification of the faceted filter or search criteria must trigger the search (see #1837);
* Search Results list must be paged;
* GUI shall display an object preview on hover; we must re-use and (if necessary) improve current preview controls;
* Item's click must open the one's details page (e.g. pipeline details, tool details etc.); user may also use `Ctrl + click` to open details on the new browser tab.

NShaforostov:
## Clarifications

Search results contain an array of objects (Cloud Pipeline entities - `file`, `storage`, `run`, etc.).

For each object in results, the additional info shall be displayed - according to the entity type.
I suggest to show the following info:

| Entity | Required info | Additional info |
| --- | --- | --- |
| **Folder** | NAME | OWNER |
| **Pipeline** | NAME | OWNER / DESCRIPTION |
| **Configuration** | NAME | OWNER / DESCRIPTION |
| **Run** | RUN_ID | OWNER / DATE_STARTED / DATE_COMPLETED |
| **Storage** | ALIAS_NAME | OWNER / PATH / DESCRIPTION |
| **Tool** | NAME | OWNER / PATH |
| **File** | NAME | OWNER / PATH / TYPE / SIZE / DATE_CHANGED |
| **Issue** | TITLE | OWNER / DESCRIPTION |

These object in results can be shown in two formats:
- line view - each result is presented by the "row" in the list, additional info is placed in line
- table view - all results are presented by the single table, additional info is placed in columns

User shall have the ability to switch between formats.
Also: last selected format should be saved for the current user and used for the next time he(she) will leverage the **Advanced Search**.

Example of the both formats:
![image](https://user-images.githubusercontent.com/45459424/111843667-a6747a00-8912-11eb-9911-312b48b090a6.png)
![image](https://user-images.githubusercontent.com/45459424/111843692-b0967880-8912-11eb-8345-c6b1fbedba8e.png)

sidoruka:
@mzueva @rodichenko let's implement the list of "additional" attributes, as @NShaforostov suggests in the following way:
* **As a step 1**: The "base" attributes (from @NShaforostov comment) shall be always indexed, returned to the client and shown in the GUI
[...]

PR #1924, opened 2021-04-14
Issue #1838: Faceted Filter search results (API)

ekazachkova:
This PR provides fix for issue #1838  - the size of aggregations increased


api/src/main/java/com/epam/pipeline/manager/search/SearchRequestBuilder.java
```diff
@@ -258,6 +259,7 @@ private QueryBuilder prepareFacetedQuery(final String requestQuery) {
 
     private void addTermAggregationToSource(final SearchSourceBuilder searchSource, final String facet) {
         final TermsAggregationBuilder aggregationBuilder = AggregationBuilders.terms(facet)
+                .size(MAX_AGGREGATIONS_SIZE)
```
mzueva: Could we move this setting to system preferences?

ekazachkova: done

PR #1927, opened 2021-04-15
1805 Versioned Storages reports

SilinPavel:
This PR provides functionality to generate changes reports for Versioned Storages


api/src/main/java/com/epam/pipeline/entity/git/GitDiff.java
```diff
@@ -0,0 +1,35 @@
...
+package com.epam.pipeline.entity.git;
+
+import lombok.Builder;
+import lombok.Data;
+
+import java.util.List;
+
+
+/**
+ * Represents Gitlab repository commit
```
tcibinan: Seems like a missleading javadoc.

api/src/main/java/com/epam/pipeline/entity/git/GitDiffEntry.java
```diff
@@ -0,0 +1,35 @@
...
+package com.epam.pipeline.entity.git;
+
+import com.epam.pipeline.entity.git.gitreader.GitReaderRepositoryCommit;
+import io.reflectoring.diffparser.api.model.Diff;
+import lombok.Builder;
+import lombok.Data;
+
+
+/**
+ * Represents Gitlab repository commit
```
tcibinan: Here too.

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineFileGenerationManager.java
```diff
@@ -61,6 +67,13 @@
         return this.generateFile(templatePath, documentTemplate);
     }
 
+    public Pair<String, byte[]> generateVersionStorageReport(final Long pipelineId,
+                                                             final GitDiffReportFilter reportFilters) {
+        return vsReportTemplateManager.generateReport(pipelineId, reportFilters);
+    }
```
tcibinan: I guess we can get rid of this method and just call `VersionStorageReportTemplateManager.generateReport` from `PipelineApiService`.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/structure/CommitDiffsGrouping.java
```diff
@@ -0,0 +1,24 @@
...
+
+    GroupType type;
+    boolean archive;
+    boolean includeDiff;
+    Map<String, List<GitDiffEntry>> diffGrouping;
+
+    public enum GroupType {
+        BY_COMMIT,
+        BY_FILE
+    }
```
tcibinan: It can be a good idea to extract this inner class to a separate file because it is used as a field type in other class as well.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/structure/CommitDiffsGrouping.java
```diff
@@ -0,0 +1,24 @@
+package com.epam.pipeline.manager.pipeline.documents.templates.structure;
```
tcibinan: It also seems nice to move this class to other similar entity classes package `com.epam.pipeline.entity.git`.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/VSReportTemplates.java
```diff
@@ -0,0 +1,73 @@
...
+    COMMIT_DIFFS("commit_diffs", () -> new VSReportTemplateDiffProcessor(new CommitDiffExtractor()));
+
+    public final String template;
+    public final Supplier<VSReportTemplateProcessor> templateResolver;
+
+    VSReportTemplates(final String template,
+                      final Supplier<VSReportTemplateProcessor> templateResolver) {
+        this.template = template;
+        this.templateResolver = templateResolver;
+    }
```
tcibinan: The constructor can be replaced with the usage of `@RequiredArgsConstructor`.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/VSReportTemplates.java
```diff
@@ -0,0 +1,73 @@
...
+            return "-";
+        }
+        return diff.getFilters().getDateTo().format(Constants.DATE_TIME_FORMATTER);
+    })),
+    FILTER_TYPES("filter_types", () -> new VSReportTemplateTextProcessor((paragraph, storage, diff, reportFilter) -> {
+        if (CollectionUtils.isNullOrEmpty(diff.getFilters().getExtensions())) {
+            return "All Files";
+        }
+        return diff.getFilters().getExtensions();
+    })),
```
tcibinan: It sure looks like these lambdas can be with simplified with the usage of `Optional`.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+    private final PreferenceManager preferenceManager;
+
+    @Autowired
+    public VersionStorageReportTemplateManager(final PipelineManager pipelineManager,
+                                               final GitManager gitManager,
+                                               final PreferenceManager preferenceManager) {
+        this.pipelineManager = pipelineManager;
+        this.gitManager = gitManager;
+        this.preferenceManager = preferenceManager;
+    }
```
tcibinan: The constructor can be replaced with the usage of `@RequiredArgsConstructor`.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+
+    public Pair<String, byte[]> generateReport(final Long pipelineId,
```
tcibinan: What do you think about replacing this pair with some explicit class such as `VersionStorageReport`?

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+        try {
+            final List<Pair<String, XWPFDocument>> diffReportFiles = prepareReportDocs(
+                    loaded, gitDiff, reportFilters
+            );
+
+            if (diffReportFiles.isEmpty()) {
+                throw new IllegalArgumentException("No data for report");
+            }
+
+            return Pair.of(writeReportToStream(loaded, outputStream, diffReportFiles), outputStream.toByteArray());
```
tcibinan: I believe it can be a little bit clearer to use two different methods for report name generation and report content generation.

```java
final String name = resolveReportName(pipeline, diffReportFiles);
final byte[] content = generateReport(pipeline, diffReportFiles);
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+
+    private List<Pair<String, XWPFDocument>> prepareReportDocs(final Pipeline loaded,
+                                                               final GitDiff gitDiff,
+                                                               final GitDiffReportFilter reportFilters)
+            throws IOException {
+        final List<Pair<String, XWPFDocument>> results = new ArrayList<>();
+        String versionStorageTemplatePath = getVersionStorageTemplatePath();
+        Assert.notNull(versionStorageTemplatePath,
+                "Version Storage Report Template not configured, please specify "
+                        + SystemPreferences.VERSION_STORAGE_REPORT_TEMPLATE.getKey());
```
tcibinan: Let's move `Assert.notNull` to `getVersionStorageTemplatePath` and inline `versionStorageTemplatePath` variable.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+        if (reportFilters.isArchive()) {
+            results.addAll(
+                    prepareDiffsForReportDoc(loaded, gitDiff,
+                        GitDiffReportFilter.builder()
+                            .groupType(reportFilters.getGroupType())
+                            .commitsFilter(reportFilters.getCommitsFilter())
+                            .includeDiff(reportFilters.isIncludeDiff())
+                            .build()
+                    )
+            );
```
tcibinan: I believe the same thing can be achieved by using `reportFilters.toBuilder().archive(false).build()` if we use the following declaration in the original entity class `@Builder(toBuilder = true)`.

```java
results.addAll(prepareDiffsForReportDoc(loaded, gitDiff, reportFilters.toBuilder().archive(false).build()));
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+
+    private void changeHeadersAndFooters(final XWPFDocument document, final Pipeline storage,
+                                         final GitDiff diff, final GitDiffReportFilter reportFilter) {
+        XWPFHeaderFooterPolicy policy = document.getHeaderFooterPolicy();
+        this.changeHeader(policy.getDefaultHeader(), storage, diff, reportFilter);
+        this.changeFooter(policy.getDefaultFooter(), storage, diff, reportFilter);
+        this.changeHeader(policy.getEvenPageHeader(), storage, diff, reportFilter);
+        this.changeFooter(policy.getEvenPageFooter(), storage, diff, reportFilter);
+        this.changeHeader(policy.getOddPageHeader(), storage, diff, reportFilter);
+        this.changeFooter(policy.getEvenPageFooter(), storage, diff, reportFilter);
```
tcibinan: It looks like there is a typo here and `getOddPageFooter` method should be used.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+        this.changeBodyElements(header.getBodyElements(), storage, diff, reportFilter);
+    }
+
+    private void changeFooter(final XWPFFooter footer, final Pipeline storage, final GitDiff diff,
+                              final GitDiffReportFilter reportFilter) {
+        if (footer == null) {
+            return;
+        }
+        this.changeBodyElements(footer.getBodyElements(), storage, diff, reportFilter);
+    }
```
tcibinan: These two methods can be replaced with a single one like this

```java
    private void changeHeaderFooter(final XWPFHeaderFooter headerFooter, final Pipeline storage, final GitDiff diff,
                                    final GitDiffReportFilter reportFilter) {
        if (headerFooter == null) {
            return;
        }
        this.changeBodyElements(headerFooter.getBodyElements(), storage, diff, reportFilter);
    }
```

Also it can be nice to use array parameter type `XWPFHeaderFooter...` to get rid of some code duplication in the call cite.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+     * @param diff - Git diff object to be retrieved for the data
+     * @param reportFilter - Filter object with date, commit information etc, to generate a report
+     */
+    private void changeBodyElements(final List<IBodyElement> getBodyElements, final Pipeline storage,
+                                    final GitDiff diff, final GitDiffReportFilter reportFilter) {
+        int size = getBodyElements.size();
+        for (int i = 0; i < size; i++) {
+            this.changeBodyElement(getBodyElements.get(i), storage, diff, reportFilter);
+            size = getBodyElements.size();
+        }
```
tcibinan: It can be simplified with the following code
```java
elements.forEach(element -> changeBodyElement(element, storage, diff, reportFilter));
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+    private void changeTable(final XWPFTable table, final Pipeline storage,
+                             final GitDiff diff, final GitDiffReportFilter reportFilter) {
+        if (table == null) {
+            return;
+        }
+        for (XWPFTableRow row : table.getRows()) {
+            for (XWPFTableCell cell : row.getTableCells()) {
+                this.changeBodyElements(cell.getBodyElements(), storage, diff, reportFilter);
+            }
+        }
```
tcibinan: It is a matter of taste but it may be nice to write sequences like this using combination of optionals and streams.

```java
        Optional.ofNullable(table).map(Stream::of).orElseGet(Stream::empty)
                .map(XWPFTable::getRows).flatMap(List::stream)
                .map(XWPFTableRow::getTableCells).flatMap(List::stream)
                .map(XWPFTableCell::getBodyElements)
                .forEach(elements -> changeBodyElements(elements, storage, diff, reportFilter));
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+                })
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+    }
+
+    private CommitDiffsGrouping.GroupType getGroupType(final GitDiffReportFilter reportFilter) {
+        if (reportFilter.getGroupType() == null) {
+            return CommitDiffsGrouping.GroupType.BY_COMMIT;
+        }
+        return reportFilter.getGroupType();
```
tcibinan: It's a common `Optional` use case. Let's use it as well.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+        return CommitDiffsGrouping.builder()
+                .includeDiff(reportFilters.isIncludeDiff())
+                .diffGrouping(
+                        getGroupType(reportFilters) == CommitDiffsGrouping.GroupType.BY_COMMIT
+                            ? gitDiff.getEntries().stream()
+                            .collect(Collectors.groupingBy(e -> e.getCommit().getCommit()))
+                            : gitDiff.getEntries().stream()
+                            .collect(Collectors.groupingBy(e -> e.getDiff().getToFileName()))
+                ).build().getDiffGrouping()
+                .entrySet()
```
tcibinan: You can simplify this code by not using `CommitDiffsGrouping` builder at all.

```java
        final Map<String, List<GitDiffEntry>> diffGrouping =
                getGroupType(reportFilters) == CommitDiffsGrouping.GroupType.BY_COMMIT
                ? gitDiff.getEntries().stream().collect(Collectors.groupingBy(e -> e.getCommit().getCommit()))
                : gitDiff.getEntries().stream().collect(Collectors.groupingBy(e -> e.getDiff().getToFileName()));
        return diffGrouping
                .entrySet()
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+                                (getGroupType(reportFilters) == CommitDiffsGrouping.GroupType.BY_COMMIT
+                                        ? REVISION + NAME_SEPARATOR : ""
+                                ) + p.getFirst().replace("/", NAME_SEPARATOR) + DOCX,
+                                report
+                        );
+                    } catch (IOException e) {
+                        log.error(e.getMessage(), e);
+                        return null;
+                    }
+                })
```
tcibinan: I would suggest merging these `map` calls to a single one and extracting its contents to some `toReport` private method. Also it would be nice to extract a few submethods from this new method so the whole structure of it would become something like this:
```java
                    try {
                        final XWPFDocument groupReport = getGroupReport(...);
                        fillTemplate(groupReport, loaded, getGroupDiff(...));
                        return Pair.of(getGroupReportName(...), groupReport);
                    } catch (IOException e) {
                        log.error(e.getMessage(), e);
                        return null;
                    }
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -0,0 +1,339 @@
...
+                        int toDelete = 0;
+                        while (report.getBodyElements().size() != 1) {
+                            IBodyElement element = report.getBodyElements().get(toDelete);
+                            if (element.getElementType() == BodyElementType.PARAGRAPH &&
+                                    ((XWPFParagraph) element).getText()
+                                            .contains(VSReportTemplates.COMMIT_DIFFS.template)) {
+                                toDelete += 1;
+                            }
+                            report.removeBodyElement(toDelete);
+                        }
```
tcibinan: To be honest I don't quite understand what is happening here. Seems like a very tricky and possibly buggy peace of code. It looks like we check element with one index and remove element with the next index. Is it intended?

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/CommitDiffExtractor.java
```diff
@@ -0,0 +1,65 @@
...
+                                        .collect(
+                                                Collectors.groupingBy(
+                                                        e -> e.getDiff().getFromFileName().contains("/dev/null")
+                                                                ? e.getDiff().getToFileName()
+                                                                : e.getDiff().getFromFileName()
+                                                )
+                                    )
+                    );
+        }
+        return diffsGroupingBuilder.build();
```
tcibinan: Let's extract diff grouping creating to a private method and inline everything.

```java
        return CommitDiffsGrouping.builder()
                .type(groupType)
                .includeDiff(reportFilter.isIncludeDiff())
                .archive(reportFilter.isArchive())
                .diffGrouping(reportFilter.isIncludeDiff() ? getDiffGrouping(diff, groupType) : null)
                .build();
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/FileListTableExtractor.java
```diff
@@ -0,0 +1,161 @@
...
+                        ? gitDiffEntry.getDiff().getFromFileName()
+                        : gitDiffEntry.getDiff().getToFileName();
+                return new Pair<>(file, gitDiffEntry.getCommit());
+            }).forEachOrdered(fileAndCommit -> {
+                if (!reported.contains(fileAndCommit.getKey())) {
+                    TableRow row = result.addRow(fileAndCommit.getKey());
+                    tableColumns.forEach((e, v) -> result.setData(row.getName(), v, getData(fileAndCommit, e)));
+                    reported.add(fileAndCommit.getKey());
+                }
+        });
```
tcibinan: This can be simplified with the usage of intermediate map.

```java
        diff.getEntries().stream()
                .collect(Collectors.toMap(this::toFile, GitDiffEntry::getCommit, (c1, c2) -> c1))
                .forEach((file, commit) -> {
                    TableRow row = result.addRow(file);
                    tableColumns.forEach((e, v) -> result.setData(row.getName(), v, getData(file, commit, e)));
                });
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/FileListTableExtractor.java
```diff
@@ -0,0 +1,161 @@
...
+            case AUTHOR:
+                return fileAndCommit.getValue().getAuthor();
+            case DATE_CHANGED:
+                return DATE_FORMAT.format(fileAndCommit.getValue().getAuthorDate());
+            case REVISION:
+                return fileAndCommit.getValue().getCommit().substring(0, 9);
+            case MESSAGE:
+                return fileAndCommit.getValue().getCommitMessage();
+            default:
+                return "";
```
tcibinan: Feels like this functions can be easily extracted to some field of `FileListTableColumn` similarly to `VSReportTemplates`.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/FileListTableExtractor.java
```diff
@@ -0,0 +1,161 @@
...
+        private final String defaultColumn;
+
+        FileListTableColumn(String value, String defaultColumn) {
+            this.value = value;
+            this.defaultColumn = defaultColumn;
+        }
+
+        public static FileListTableColumn byValue(final String value) {
+            return BY_VALUE.get(value);
+        }
```
tcibinan: It looks like this method and the corresponding `BY_VALUE` map are not used anywhere.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/FileListTableExtractor.java
```diff
@@ -0,0 +1,161 @@
...
+    public Object apply(final XWPFParagraph xwpfParagraph, final Pipeline storage, final GitDiff diff, GitDiffReportFilter reportFilter) {
+        Matcher matcher = PATTERN.matcher(xwpfParagraph.getText());
+        final Map<FileListTableColumn, String> tableColumns;
+        if(matcher.matches()) {
+            final String tableStructure = matcher.group(1);
+            tableColumns = parseTableStructure(tableStructure);
+        } else {
+            tableColumns = Arrays.stream(FileListTableColumn.values())
+                    .collect(Collectors.toMap(c -> c, c -> c.defaultColumn));
+        }
```
tcibinan: Extract columns resolving to a private method.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/RevisionListTableExtractor.java
```diff
@@ -0,0 +1,155 @@
+/*
```
tcibinan: Same stuff as in `FileListTableExtractor`.

api/src/main/java/com/epam/pipeline/manager/git/GitReaderApiBuilder.java
```diff
@@ -0,0 +1,111 @@
+/*
```
tcibinan: It seems like this class is not used anywhere. Probably we can delete it.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/AbstractVSReportTemplateProcessor.java
```diff
@@ -0,0 +1,94 @@
+/*
```
tcibinan: Let's get rid of this abstract class.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/VSReportTemplateTableProcessor.java
```diff
@@ -0,0 +1,177 @@
...
+                    } else if (colIndex == -1 && rowIndex >= 0) {
+                        text = table.getRows().get(rowIndex).getName();
+                        xwpfRun.setBold(true);
+                    } else if (rowIndex >= 0 && colIndex >= 0) {
+                        Object cellData = table.getData(rowIndex, colIndex);
+                        text = cellData != null ? cellData.toString() : "";
+                    }
+                    xwpfRun.setText(text, 0);
+                }
+            }
```
tcibinan: This peace of code can be simplified like this

```java
            Assert.isTrue(table.isContainsHeaderRow() && table.isContainsHeaderColumn(),
                    "Both rows and columns headers are not supported.");
            
            if (table.isContainsHeaderRow()) {
                final XWPFTableRow headerRow = xwpfTable.insertNewTableRow(0);
                IntStream.range(0, table.getColumns().size()).forEach(colIndex ->
                        addHeaderRowCell(runTemplate, table, headerRow, colIndex));
                IntStream.range(1, table.getRows().size()).forEach(rowIndex -> {
                    final XWPFTableRow row = xwpfTable.insertNewTableRow(rowIndex);
                    IntStream.range(0, table.getColumns().size()).forEach(colIndex ->
                        addCell(runTemplate, table, rowIndex, row, colIndex));
                });
            }

            if (table.isContainsHeaderColumn()) {
                IntStream.range(0, table.getRows().size()).forEach(rowIndex -> {
                    final XWPFTableRow row = xwpfTable.insertNewTableRow(rowIndex);
                    addHeaderColumnCell(runTemplate, table, rowIndex, row);
                    IntStream.range(1, table.getColumns().size()).forEach(colIndex ->
                            addCell(runTemplate, table, rowIndex, row, colIndex));
                });
            }
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/VSReportTemplateDiffProcessor.java
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+    }
+
+    private void addHeader(final XWPFParagraph paragraph, final String fontFamily, final int fontSize,
+                           final CommitDiffsGrouping.GroupType type, final String groupingKey,
+                           final List<GitDiffEntry> diffEntries) {
+
+        if (type.equals(CommitDiffsGrouping.GroupType.BY_COMMIT)) {
+            insertTextData(paragraph, "In revision ", false, fontFamily, fontSize + 4, false);
```
tcibinan: Let's extract this font increasing int to a class field.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/VSReportTemplateDiffProcessor.java
```diff
@@ -0,0 +1,313 @@
...
+        run.setFontFamily(fontFamily);
+        run.setFontSize(fontSize);
+        run.setBold(bold);
+        run.setText(text);
+        if (lineBreak) {
+            run.addBreak(BreakType.TEXT_WRAPPING);
+        }
+    }
+
+    private void fillTableWithDiffHunk(final String fontFamily, final int fontSize,
```
tcibinan: It's pretty hard to propose anything here but it sure looks like some private methods can be extracted to lighten up the code flow a bit.

tcibinan (commented):
I hope the following comments will improve the overall code readability. In addition to explicit comments I've prepared a list of things that will improve most of the files:

- Use `final` keyword wherever possible.
- Avoid using `this.method()` syntax.
- Replace asterisk imports with explicit ones.
- Remove line breaks before closing brackets in most calls.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
SilinPavel: Not exactly
In order to not to introduce new report template, we clean up existing on and skip only paragraph where `commit_diffs` template exists
regarding to `removeBodyElement` there shouldn't be any unexpected exceptions such as NPE or IndexOutOfBoundException because this method check that `toDelete` index is in bounds of bodyElements array

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineFileGenerationManager.java
SilinPavel: I decided to do it in this way to not to produce additional dependency in PipelineApiService

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
SilinPavel: there is a tricky part, when we call `changeBodyElement` we could even add additional body elements to the document, in this case size will change, that why we need to update it every loop, I add additional comment here, to be a more clear on that

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -17,53 +17,41 @@
...
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+import java.io.*;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
tcibinan: Please don't use asterisk imports. Please do the same in other similar places in this and other classes as well.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -78,131 +66,99 @@
...
+    private String resolveReportName(final Pipeline loaded,
+                                     final List<Pair<String, XWPFDocument>> diffReportFiles) {
+        if (diffReportFiles.size() == 1) {
+            return HISTORY + NAME_SEPARATOR + loaded.getName() + NAME_SEPARATOR
+                            + REPORT_FILE_NAME_DATE_FORMAT.format(DateUtils.now()) + DOCX;
+        } else {
+            return HISTORY + NAME_SEPARATOR + loaded.getName() + NAME_SEPARATOR
+                            + REPORT_FILE_NAME_DATE_FORMAT.format(DateUtils.now()) + ZIP;
+        }
+    }
```
tcibinan: We can get rid of duplication here pretty easily.

```java
    private String resolveReportFileName(final Pipeline loaded,
                                         final List<Pair<String, XWPFDocument>> diffReportFiles) {
        return resolveReportName(loaded) + resolveReportFileExtension(diffReportFiles);
    }
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -78,131 +66,99 @@
...
-                        GitDiffReportFilter.builder()
-                            .groupType(reportFilters.getGroupType())
-                            .commitsFilter(reportFilters.getCommitsFilter())
-                            .includeDiff(reportFilters.isIncludeDiff())
-                            .build()
+                    prepareDiffsForReportDoc(
+                        pipeline, gitDiff,
+                        reportFilters.toBuilder().archive(false).build()
                     )
             );
```
tcibinan: Please remove line breaks before closing brackets as we do mostly in the project. Please do the same in other similar places in this and other classes as well.

```java
            results.addAll(prepareDiffsForReportDoc(pipeline, gitDiff, 
                    reportFilters.toBuilder().archive(false).build()));
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -78,131 +66,99 @@
...
-                                .entries(pair.getSecond())
-                                .filters(gitDiff.getFilters())
-                                .build()
-                        )
-                ).map(p -> {
+                .map(entry -> {
+                    final Pair<String, GitParsedDiff> toReport = Pair.of(
+                            entry.getKey(),
+                            GitParsedDiff.builder().entries(entry.getValue()).filters(gitDiff.getFilters()).build()
+                    );
```
tcibinan: This pair object should be removed in advance to two local variables.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
```diff
@@ -267,15 +204,17 @@ private void changeFooter(final XWPFFooter footer, final Pipeline storage, final
...
     private void changeBodyElements(final List<IBodyElement> getBodyElements, final Pipeline storage,
-                                    final GitDiff diff, final GitDiffReportFilter reportFilter) {
+                                    final GitParsedDiff diff, final GitDiffReportFilter reportFilter) {
         int size = getBodyElements.size();
         for (int i = 0; i < size; i++) {
             this.changeBodyElement(getBodyElements.get(i), storage, diff, reportFilter);
+            // we need to check size every time because after replacing some placeholder
+            // with a value there could be more elements then before
             size = getBodyElements.size();
         }
```
tcibinan: Let's use `while` for it since it will make more sense for a code reader.
```java
        int i = 0;
        while (i < getBodyElements.size()) {
            this.changeBodyElement(getBodyElements.get(i), storage, diff, reportFilter);
            i += 1;
        }
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/VSReportTemplates.java
```diff
@@ -34,24 +37,32 @@
...
-        }
-        return diff.getFilters().getExtensions();
-    })),
+    FILTER_FROM("filter_from",
+        () -> new VSReportTemplateTextProcessor(
+            (paragraph, storage, diff, reportFilter) ->
+                Optional.ofNullable(
+                        diff.getFilters().getDateFrom()
+                ).map(d -> d.format(Constants.DATE_TIME_FORMATTER)
+                ).orElse("-"))
```
tcibinan: Why do you put closing bracket to a start of a line? I believe it is not commonly used and probably we should stick with the usual syntax. Please do the same in other similar places in this and other classes as well.

```java
                    Optional.ofNullable(diff.getFilters().getDateFrom())
                            .map(d -> d.format(Constants.DATE_TIME_FORMATTER))
                            .orElse("-"))
```

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/VSReportTemplates.java
```diff
@@ -34,24 +37,32 @@
...
+    ),
+    FILTER_TYPES("filter_types",
+        () -> new VSReportTemplateTextProcessor(
+            (paragraph, storage, diff, reportFilter) -> {
+                if (CollectionUtils.isNullOrEmpty(diff.getFilters().getExtensions())) {
+                    return "All Files";
+                }
+                return diff.getFilters().getExtensions();
```
tcibinan: Please use optional here as well.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/VSReportTemplateDiffProcessor.java
```diff
@@ -236,29 +228,50 @@ private void fillTableWithDiffHunk(final String fontFamily, final int fontSize,
...
     }
 
+    private String getLineColor(Line line) {
+        if (line.getLineType().equals(Line.LineType.TO)) {
+            return COLOR_GREEN;
+        } else if (line.getLineType().equals(Line.LineType.FROM)) {
+            return COLOR_RED;
+        }
+        return COLOR_WHITE;
+    }
```
tcibinan: Let's use `switch` here.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/VSReportTemplateTableProcessor.java
```diff
@@ -123,51 +156,56 @@ void insertData(XWPFParagraph splittedParagraph, XWPFRun runTemplate, XmlCursor
...
+                final XWPFTableRow row = xwpfTable.insertNewTableRow(rowIndex);
+                IntStream.range(
+                        0, table.getColumns().size() - 1
+                ).forEach(colIndex ->
+                        setCellData(
+                            runTemplate, row, false,
+                            () -> {
+                                final Object cellData = table.getData(rowIndex, colIndex);
+                                return cellData != null ? cellData.toString() : "";
+                            }
```
tcibinan: Why do not use `Optional` here?

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/processors/versionedstorage/processor/VSReportTemplateTableProcessor.java
```diff
@@ -29,92 +29,125 @@
...
+            int end = matcher.end();
+            boolean dataInserted = false;
+            boolean shouldMoveRun = false;
+            int globalStartIndex = 0;
+            int runToRemoveIndex = 0;
+            XWPFParagraph currentParagraph = null;
+            final List<XWPFRun> runs = paragraph.getRuns();
+            final XmlCursor xmlCursor = paragraph.getCTP().newCursor();
+            xmlCursor.toNextSibling();
+            outer: for (XWPFRun run : runs) {
```
tcibinan: Is it possible not to use labels as before?

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineFileGenerationManager.java
tcibinan: It seems like a good intention but from my point of view method `generateVersionStorageReport` feels a little bit odd in `PipelineFileGenerationManager` class. I leave it at your discretion.

api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/VersionStorageReportTemplateManager.java
tcibinan: Ok, I see.

tcibinan (approved):

PR #1941, opened 2021-04-22
issue 1805 Git Reader auth

SilinPavel:
this PR adds jwt auth for git-reader service
It also adds health check for git-reader deployment


deploy/contents/install/install-config
```diff
@@ -193,10 +193,11 @@ CP_SP_INTERNAL_HOST=cp-sensitive-proxy.default.svc.cluster.local
 CP_REDIS_INTERNAL_HOST=cp-redis.default.svc.cluster.local
 CP_REDIS_INTERNAL_PORT=30097
 
-# Other params, specified on command line
-#
-
 # cp-gitlab-reader
 CP_GITLAB_READER_WORKER_COUNT=2
 CP_GITLAB_READER_INTERNAL_PORT=35800
+CP_GITLAB_READER_INTERNAL_HOST=http://cp-gitlab-reader.default.svc.cluster.local
```
mzueva: I think it is better to keep host without scheme as `cp-gitlab-reader.default.svc.cluster.local`

git-reader/gitreader/src/utils/auth_utils.py
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import jwt
+
+
+def auth(request):
```
mzueva: I'd suggest to use Flask HTTPAuth for authorization as it is used in `fsbrowser`

deploy/docker/cp-gitlab-reader/init.sh
```diff
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 cd  $CP_GITLAB_READER_HOME/gitreader/gitreader
+openssl x509 -pubkey -noout -in $CP_API_SRV_CERT_DIR/jwt.key.x509  > $CP_GITLAB_READER_HOME/pub-jwt-key.pem
```
mzueva: Can't we use `$CP_API_SRV_CERT_DIR/jwt.key.public` instead of generation of a new public key?

git-reader/gitreader/src/utils/auth_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    jwt_token = request.headers.get('Authorization', None)
+    if jwt_token:
+        try:
+            payload = jwt.decode(
+                jwt_token.replace("Bearer ", ""),
+                open(
```
mzueva: I think it is better to initialise key once rather then reading it for each request

api/src/main/java/com/epam/pipeline/manager/git/ApiBuilder.java
```diff
@@ -47,17 +47,19 @@
...
     public ApiBuilder(final Class<T> apiClientClass, final String apiHost,
-                      final String adminToken, final String dateFormat) {
+                      final String adminToken, final String bearerToken, final String dateFormat) {
         this.apiClientClass = apiClientClass;
         this.dateFormat = dateFormat;
         this.connectTimeout = TIMEOUT;
         this.readTimeout = TIMEOUT;
         this.apiHost = apiHost;
         this.adminToken = adminToken;
+        this.bearerToken = bearerToken;
```
mzueva: I don't think that we'll ever need two tokens at the same time. Can we switch to `String authHeaderName` and `String token` fields instead of passing two tokens?

mzueva (changes requested):

mzueva (approved):

Issue #1805, opened 2021-02-15
Versioned Storages


NShaforostov:
## Background

In some cases, Cloud Pipeline users wants not only store their data in storages and place computation results there but also have the full value system of the revision control of that data - to view revisions, history of changes, diffs between revisions. Version control in that case can be really useful - because it allows to observe all changes of the data since the original data can сhange quit quickly and often.
So far, for separate storages types (e.g. `AWS s3 buckets`), there is the ability to enable the _versioning_ option. But it is not the same. _Versioning_ allows to manage the versions of the certain file, not the revisions of the full storage, which revision can contain changes of several files or folders.
The similar ability is currently implemented for Pipeline objects - they are GitLab repositories under the hood, all changes can be performed only after commit, etc. We shall implement analogical functionality for data storages.

Users will be able to use such storages during runs but before the computation, the data from that storages shall be cloned into the running instance. After the computation, there will be the ability to commit results back to the storage.

## Approach

Add a new type of storage - **_Versioned Storage_** (further - VS).
Such type of storage should be a wrapper for a GitLab repository.

Users shall have the ability via Web GUI:
- to manage the VS:
  - create a new VS, specify its name and initial structure (empty or predefined folder-tree)
  - edit the VS (rename/remove)
  - manage the permissions on the VS (similar to other Cloud Pipeline objects)
- to read data (files and folders) from the VS:
  - list the VS content, navigate through the folders
  - view file content
  - download files to the local workstation
- to write data to the VS:
  - create files and folders 
  - upload files from the local workstation
  - edit files content, rename files and folders
  - remove files and folders
- to use version control capabilities:
  - each change of the VS content should be accompanied with the corresponding commit and a new revision for the VS (contains only differences with the previous revision)
  - specify the commit message ("comment") on each content change
  - view info of the last revision touched the file/folder in the VS (`SHA-1` hash, date, author, message of the commit)
  - view history of commits of the full VS or separate folder/file
  - generate report of the occurred VS changes (commit history, diffs, etc.) - in `docx` templated format, admins shall have the ability to change the template
- to launch runs with cloning content of the VS into the instance (and use it via any available in Cloud Pipeline running environments - _Remote Desktop_, _SSH_, _Web IDE_, etc.):
  - cloning of the content can be performed in two ways - simultaneously with the run initialization (launch the run directly from the VS) or for the previously initialized run (clone the content to any already launched run)
  - manage the cloned data in the launched instance (add/edit/remove)
[...]

PR #1946, opened 2021-04-29
Versioned storages (GUI)

rodichenko:
This PR implements GUI part for the #1805

### DIFF
--- a/client/src/components/pipelines/browser/Folder.js
+++ b/client/src/components/pipelines/browser/Folder.js
@@ -911,37 +911,42 @@ export default class Folder extends localization.LocalizedReactComponent {
   createVSFolder = async (path, opts) => {
     if (path && opts.lastCommitId) {
       const request = new PipelineFolderUpdate(opts.id);
-      const hide = message.loading(`Creating folders '${path}'...`, 0);
       await request.send({
         lastCommitId: opts.lastCommitId,
         path,
         comment: `creating folders ${path}`
       });
-      hide();
       if (request.error) {
         message.error(request.error, 5);
       } else {
         return request.value;
       }
     }
+    return undefined;
   };
 
   bulkCreateVSFolders = async (foldersStructure, pipeline) => {
     if (!foldersStructure || !foldersStructure.length) {
       return;
     }
-    let nextlastCommitId = pipeline.currentVersion.commitId;
+    let lastCommitId = pipeline.currentVersion.commitId;
     const id = pipeline.id;
     const paths = splitFolderPaths(foldersStructure);
-    for (const path of paths) {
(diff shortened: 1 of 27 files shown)


client/src/components/pipelines/browser/Folder.js
```diff
@@ -894,6 +911,73 @@ export default class Folder extends localization.LocalizedReactComponent {
     await this.createPipeline({name, description, repository, token});
   };
 
+  createVersionedStorage = async ({
+    name,
+    description,
+    repository,
+    pipelineType
```
rodichenko: @AleksandrGorodetskii since this method is used only for *Versioned Storages* creation, we don't need to pass `pipelineType` as an argument of the method - it will always be the same (`VERSIONED_STORAGE`).

client/src/components/pipelines/browser/Folder.js
```diff
@@ -894,6 +911,73 @@ export default class Folder extends localization.LocalizedReactComponent {
...
+      this.props.pipelines.fetch();
+      await this.props.folder.fetch();
+      if (this.props.onReloadTree) {
+        this.props.onReloadTree(true);
+      }
+    }
+  };
+
+  onCreateVersionedStorage = async ({name, description, repository}) => {
+    const pipelineType = 'VERSIONED_STORAGE';
```
rodichenko: @AleksandrGorodetskii we don't need `pipelineType` here: this method only checks the *repository* existence.

client/src/components/pipelines/browser/Folder.js
```diff
@@ -894,6 +911,73 @@ export default class Folder extends localization.LocalizedReactComponent {
...
+        return;
+      }
+    }
+    await this.createVersionedStorage({
+      name,
+      description,
+      repository,
+      pipelineType
+    });
+  }
```
rodichenko: @AleksandrGorodetskii this is mostly the same as `onCreatePipeline` method. We can refactor the code in the following way:
- create (extract logic from `onCreatePipeline`) a new method `checkRepositoryExistance(pipelineOpts, callback)` that checks repository existence and calls `callback(pipelineOpts)` if succeeded;
- use this method for creating a regular pipeline and for the versioned storage:
```javascript
/* pipeline creation modal callback */
<EditPipelineForm
          onSubmit={this.folderOperationWrapper(opts => this.checkRepositoryExistance(opts, this.createPipeline)}
/* versioned storage creation modal callback */
<VersionedStorageDialog
          onSubmit={this.folderOperationWrapper(opts => this.checkRepositoryExistance(opts, this.createVersionedStorage)}
```

rodichenko (commented):
@AleksandrGorodetskii please take a look

Issue #1948, opened 2021-04-30
Allow general users to run a job on behalf of the other account


sidoruka:
**Background**
#1085 introduced support for running jobs on behalf of the other users for `ROLE_ADMIN` only. This feature is widely requested by the `general` users.
We need to implement separate permissions to grant access for specific users/groups rights to "run as" some other users.

**Approach**
* Each user shall get an additional `Launch option` in the profile: `Can run as this user` (default value: `None`)
![image](https://user-images.githubusercontent.com/20998291/116683875-dce4f080-a9b8-11eb-830d-9ef6c392334b.png)
* This option shall allow to select a list of users/groups/roles. Similar to the `Share run`, including the `SSH` option
![image](https://user-images.githubusercontent.com/20998291/116683882-dfdfe100-a9b8-11eb-9b50-4f09f58a77a0.png)
* Any user (further referred as `runner user`), who is explicitly selected for this new option or is a member of a selected group (both internal roles and SAML groups) shall be able to:
  * Run any job on behalf of `another` user (permissions shall be controlled, as if that `another` user was launching on his own)
  * Such jobs shall be automatically shared with the `runner user` (SSH access shall enabled if configured in the `Can run as this user` option)
* The `pipe` CLI functionality from #1085 shall be updated to respect this configuration and allow `general` users to perform `run as`, according to the permissions.

**Example**
* `USER2` is set as `Can run as this user, SSH is allowed` for `USER1`
* `USER2` launches the job via CLI with `--user USER1`
* The job is started and it's configuration looks like:
  * OWNER: `USER1`
  * SHARED WITH: `USER2 with SSH`

rodichenko:
@sidoruka GUI part implemented (#1964)

rodichenko:
@sidoruka @mzueva backported to `stage/0.16/issue_1939_runas` (511d51fa022eaa88db6bacd753e460f296c1353f)

maryvictol:
**Test cases** were created by #2556  and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/Pipe_CLI/1948_launch_run_as_different_user).

maryvictol:
**e2e-tests** were implemented via https://github.com/epam/cloud-pipeline/pull/2556.

maryvictol:
[...]

PR #1961, opened 2021-05-13
Allow general users to run a job on behalf of the other account

ekazachkova:
This PR provides implementation for issue #1948 

A new field `runAs` added to the pipeline run object (`POST /run` method). This value is the user name, which is going to be the `OWNER` of the job.

Also, a new API methods were added:
- `POST users/{id}/runners` with request body:
```
[{
    "name": "<user or role/group name>",
    "principal": true/false,
    "accessType": "ENDPOINT"/"SSH"
}]
```
that saves users or roles, who are allowed to launch pipelines on behalf of the specified user
- `GET users/{id}/runners` that load users/roles, who are allowed to launch pipelines on behalf of the specified user

### DIFF
--- a/api/src/main/java/com/epam/pipeline/controller/user/UserController.java
+++ b/api/src/main/java/com/epam/pipeline/controller/user/UserController.java
@@ -409,12 +409,13 @@ public class UserController extends AbstractRestController {
     @PostMapping("/users/{id}/runners")
     @ResponseBody
     @ApiOperation(
-            value = "Adds runners to user",
-            notes = "Adds runners to user",
+            value = "Updates runners to user",
+            notes = "Updates runners to user",
             produces = MediaType.APPLICATION_JSON_VALUE)
     @ApiResponses(value = {@ApiResponse(code = HTTP_STATUS_OK, message = API_STATUS_DESCRIPTION)})
-    public Result<List<RunnerSid>> addRunners(@PathVariable final Long id, @RequestBody final List<RunnerSid> runners) {
-        return Result.success(userApiService.addRunners(id, runners));
+    public Result<List<RunnerSid>> updateRunners(@PathVariable final Long id,
+                                                 @RequestBody final List<RunnerSid> runners) {
+        return Result.success(userApiService.updateRunners(id, runners));
     }
 
     @GetMapping("/users/{id}/runners")
--- a/api/src/main/java/com/epam/pipeline/controller/user/UserController.java
+++ b/api/src/main/java/com/epam/pipeline/controller/user/UserController.java
@@ -27,6 +27,7 @@ import com.epam.pipeline.entity.user.CustomControl;
 import com.epam.pipeline.entity.user.GroupStatus;
 import com.epam.pipeline.entity.user.PipelineUser;
 import com.epam.pipeline.entity.user.PipelineUserEvent;
+import com.epam.pipeline.entity.user.RunnerSid;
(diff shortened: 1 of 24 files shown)


api/src/main/java/com/epam/pipeline/manager/user/UserRunnersManager.java
```diff
@@ -0,0 +1,99 @@
...
+                .filter(runnerSid -> isRunnerAllowedForUser(runnerSid, userName))
+                .findAny()
+                .orElse(ListUtils.emptyIfNull(allowedRunners).stream()
+                        .filter(runnerSid -> isRunnerAllowedForRoles(runnerSid, user.getRoles()))
+                        .findAny()
+                        .orElse(null));
+    }
+
+    private PipelineUser getUserOrThrow(final Long id) {
+        return Optional.of(pipelineUserRepository.findOne(id)).orElseThrow(() -> new IllegalArgumentException(
```
mzueva: I think `Optional.of` will throw an error in case of `null` argument

api/src/main/java/com/epam/pipeline/manager/user/UserRunnersManager.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private boolean isRunnerAllowedForRoles(final RunnerSid runnersAclSid, final List<Role> roles) {
+        return !runnersAclSid.isPrincipal()
+                && ListUtils.emptyIfNull(roles).stream()
+                .anyMatch(role -> role.getName().equalsIgnoreCase(runnersAclSid.getName()));
+    }
+
+    private void validateRunner(final RunnerSid runnerSid) {
+        Assert.notNull(runnerSid.getName(), messageHelper.getMessage(
```
mzueva: It better to check `isNotBlank`

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -1465,4 +1479,28 @@ private String trimToolVersion(final String docketImage) {
...
+
+        final RunSid runSid = new RunSid();
+        runSid.setName(currentUser.toUpperCase());
+        runSid.setIsPrincipal(true);
+        runSid.setAccessType(allowedAccessType);
+
+        if (CollectionUtils.isEmpty(runSids)) {
+            return Collections.singletonList(runSid);
+        }
+        runSids.add(runSid);
```
mzueva, line 1492: To prevent errors, I'd suggest to return a new collection

api/src/main/java/com/epam/pipeline/controller/user/UserController.java
```diff
@@ -404,4 +405,26 @@ public void exportUsers(@RequestBody PipelineUserExportVO attr, HttpServletRespo
         final MultipartFile file = consumeMultipartFile(request);
         return Result.success(userApiService.importUsersFromCsv(createUser, createGroup, createMetadata, file));
     }
+
+    @PostMapping("/users/{id}/runners")
+    @ResponseBody
+    @ApiOperation(
+            value = "Adds runners to user",
```
mzueva: I'd suggest to call this methods and underlying method in `apiService` - `updateRunners`, as it allows not only add but also remove entities.

mzueva (changes requested):

mzueva (approved):

Issue #1949, opened 2021-04-30
Allow to force "run as" and "sharing" options for the pipelines


sidoruka:
**Background**
Once #1948 is implemented, we shall be able to force certain pipelines to be run on behalf of the other users.

**Approach**
* Add support for the following additional options in the pipeline's configuration (config.json):
  * `run_as` - NameID of the user, which is going to be the `OWNER` of the job
  * `share_with_users` - pipeline job shall be automatically shared with these users, once launched (SSH flag shall configurable as well)
  * `share_with_groups` -  pipeline job shall be automatically shared with these groups (including SAML groups), once launched (SSH flag shall configurable as well)
* These parameters shall not be exposed to the pipeline configuration and launch GUI (at least for now)

maryvictol:
Test case was created by #2151 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/Run_tools_in_sandbox/1949_write_access_to_bucket_with_runas.md).

maryvictol:
**e2e-test** was implemented via https://github.com/epam/cloud-pipeline/pull/2148.

maryvictol:
**Daily e2e-test** results are located and updated [here](https://github.com/epam/cloud-pipeline/blob/e2e-reports/e2e/reports/AWS/test/classes/com.epam.pipeline.autotests.RunAsTest.html) (for `AWS`).

PR #1966, opened 2021-05-17
Allow to force "run as" and "sharing" options for the pipelines - fixes

ekazachkova:
This PR provides fixes for issue #1949


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineConfigurationManager.java
```diff
@@ -205,6 +205,9 @@ public PipelineConfiguration mergeParameters(PipelineStart runVO, PipelineConfig
                 runVO.getWorkerCmd() == null ? defaultConfig.getWorkerCmd() : runVO.getWorkerCmd());
         configuration.setDockerImage(chooseDockerImage(runVO, defaultConfig));
         configuration.buildEnvVariables();
+        configuration.setRunAs(defaultConfig.getRunAs());
```
mzueva: We should merge with `runVO` value here

mzueva (changes requested):

Issue #1713, opened 2021-01-11
Seamless authentication in AWS


sidoruka:
**Backround**
There are two use cases that shall be considered:
* Some of the users are running jobs in the Cloud Pipeline, that access data/services located in the external AWS accounts (e.g. S3, DynamoDB). This requires the user to specify the authentication keys explicitly (either in the shell session or in the R/Python scripts). This is not user-friendly and not secure, if the users include the credentials into the scripts.
* There are also users who would like to leverage R/Python libraries, that have embedded S3 support. This type of workload is not supported at all, users have to download data locally first (via `pipe`) and then perform the processing.

**Approach**
It is suggested to automate the Cloud Provider authentication for the user’s scripts. Users shall be able to execute any request to the Cloud Provider’s API, from inside the Cloud Pipeline environment, without an authentication request.
As a result of this task:
•	Administrator shall be able to configure the user’s access permissions in the Cloud Pipeline AWS account or provide credentials for the external AWS account;
•	All the requests to the AWS authentication will be handled by the Cloud Pipeline "Metadata" service, which will authenticate the user with the configured credentials;
•	Users will be able to use the AWS API without the authentication request.

mzueva:
### Implementation
1. Create new interface `CloudProfileCredentials` with `AWSProfileCredentilals` implementation with the following fields
    - `id` [Long]
    - `cloudProvider` [enum]
    - `profileName` [String]
    - `assumedRole` [String]
    - `policy` [String]
2. `CloudProfileCredentials` shall be stored in DB, all management operations shall be allowed to admins only.
3. Add a new API method to attach `CloudProfileCredentials` to `User` or `Role` entity (many to one), one of profiles shall be optionally marked as `default`. Add a new API method to get profiles associated with a user or a role. These operations shall be allowed to admins only.
4. Add a new API method to generate temporary credentials using `CloudProfileCredentials` `id`. API method shall be allowed to users who have profile `id` attached to them or one of their roles and admins.
5. Add a new field `mountCredentialsRule` to `AbstractCloudRegion` class  with the following allowed values: 
    - `NONE` - do not configure credentials for runs in this region
    - `ALL` - configure all credentials allowed for user
    - `CLOUD` - configure credentials allowed for user for the same cloud provider only
[...]

PR #1985, opened 2021-05-26
Seamless authentication in AWS - fix for empty policy

ekazachkova:
#### Background
`GET /cloud/credentials/generate/{id}` method fails in case if policy is empty string

This PR (connected with issue #1713) provides fix for this situation

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSUtils.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSUtils.java
@@ -33,6 +33,7 @@ import org.apache.commons.lang3.StringUtils;
 public final class AWSUtils {
 
     private static final String EMPTY_KEY_ARN = "NONE";
+    private static final String EMPTY_JSON = "{}";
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSUtils.java
```diff
@@ -68,18 +68,22 @@ public static String getRoleValue(final S3bucketDataStorage storage, final AwsRe
...
+    }
+
     public static TemporaryCredentials generate(final Integer duration, final String policy, final String role,
                                                 final String profile, final String regionCode) {
         final String sessionName = "SessionID-" + PasswordGenerator.generateRandomString(10);
 
         final AssumeRoleRequest assumeRoleRequest = new AssumeRoleRequest()
                 .withDurationSeconds(duration)
-                .withPolicy(policy)
+                .withPolicy(getPolicy(policy))
```
mzueva, line 82: Can we also check for empty json `{}`?

Issue #1715, opened 2021-01-11
Multizone jobs and data access


sidoruka:
**Background**
At the moment, EDGE service can be deployed only in one of the Cloud Regions. While the users and data storages can be located across the globe.
So the users are experiencing the following issues:
•	Higher connection latencies for the interactive sessions (SSH/NoMachine/Web IDEs);
•	Lower data transfer performance when moving data between the regions (e.g. from the EU storage to the US instances).

**Approach**
To address these issues – Cloud Pipeline shall be capable of:
* Running, by default, compute jobs in the region, where the data is located:
  * If the user starts a job/pipeline, that uses EU data – the instance shall be launched in the same EU region as well
  * Default instance region will set according to the locality of the data
* Access the interactive sessions via the "local" endpoints:
  * If the job is run in the EU - the user shall be routed via the EU-hosted EDGE, by default
  * Users shall be able to choose the endpoint URL from an appropriate region as well (e.g. still be able to go through the US EDGE endpoint)

sidoruka:
API:
* Run's ServiceURL - map: 
  * key: region
  * value: current edge-formatted json
* SSH/BROWSE:
  * List EDGE services from different regions (based on label)
  * Get SSH/BROWSE URL's based on the region from a list
* Default EDGE region - in preferences

EDGE:
* Deployment of `N` EDGE instances
* EDGE shall support adding ServiceURLs into array and include region/default information
* `Route53` - create endpoints in corresponding region
* `Custom` domains  (pretty URL) - TBD

GUI:
* Determine current user's region (based on the physical location)
* Choose EDGE endpoints (either "interactive" or SSH/BROWSE) based on the physical location or default region (if a location can't be determined or does not have a corresponding EDGE instance)
* Allow manual selection of the EDGE endpoint

CLI:
* Support multi-regions in `pipe ssh` and `pipe tunnel` - use a default region or allow to override via a command-line option
* Verify any other possible EDGE usage in CLI

maryvictol:
@mzueva, @ekazachkova
Autocloning VS into the running instance at launch a run from the VS page doesn't work.

maryvictol:
@mzueva, @ekazachkova
***Endpoint*** link should be hidden for stopped and paused runs.

maryvictol:
@rodichenko 
Pop up with list of regions should be closed after selection one of them by user.

maryvictol:
> @rodichenko
[...]

PR #2019, opened 2021-06-07
Issue #1715: Multizone jobs and data access - API part (WIP)

ekazachkova:
This PR provides implementation for issue  #1715 

The current PR provides the following changes:
- a new API method `GET /edge/services` added. This method returns all available edge services
- API method `GET /run/<runId>/fsbrowser` updated and returns map (key - edge region, value - fsbrowser url)
- API method `GET /run/<runId>/ssh` aupdated and returns map (key - edge region, value - ssh url)
- API method `POST /run/<runId>/serviceUrl` accepts optional `region` request parameter. If this parameter is not specified default edge region will be used.
- API method `GET /cluster/edge/externalUrl` accepts optional `region` request parameter. If this parameter is not specified default edge region will be used.
- `service_url` DB column removed from `pipeline_run` table. A new table `pipeline_run_service_url` created instead.
- a new preference `default.edge.region` added
- `PipelineRun#serviceUrl` field changed from String to Map<String, String> where key - edge region, value - service url object (the format remains the same)
- `sync-routes.py` requires a new envvar `CP_EDGE_REGION`
- `pipe tunnel start` accepts a new optional option `--region (-rg)`


api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
```diff
@@ -1035,6 +1023,21 @@ private static Array mapListToSqlArray(List<Long> list, Connection connection) {
         return Comparator.comparing(BaseEntity::getId).reversed();
     }
 
+    private Map<String, String> loadServiceUrlByRunId(final Long runId) {
+        return StreamSupport.stream(pipelineRunServiceUrlRepository.findByPipelineRunId(runId).spliterator(), false)
+                .collect(Collectors.toMap(PipelineRunServiceUrl::getRegion, PipelineRunServiceUrl::getServiceUrl));
+    }
+
+    private List<PipelineRun> addServiceUrls(final List<PipelineRun> loadedRuns) {
```
mzueva, line 1037: I'd ask to create a method to load service urls for list of ids

mzueva (changes requested):

deploy/docker/cp-edge/sync-routes.py
```diff
@@ -671,14 +673,28 @@ def update_svc_url_for_run(run_id):
...
+        return None
+
+
 do_log('============ Started iteration ============')
 
 kube_api = HTTPClient(KubeConfig.from_service_account())
 kube_api.session.verify = False
 
-edge_kube_service = Service.objects(kube_api).filter(selector={EDGE_SVC_ROLE_LABEL: EDGE_SVC_ROLE_LABEL_VALUE})
+default_edge_region = find_preference(API_GET_DEFAULT_EDGE_REGION_PREF, 'default.edge.region')
```
SilinPavel, line 690: lets add some warning here that we cant find default edge region, because if `default_edge_region` == None we will never create dns record for any of edge server -> will never to be able to use such link -> it is good to have at least warning in EDGE logs about it and may be we need to log it in run logs as well

deploy/docker/cp-edge/sync-routes.py
```diff
@@ -811,7 +822,8 @@ def update_svc_url_for_run(run_id):
 for added_route in routes_to_add:
         service_spec = services_list[added_route]
 
-        if service_spec["create_dns_record"] and not service_spec["custom_domain"]:
+        if service_spec["create_dns_record"] and not service_spec["custom_domain"] \
+                and default_edge_region == edge_region:
```
SilinPavel: I believe we should look at this `if` closely
so if `service_spec["create_dns_record"] and not service_spec["custom_domain"]` is true but `default_edge_region == edge_region` is false? we will go to `else` branch and create simple `create_service_location` and also update a link for this region 
but as I understand we want create link only for default edge and do nothing for the others

SilinPavel (changes requested):

PR #2027, opened 2021-06-10
issue 1805 fix bug when report can't be built if diffs is empty

SilinPavel:
This PR fixes bug when report can't be built if diffs is empty

1) We assumed that place holders for tables, page breaks, and commit diffs are located in separate paragraph from other text information
2) If it locates in the same paragraph with some text such text will be replaced with content of the placeholder
3) We always creates tables, even if it empty, in this case table with column header will be created


api/src/main/java/com/epam/pipeline/manager/pipeline/documents/templates/versionedstorage/processor/VSReportTemplateDiffProcessor.java
```diff
@@ -110,13 +101,19 @@ void insertData(final XWPFParagraph paragraph, final XWPFRun runTemplate, final
...
 
             addHeader(lastParagraph, fontFamily, fontSize, diffsGrouping.getType(), diffGroupKey, diffEntries);
-            for (GitParsedDiffEntry diffEntry : diffEntries) {
+            for (int i = 0; i < diffEntries.size(); i++) {
+                GitParsedDiffEntry diffEntry = diffEntries.get(i);
                 lastParagraph = addDescription(
-                        lastParagraph, fontFamily, fontSize, diffsGrouping.getType(), diffEntry);
+                        lastParagraph, fontFamily, fontSize, diffsGrouping.getType(),
+                        diffEntry, !isLastEntryInGroup(diffEntries.size(), i));
             }
```
tcibinan: I guess we can slightly simplify this
```java
            for (int i = 0; i < diffEntries.size(); i++) {
                GitParsedDiffEntry diffEntry = diffEntries.get(i);
                lastParagraph = addDescription(lastParagraph, fontFamily, fontSize, diffsGrouping.getType(), diffEntry);
            }
            addIndent(lastParagraph);
```

Issue #2039, opened 2021-06-23
`pipe` CLI shall respect `-e` and `-f` combination


sidoruka:
When using "pipe storage cp -r", the copying is aborted with the message "Flag --force (-f) is required to overwrite files in the destination data." even when "-f" is not necessary, because conflicting files are skipped by -e and/or -s.

This prevents performing bulk copies where specifics files are known to conflict and excluded, while the files of interests are not. This happens: when some files have been added to a source directory and should be copied to a destination directory (that was otherwise unmodified; -s should allow this); when a specific subdirectory should be ignored because it contains machine-specific info that must not be copied, but the rest of the structure should (-e should allow this); when a combination of the above occurs (using both -e and -s should allow this, this is my use-case).
Using the -f option is dangerous in those scenarios, as it allows unintended overwrites causing data loss (e.g. due to typos in the -e filter, or when conflicts should be handled manually so -s was used).

I believe the check for conflicting files and the need for the "--force" option should be performed after the filelist has been filtered by -e and/or -s.

sidoruka:
Shall be cherry-picked to `release/0.16`

PR #2043, opened 2021-06-28
`pipe` CLI shall respect `-e` and `-f` combination

ekazachkova:
The current PR provides fixes introduces in issue #2039 

`-f` flag is not required for copy into existing folder. If the same file exists into source and destination (and shall be copied according to `-e`/`-i`/`-s` options) an error will be occurred in case `-f` is not specified. 

WARN:
The performance may decrease if `-f` flag is not specified since we shall check all destinations

### DIFF
--- a/pipe-cli/src/utilities/datastorage_operations.py
+++ b/pipe-cli/src/utilities/datastorage_operations.py
@@ -73,9 +73,6 @@ class DataStorageOperations(object):
                 click.echo('-n (--threads) is not supported for Windows OS', err=True)
                 sys.exit(1)
             relative = os.path.basename(source) if source_wrapper.is_file() else None
-            can_be_skipped = skip_existing or include or exclude
-            if not force and not destination_wrapper.is_empty(relative=relative) and not can_be_skipped:
-                cls._force_required()
 
             # append slashes to path to correctly determine file/folder type
             if not source_wrapper.is_file():
@@ -150,6 +147,10 @@ class DataStorageOperations(object):
                 if not quiet:
                     click.echo("Skipping file {} since it matches exclude patterns [{}]."
                                .format(full_path, ",".join(exclude)))
+                continue
+
+            if force and not skip_existing:
+                filtered_items.append(item)
                 continue
 
             destination_key = manager.get_destination_key(destination_wrapper, relative_path)
--- a/pipe-cli/src/utilities/datastorage_operations.py
+++ b/pipe-cli/src/utilities/datastorage_operations.py
@@ -91,8 +91,8 @@ class DataStorageOperations(object):
             permission_to_check = os.R_OK if command == 'cp' else os.W_OK
             manager = DataStorageWrapper.get_operation_manager(source_wrapper, destination_wrapper, command)
             items = files_to_copy if file_list else source_wrapper.get_items()
-            items = cls.filter_items(items, manager, source_wrapper, destination_wrapper, permission_to_check,
-                                     include, exclude, force, quiet, skip_existing)
+            items = cls._filter_items(items, manager, source_wrapper, destination_wrapper, permission_to_check,
+                                      include, exclude, force, quiet, skip_existing)
             sorted_items = list()
             transfer_results = []
             for item in items:
@@ -104,14 +104,14 @@ class DataStorageOperations(object):
                 else:
                     transfer_result = manager.transfer(source_wrapper, destination_wrapper, path=full_path,
                                                        relative_path=relative_path, clean=clean, quiet=quiet, size=size,
-                                                       tags=tags, skip_existing=skip_existing)
+                                                       tags=tags)
                     if not destination_wrapper.is_local() and transfer_result:
                         transfer_results.append(transfer_result)
                         transfer_results = cls._flush_transfer_results(source_wrapper, destination_wrapper,
                                                                        transfer_results, clean=clean)
             if threads:
                 cls._multiprocess_transfer_items(sorted_items, threads, manager, source_wrapper, destination_wrapper,
-                                                 clean, quiet, tags, skip_existing)
+                                                 clean, quiet, tags)
             else:
                 if not destination_wrapper.is_local():
                     cls._flush_transfer_results(source_wrapper, destination_wrapper,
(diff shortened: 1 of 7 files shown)


pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -94,45 +92,27 @@ def cp(cls, source, destination, recursive, force, exclude, include, quiet, tags
             permission_to_check = os.R_OK if command == 'cp' else os.W_OK
             manager = DataStorageWrapper.get_operation_manager(source_wrapper, destination_wrapper, command)
             items = files_to_copy if file_list else source_wrapper.get_items()
+            items = cls._filter_items(items, manager, source_wrapper, destination_wrapper, permission_to_check,
```
tcibinan, line 100: Probably the method `_filter_items` can be inlined to have only a single iteration over all items. I think it's better in terms of code readability rather than performance concerns.

pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -141,6 +121,65 @@ def cp(cls, source, destination, recursive, force, exclude, include, quiet, tags
...
+                need_to_overwrite = not manager.skip_existing(source_key, source_size, destination_key,
+                                                              destination_size, quiet)
+                if need_to_overwrite and not force:
+                    cls._force_required()
+                if need_to_overwrite:
+                    filtered_items.append(item)
+                continue
+            if not force:
+                cls._force_required()
+            filtered_items.append(item)
```
tcibinan: Theoretically we can get rid of one HEAD call per file approach in advance to just destination listing. Is it possible for all providers? Do you think it can be a good idea for a new issue?

pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -141,6 +121,65 @@ def cp(cls, source, destination, recursive, force, exclude, include, quiet, tags
...
+
+            destination_key = manager.get_destination_key(destination_wrapper, relative_path)
+            destination_size = manager.get_destination_size(destination_wrapper, destination_key)
+            destination_is_empty = destination_size is None
+            if destination_is_empty:
+                filtered_items.append(item)
+                continue
+            if skip_existing:
+                source_key = manager.get_source_key(source_wrapper, full_path)
+                source_size = manager.get_source_size(source_wrapper, source_key, size)
```
tcibinan: It seems like we can just use `size` here without `manager.get_source_size` method at all.

pipe-cli/src/utilities/datastorage_operations.py
tcibinan: Nah, it's a bad idea. We want to know if `force` flag is required before we start transferring. The current implementation is fine.

pipe-cli/src/utilities/datastorage_operations.py
ekazachkova: done

NShaforostov:
I've tested this new functionality in `AWS` and `Azure` deployments.
Also, the previous functionality was verified.
In general, all works well.

Minor note was encountered in `Azure` deployment:
when performing `cp`/`mv` command from disk to storage or between two storages, the commands are being performed fully and at the end (after some pause) the message appears - "Error: API respond with http status 404."
It looks in the following way:
image

As for the performance, it really decreases on operations with lots of files. See benchmarking (I've been coping/moving different count of files from disk to storage):
|  | Previously |  | Current |  | _Delta_ |
|---|---|---|---|---|---|
|  |  | With **`-f`** |  | With **`-f`** |  |
| 15 files, each 10MB, on 5 hierarchy nesting levels | 0m 7.33s | 0m 8.8s | 0m 8.53s | 0m 6.82s | 1.21s |
| 15 files, each 100MB, on 5 hierarchy nesting levels | 0m 17.91s | 0m 18.69s | 0m 18.77s | 0m 18.69s | 0.86s |
| 150 files, each 10MB, on 10 hierarchy nesting levels | 0m 58.32s | 1m 5.8s | 1m 7.51s | 1m 4.86s | 9.18s |
| 180 files, each 50MB, on 30 hierarchy nesting levels | 2m 12.29s | 2m 27.57s | 2m 27.8s | 2m 13.01s | 15.51s |
| 1500 files, each 1MB, on 10 hierarchy nesting levels | 3m 54.01s | 3m 47.54s | 4m 53.44s | 3m 53.89s | 59.42s |
| 1500 files, each 1MB, on 100 hierarchy nesting levels | 4m 1.89s | 3m 59.8s | 5m 18.23s | 4m 16.1s | 1m 16.34s |
| 1500 files, each 5MB, on 20 hierarchy nesting levels | 5m 10.4s | 5m 51.99s | 6m 23.35s | 5m 23.76s | 1m 12.95s |
| 10500 files, each 1MB, on 1 hierarchy nesting level | 26m 50.17s |  | 33m 44.53s |  | 6m 54.36s |

Where:
**Previously** column - column with `cp`/`mv` command performing duration via the "old" `pipe` version
**Current** column - column with `cp`/`mv` command performing duration via the current `pipe` implementation from the PR
**_Delta_** column - column with delta between **Current** and **Previously** values

sidoruka:
@ekazachkova @mzueva as this new functionality introduces such performance penalty (as @NShaforostov has noted), let's make it optional. I.e.:
* Default behavior of the `pipe storage cp` shall be left untouched (required `-f` and do not analyse the destination contents)
* The new behavior shall be enabled explicitly via `--verify-destination|-vd` flag

So the advanced users, who are aware of such capabilities may enable it for their needs. But the general users won't be affected.

pipe-cli/src/utilities/datastorage_operations.py
tcibinan: Discussed and decided to not implement this due to the extended memory usage and several badly performant scenarios.

mzueva (approved):

NShaforostov:
> * Default behavior of the `pipe storage cp` shall be left untouched (required `-f` and do not analyse the destination contents)
> * The new behavior shall be enabled explicitly via `--verify-destination|-vd` flag
> 
> So the advanced users, who are aware of such capabilities may enable it for their needs. But the general users won't be affected.

I've tested new functionality.
Works well

Issue #1832, opened 2021-02-26
Allow to run Windows-based applications in the Cloud Pipeline


sidoruka:
**Background**
At the moment, Cloud Pipeline is a fully Linux-based solution, but there is a growing demand on running Windows-based apps as well (especially for the `Modelling and Simulation`, `PK/PD` use cases).
There was an attempt (#755) to implement this via the Windows containers, but it failed as only "headless" applications can be run this way. The majority of the relevant apps are GUI-based.
Let's get back to this task and implement it in a different way. The final goal is the same: "Overall, the look and feel of Windows-based apps shall be the same as for the Linux (to the best possible extent)"

**Approach**
* Overall an approach is shown on the diagram below
```
            Baked into the AMI                                                                                           
                    |               Custom AMI per each application                                                      
                    |               Based on Win Server Containers 2019                                Linux             
+-------------------|--------------------------------------------------+              +--------------+ EC2 instances     
| EC2 instance      |                                                  |              | +----------------+               
|                   |                                                  |              | |            |   |               
|                   |                                                  |              | |            |   |               
|  +----------------v----------------+                                 |              +-|------------+   |               
|  |                                 |                                 |                +--------^-------+               
|  |    Business application(s)      |                                 |                         |                       
|  |                                 |                                 |                         |                       
|  +---------------------------------+                                 | Both windows and Linux  |                       
|                                                                      | nodes are connected to  |                       
|  +---------------------------------+    +-------------------------+  | the same cluster        |                       
|  |                                 |    |                         |  |               +---------v--------+              
|  |    Cloud Pipeline services      |    |   Kube services         |  |               |                  |              
|  |                                 |    |   (kubelet, flannel,    <------------------>   Kube master    |              
|  |  +-------------------------+    |    |   canal, proxy)         |  |               |                  |              
|  |  | Cloud Data              |    |    |                         |  |               +------------------+              
|  |  | (preconfigured with     |    |    +-------------------------+  |                                                 
|  |  | the user's credentials) |    |                                 |                                                 
|  |  +-------------------------+    |                                 |                                                 
|  |                                 |    +-------------------------+  |                                                 
|  |  +-------------------------+    |    |                         |  |                                                 
|  |  |                         |    |    |   NoMachine proxy       |  |               +------------------+    User's NX 
|  |  | WebDav-mounted drive Z: |    |    |   (Linux container      |  |               |                  |    connection
[...]

PR #2045, opened 2021-07-01
Add support for automated Windows kube service management

tcibinan:
Relates to #1832.

The pull request brings support for Windows kube service management. From now on a corresponding service will be created for each Windows node which serves a single port. The serving port can be configured by `cluster.kube.windows.service.port` system preference which defaults to 22. Once a Windows node is terminated or scaled down the corresponding service is also shut down.


api/src/main/java/com/epam/pipeline/manager/cloud/CloudFacadeImpl.java
```diff
@@ -290,6 +300,31 @@ private AbstractCloudRegion getRegionByRunId(final Long runId) {
         }
     }
 
+    private void scaleUpNodeServices(final RunInstance instance) {
+        if (KubernetesConstants.WINDOWS.equals(instance.getNodePlatform())) {
```
mzueva: I'd suggest to move this logic to kube-related classes

tcibinan: Done

mzueva (approved):

PR #2046, opened 2021-07-02
Issue #2044 DTS preferences management [API]

Wedds:
This PR is related to issue #2044

It allows specifying some preferences for a data transfer registry. 
- Entity structure updated
- Methods for registry preferences management were implemented and corresponding endpoints were exposed in DtsRegistryController
- New role `DTS_ROLE_MANAGER`
- Expose `data-transfer-service.jar` and stub of a script for running DTS locally into API's static resources
- Required DB migrations created

Issue #2044, opened 2021-06-29
Support autonomous data synchronization in DTS


tcibinan:
**Background**

The following tasks are required to support autonomous data synchronization in DTS.

# @Wedds

@Wedds Could you please implement the following tasks?

**API**
1. [x] Add support for DTS specific preferences
    Add specific preferences field to DTS table and add API methods for DTS specific preferences upsert and delete operations.
1. [x] Expose DTS resources via static files
    Expose data-transfer-service.jar via static files in /deploy/docker/cp-api-srv/init-api and expose DTS launcher script via static files.

**CLI**
1. [x] Add support for `pipe dts create` command
1. [x] Add support for `pipe dts list` command
1. [x] Add support for `pipe dts list [dts id or dts name]` command
    The command shall show all DTS specific preferences as well as common single dts information
1. [x] Add support for `pipe dts update [dts id or dts name] [preference name] [preference value]` command
1. [x] Add support for `pipe dts delete [dts id or dts name] [preference name]` command

**DTS**
1. [x] Add support for DTS specific preferences synchronization
    Create DTS specific preferences watcher which will constantly download all DTS specific preferences and provide them to all java managers in a thread-safe manner. All java managers should not fail if some required DTS specific preferences are missing.
1. [x] Add support for graceful application shutdown on restart=true DTS specific preference
1. [x] Add support for local / cloud directory synchronization
        It should upload all file which were modified since the previous successful synchronization. Several independent synchronization directories should be supported. Synchronization should start by DTS specific cron.

#  @tcibinan

Tasks for myself

**DTS**
1. [x] Embed java to data-transfer-service.jar using gradle

**DTS launcher**
1. [x] Create simple powershell script which downloads DTS + pipe and starts DTS
1. [x] Add support for powershell script self-registration as scheduled task

maryvictol:
**Comments:**

1. [ ] If `DTS_NAME` isn't specified then command `pipe dts preferences update "$DTS_NAME" -p 'dts.local.sync.rules=[....]` gives unclear error message 
```
Traceback (most recent call last):
  File "pipe-cli/pipe.py", line 1804, in <module>
  File "pipe-cli/src/utilities/custom_abort_click_group.py", line 38, in __call__
  File "site-packages/click/core.py", line 697, in main
  File "site-packages/click/core.py", line 1066, in invoke
  File "site-packages/click/core.py", line 1066, in invoke
  File "site-packages/click/core.py", line 1066, in invoke
[...]

PR #2047, opened 2021-07-02
Issue #2044 Implement `dts` group for pipe CLI

Wedds:
This PR is related to issue #2044 

It brings new `dts` group into pipe CLI, consist of the following commands:
- `create` - to register a new DTS
- `list` <registry_id> - to receive information about all of the registered DTS or a specific one

`preferences` subgroup with 2 commands:
- `update` - to update preferences for an existing DTS
- `delete` to remove certain preferences from DTS

All of the commands support output in a pretty format as a table (default behavior) or JSON-formatted (if `--json-out`/`-jo` flag is used)

Depends on PR #2046

### DIFF
--- a/pipe-cli/src/api/dts.py
+++ b/pipe-cli/src/api/dts.py
@@ -1,0 +1,74 @@
+# Copyright 2017-2021 EPAM Systems, Inc. (https://www.epam.com/)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .base import API
+from ..model.dts_model import DtsModel
+
+import json
+
+
+class DTS(API):
+
+    def __init__(self):
+        super(DTS, self).__init__()
+
+    @classmethod
+    def create(cls, url, name, schedulable, prefixes, preferences):
+        api = cls.instance()
+        if not prefixes:
+            prefixes = []
+        if not preferences:
+            preferences = {}
+        dts = {
+            'url': url,
+            'name': name,
+            'schedulable': schedulable,
+            'prefixes': prefixes,
+            'preferences': preferences
+        }
+        response = api.call('/dts', json.dumps(dts), http_method='post')
+        return DtsModel().load(response['payload'])
+
+    @classmethod
+    def load_all(cls):
+        api = cls.instance()
+        response = api.call('/dts', None)
+        registries = []
+        for registry_json in response['payload']:
+            registries.append(DtsModel().load(registry_json))
+        return registries
+
+    @classmethod
+    def load(cls, registry_id):
+        api = cls.instance()
+        response = api.call('/dts/{}'.format(registry_id), None)
+        return DtsModel().load(response['payload'])
+
+    @classmethod
+    def update_preferences(cls, registry_id, preferences):
+        api = cls.instance()
+        data = {
+            'preferencesToUpdate': preferences
+        }
+        response = api.call('/dts/{}/preferences'.format(registry_id), json.dumps(data), http_method='put')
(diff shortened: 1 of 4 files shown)


pipe-cli/src/api/dts.py
```diff
@@ -0,0 +1,74 @@
...
+        if not preferences:
+            preferences = {}
+        dts = {
+            'url': url,
+            'name': name,
+            'schedulable': schedulable,
+            'prefixes': prefixes,
+            'preferences': preferences
+        }
+        response = api.call('/dts', json.dumps(dts), http_method='post')
```
tcibinan, line 40: Could you please use `api.retryable_call` in all similar calls in the class?

pipe-cli/src/api/dts.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self):
+        super(DTS, self).__init__()
+
+    @classmethod
+    def create(cls, url, name, schedulable, prefixes, preferences):
+        api = cls.instance()
+        if not prefixes:
+            prefixes = []
+        if not preferences:
+            preferences = {}
```
tcibinan, line 32: We can use the following style for similar calls.

```python
prefixes = prefixes or []
preferences = preferences or {}
```

pipe-cli/src/model/dts_model.py
```diff
@@ -0,0 +1,49 @@
...
+        dts = cls()
+        dts.id = json['id']
+        dts.name = json['name']
+        dts.url = json['url']
+        dts.created_date = date_utilities.server_date_representation(json['createdDate'])
+        dts.schedulable = json['schedulable']
+        if 'prefixes' in json:
+            dts.prefixes = json['prefixes']
+        if 'preferences' in json:
+            dts.preferences = json['preferences']
```
tcibinan, line 41: Similarly here.
```python
dts.prefixes = json.get('prefixes')
dts.preferences = json.get('preferences')
```

pipe-cli/src/utilities/dts_operations_manager.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def _convert_preferences_list_to_dict(self, preferences_list):
+        preferences_dict = {}
+        for preference_entry in preferences_list:
+            preference_value_and_key = preference_entry.split(self._PREF_DELIMITER)
+            if len(preference_value_and_key) != 2:
+                click.echo('Error [%s]: preference should contain exactly one delimiter!' % preference_entry, err=True)
+                sys.exit(1)
+            else:
+                preferences_dict[preference_value_and_key[0]] = preference_value_and_key[1]
```
tcibinan, line 69: Let's split by the first delimiter occurrence and as a result support values with `self._PREF_DELIMITER` in it.

```python
preference_value_and_key = preference_entry.split(self._PREF_DELIMITER, 1)
```

pipe-cli/src/utilities/dts_operations_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    def _init_table(self):
+        table = prettytable.PrettyTable()
+        table.field_names = self._DTS_TABLE_HEADERS
+        table.align = "l"
+        table.header = True
+        return table
+
+    def _convert_registry_to_prettytable_row(self, registry):
+        flat_preferences = []
+        for preference in registry.preferences.items():
```
tcibinan, line 96: Let's use the following syntax

```python
for preference, value in registry.preferences.items():
```

pipe-cli/src/utilities/dts_operations_manager.py
```diff
@@ -0,0 +1,99 @@
...
+        click.echo(table)
+
+    def _init_table(self):
+        table = prettytable.PrettyTable()
+        table.field_names = self._DTS_TABLE_HEADERS
+        table.align = "l"
+        table.header = True
+        return table
+
+    def _convert_registry_to_prettytable_row(self, registry):
```
tcibinan, line 94: Let's show preferences only if a single registry is shown. Additionally let's use slightly different output view for single registry. As a result it would be nice to have something similar to how `pipe view-tools` behaves.

```
> pipe dts list
Id Name CreatedDate
1 dts1 date1
2 dts2 date2

> pipe dts list dts1
Id: 1
Name: dts1
Url: url1
Created date: date
Schedulable: true
Prefixes:
prefix1
prefix2
Preferences:
preferences1=value1
preferences2=value2
```

tcibinan:
Decided to make the review changes as a separate pull request.

Issue #2023, opened 2021-06-09
Metadata control improvements


kamyshova:
## **Background**
At the moment, there are many new metadata features in Cloud Pipeline thanks to the #1965 issue. It would be great to extend control methods even further to manage metadata entities:

- **column sorting extension**
   Sorting multiple columns independently.
- **metadata entities display setting**
   Pre-configuring the display order and visibility of metadata columns.
- **metadata entities autofill**
  Filling metadata entity with data that are based on data in other entity automatically.
- **resolve variables for rerun**
  Resolving of default environment variables for a rerun.
- **samples preselect for a rerun**
  Preselecting the samples to modify sample information for a rerun.

### Column sorting extension

Currently, the users can combine column filters with each other in an applied order. It would be convenient for users to combine column sorting in the same way. 

**Acceptance criteria:**

- Column sortings should be combined with each other and filters in an applied order
- If the column sorting is set, this should be visible to the user (e.g. the sort icon will be highlighted)

### Metadata entities display setting
Currently, the users can upload metadata from the CSV or TSV file. Attributes of the uploaded metadata object will be displayed in columns. If the uploaded sample sheet has a large number of columns, a lot of columns will displayed on the UI that will confuse users. It would be convenient to add a new control to pre-configuring the display order and visibility of metadata columns for a large number of columns.

**Acceptance criteria:**

- `Show columns` control should be extended in the Metadata entity space
  - the control should be shown automatically after uploading a metadata object from a file
  - the control should contain a list of selectable metadata columns from a file to display
  - a user shall be able to order the metadata columns to display as they see fit
- Only selected in the `Show columns` control metadata columns should be shown in the main table
- A user shall be able to modify a list of metadata columns to display in the `Show columns` control
- A user shall be able to order the metadata columns as they see fit
- The display order and visibility of columns should be preserved when leaving the metadata page

### Metadata entities autofill
Currently, it is possible to add `key:value` pairs to an existing sample in the metadata table. It would be convenient to add an autofill feature to fill metadata entities with data that are based on data in other entities in a column. For example, a user wants to add a new `key:value` pair `BCL:BCL001` to a filtered subset of samples. It would be convenient to apply this to all samples in this BCL column by clicking on a single cell and drag the fill handle on the rest of the column.

**Acceptance criteria:**

- A new autofill control should be introduced in the Metadata entity space
  - the fill handle should allow dragging the value of a cell to other cells
  - the autofill should only be allowed within one column
[...]

PR #2050, opened 2021-07-07
Fix for metadata autofill: Cells aren't filled exactly following of pattern if the value's numeric part starts with 0

ZMaratovna:
This PR implements fix issue 'Cells aren't filled exactly following of pattern if the value's numeric part starts with 0' for (#2023)


ZMaratovna:
Close due to wrong branch

PR #2055, opened 2021-07-08
GUI fix: remove PAUSE  and LOCAL STORAGES links for windows-tool

ZMaratovna:
This PR implements fix for #1832

### DIFF
--- a/client/src/components/main/home/panels/components/CardsPanel.js
+++ b/client/src/components/main/home/panels/components/CardsPanel.js
@@ -177,40 +177,43 @@ export default class CardsPanel extends React.Component {
             type="actions-container-background"
             className={styles.actionsContainerBackground} />
           {
-            actions.map((action, index, array) => {
-              return (
-                <Row
-                  type="flex"
-                  justify="start"
-                  align="middle"
-                  key={index}
-                  className={styles.actionButton}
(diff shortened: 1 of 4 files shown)


client/src/components/main/home/panels/components/CardsPanel.js
```diff
@@ -177,40 +177,43 @@ export default class CardsPanel extends React.Component {
...
-                          <Popover
-                            onVisibleChange={onVisibleChange}
-                            content={action.overlay}>
-                            <span style={action.style}>{action.title}</span>
-                          </Popover>
-                        )
-                        : <span style={action.style}>{action.title}</span>
-                    }
+            actions
+              .filter((action) => (action.title !== 'PAUSE' && action.icon !== 'fork') ||
```
rodichenko: We shouldn't rely on **icon** and **title** while rendering run actions (as they may change); instead we should consider run's **platform** when we create **actions** array and _do not add_ **pause** action for windows-based runs

rodichenko (changes requested):

client/src/components/main/home/panels/components/CardsPanel.js
ZMaratovna: Done!

rodichenko (approved):

PR #2057, opened 2021-07-09
Issue 2044 Autonomous data synchronization service

Wedds:
This PR is related to issue #2044 

It brings the implementation of autonomous synchronization:
- new `AutonomousSynchronizationService` wrapping  `TransferService` and execute synchronization based on active rules
- Sync rules are pulled from the preference, containing serialized `List<AutonomousSyncRule>`
- Separate pool is used for autonomous transfers


data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/DataUploaderProviderManager.java
```diff
@@ -32,8 +32,17 @@
...
+    }
+
+    @Async("autonomousTransferExecutor")
+    public void transferLocalData(final TransferTask transferTask) {
+        executeTransfer(transferTask);
+    }
+
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
-    public void transferData(TransferTask transferTask) {
+    private void executeTransfer(final TransferTask transferTask) {
```
tcibinan: What do you think if we extract `DataUploaderProviderManager` interface and create three implementations:
- `SyncDataUploaderProviderManager`
- `AsyncDataUploaderProviderManager`
- `AutonomousAsyncDataUploaderProviderManager`

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/model/AutonomousSyncRule.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.epam.pipeline.dts.transfer.model;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
```
tcibinan: Let's use `@Value` here.

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/TransferService.java
```diff
@@ -24,9 +24,16 @@
...
+                                         @NonNull StorageItem destination,
+                                         List<String> included) {
+        return runTransferTask(source, destination, included, false);
+    }
+
     TransferTask runTransferTask(@NonNull StorageItem source,
                                  @NonNull StorageItem destination,
-                                 List<String> included);
+                                 List<String> included,
+                                 boolean isAutonomousTransfer);
```
tcibinan: I believe we can get rid of this additional method if we configure another bean of class `TransferServiceImpl` which will be used in `AutonomousSynchronizationService`.

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/impl/DtsSynchronizationService.java
```diff
@@ -69,10 +81,27 @@ public void synchronizePreferences() {
...
+    private void updatePreferences(final Map<String, String> updatedPreferences) {
         preferences.clear();
         preferences.putAll(updatedPreferences);
     }
 
+    private void sendSyncRules() {
+        final String rulesAsString = preferences.getOrDefault(dtsLocalSyncRulesKey, "[]");
+        try {
+            final List<AutonomousSyncRule> rules = new ObjectMapper()
+                .readValue(rulesAsString, new TypeReference<List<AutonomousSyncRule>>() {});
```
tcibinan: Let's move this code to `PreferenceManager.getSyncRules` and poll sync rules right from `AutonomousSynchronizationService`.

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/impl/AutonomousSynchronizationService.java
```diff
@@ -0,0 +1,155 @@
...
+        this.transferService = transferService;
+        this.activeSyncRules = ConcurrentHashMap.newKeySet();
+        this.activeTransferTasks = new HashMap<>();
+    }
+
+    @Scheduled(cron = "${dts.autonomous.sync.cron}")
+    public void synchronizeFilesystem() {
+        processActiveTasks();
+        submitTasksForAwaitingRules();
+    }
```
tcibinan: We need to synchronise directories by cron specified in DTS specific preferences. To do so we can use f.e. `CronSequenceGenerator` to calculate next cron time. With the next cron local calculated we can just check if the time has come and only when start new transfers.

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/impl/DtsSynchronizationService.java
```diff
@@ -17,8 +17,11 @@
 package com.epam.pipeline.dts.transfer.service.impl;
```
tcibinan: What do you think about merging `DtsSynchronizationService` and `AutonomousSynchronizationService`? The distinction between them doesn't seem quiet obvious.

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/impl/AsyncDataUploaderProviderManager.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.epam.pipeline.dts.transfer.model.TransferTask;
+import com.epam.pipeline.dts.transfer.service.DataUploaderProvider;
+import com.epam.pipeline.dts.transfer.service.TaskService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Service;
+
+@Service("commonDataUploaderProviderManager")
+public class AsyncDataUploaderProviderManager extends SyncDataUploaderProviderManager {
```
tcibinan: It's usually better to use composition over inheritance.

```java
@Service("commonDataUploaderProviderManager")
public class AsyncDataUploaderProviderManager implements DataUploaderProviderManager {

     private final DataUploaderProviderManager syncDataUploaderProviderManager;
     
     @Async("taskExecutor")
     @Override
     public void transferData(final TransferTask transferTask) {
         syncDataUploaderProviderManager.transferData(transferTask);
     }
}
```

data-transfer-service/src/main/java/com/epam/pipeline/dts/transfer/service/impl/AutonomousAsyncDataUploaderProviderManager.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.epam.pipeline.dts.transfer.model.TransferTask;
+import com.epam.pipeline.dts.transfer.service.DataUploaderProvider;
+import com.epam.pipeline.dts.transfer.service.TaskService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Service;
+
+@Service("autonomousDataUploaderProviderManager")
+public class AutonomousAsyncDataUploaderProviderManager extends SyncDataUploaderProviderManager{
```
tcibinan: Same here.

data-transfer-service/src/main/java/com/epam/pipeline/dts/sync/service/impl/DtsSynchronizationService.java
```diff
@@ -16,26 +16,244 @@
...
+                                     final TaskRepository taskRepository,
+                                     final PreferenceService preferenceService,
+                                     final ShutdownService shutdownService) {
+        this.pipeCredentials = new PipelineCredentials(pipeApiUrl, pipeApiToken);
+        this.taskRepository = taskRepository;
+        this.transferService = autonomousTransferService;
+        this.shutdownService = shutdownService;
+        this.preferenceService = preferenceService;
+        this.activeSyncRules = new ConcurrentHashMap();
+        this.activeTransferTasks = new HashMap<>();
```
tcibinan: As long as we have a method being executed by schedule then several threads can possibly access this map. I don't really think it may cause any issues here since we don't have any concurrent updates. But still I would use `ConcurrentHashMap` here just to be sure.

tcibinan (approved):

PR #2061, opened 2021-07-13
GUI multizone selector

ZMaratovna:
This PR implements  selector for multizone support


client/src/components/runs/logs/Log.js
```diff
@@ -1368,17 +1368,22 @@ class Logs extends localization.LocalizedReactComponent {
...
             <td>
               <ul>
                 {
-                  urls.map((url, index) =>
-                    <li key={index}><a href={url.url} target="_blank">{url.name || url.url}</a></li>
+                  regionedUrls.map((url, index) =>
+                    <li key={index}>
+                      <RegionsDropdown
+                        title={url.name || url.url}
+                        regions={Object.assign({}, ...regionedUrls.map(({url}) => url))}
```
rodichenko: `Object.assign(...)` will return wrong `region`:`url` configuration - it will join **all** configurations for all endpoints instead of the configuration for particular one

client/src/components/runs/logs/Log.js
```diff
@@ -1368,17 +1368,22 @@ class Logs extends localization.LocalizedReactComponent {
...
+                        run={this.runPayload}
+                        defaultRegion={regionedUrls.map(url => this.props.multiZone.getDefaultURLRegion(url.url))[index]}
```
rodichenko: We don't need `regionedUrls.map` here - we already have `url` (line `1378`), which is an object of `{name: string, url: {region: url, region: url, ...}}` - so we can get default region simply by `this.props.multiZone.getDefaultURLRegion(url.url)`

client/src/components/runs/logs/RegionsDropdown.js
```diff
@@ -0,0 +1,76 @@
...
+    >
+      <Icon type="down" style={{color: '#108ee9', marginLeft: 5}} />
+    </Dropdown>
+  </div>);
+}
+RegionsDropdown.propTypes = {
+  regions: PropTypes.object,
+  defaultRegion: PropTypes.string,
+  title: PropTypes.string,
+  run: PropTypes.object
```
rodichenko: `RegionsDropdown` shouldn't accept `run` object - instead, it must accept *only*:
- **regions** - an object of `{regionA: url, regionB: url, ...}` format;
- **defaultRegion** - string;
- **title**

This component should render `AWSRegionTag` for **every** `region` in **regions** (look at `regionUID` property)

client/src/components/runs/logs/RegionsDropdown.js
```diff
@@ -0,0 +1,76 @@
...
+import PropTypes from 'prop-types';
+
+import AWSRegionTag from '../../special/AWSRegionTag';
+import {
+  Dropdown,
+  Menu,
+  Icon
+} from 'antd';
+
+export function RegionsDropdown (props) {
```
rodichenko: Rename this component to `MultizoneUrl`

rodichenko (changes requested):

client/src/components/runs/logs/RegionsDropdown.js
ZMaratovna: @rodichenko  Done!

client/src/components/runs/logs/RegionsDropdown.js
ZMaratovna: @rodichenko  Done!

client/src/components/runs/logs/Log.js
ZMaratovna: @rodichenko  Done!

client/src/components/runs/logs/Log.js
ZMaratovna: @rodichenko Done!

rodichenko (approved):

PR #2080, opened 2021-07-22
Support heartbeat in data transfer service

tcibinan:
Relates to #2044.

The pull request brings support for data transfer service heartbeat.

## System preferences

Several new system preferences are introduced.

- `dts.monitoring.period.seconds` configures how often API should monitor data transfer services in seconds. By default it does it every minute.
- `dts.offline.timeout.seconds` configures a timeout in seconds after which data transfer service is considered offline if no heartbeats were sent during that period. By default timeout is five minutes.

## Observation

The following pipe cli command can be used to check status of all data transfer services.

```bash
pipe dts list

# +----+------------+---------+
# | ID | Name       | Status  |
# +----+------------+---------+
# | 1  | firstdts   | ONLINE  |
# +----+------------+---------+
# | 2  | seconddts  | OFFLINE |
# +----+------------+---------+
```

## Configuration

_Notice that heartbeat is disabled in data transfer service by default._ To enable heartbeat the following preference should be set for each individual data transfer service: `dts.heartbeat.enabled`. This can be done using the following cli command.

```bash
pipe dts preferences update dtsname -p dts.heartbeat.enabled=true
```

## Related changes

Besides the introduction of data transfer service heartbeat the following smaller changes are also covered by this pull request:
- Disable support for registering two data transfer services with the same name in api.
- Support `pipe dts delete` command in cli.
- Support `--trace` flag in all `pipe dts` commands in cli.
- Update `--help` output for all `pipe dts` commands in cli.

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -1858,13 +1858,13 @@ def update_dts_preferences(registry_name_or_id, preference, json_out, trace):
 
 @preferences.command(name='delete')
 @click.argument('registry-name-or-id', required=True, type=str)
-@click.option('--key', '-k', multiple=True, type=str,
+@click.option('--preference', '-p', multiple=True, type=str,
               help="Key of the preference, that should be removed. Multiple options supported")
 @click.option('--json-out', '-jo', required=False, is_flag=True, help='Defines if output should be JSON-formatted')
 @click.option('--trace', required=False, is_flag=True, default=False, help=TRACE_OPTION_DESCRIPTION)
 @Config.validate_access_token
 @stacktracing
-def delete_dts_preferences(registry_name_or_id, key, json_out, trace):
+def delete_dts_preferences(registry_name_or_id, preference, json_out, trace):
     """
     Deletes preferences for the given data transfer service.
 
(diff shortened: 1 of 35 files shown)


mzueva (approved):

pipe-cli/pipe.py
```diff
@@ -1753,28 +1752,83 @@ def dts():
...
     DtsOperationsManager().list(registry_name_or_id, json_out)
 
 
+@dts.command(name='delete')
+@click.argument('registry-name-or-id', required=True, type=str)
+@click.option('--json-out', '-jo', required=False, is_flag=True, help='Defines if output should be JSON-formatted')
+@click.option('--trace', required=False, is_flag=True, default=False, help=TRACE_OPTION_DESCRIPTION)
+@Config.validate_access_token
+@stacktracing
+def create_dts(registry_name_or_id, json_out, trace):
```
Wedds: Looks like a typo mistake

tcibinan: Thanks, fixed.

Wedds (approved):

Issue #2086, opened 2021-07-22
The OWNER of the Filesystem storage shall be able to convert it into a Versioned Storage


sidoruka:
**Background**
Some of the datasets/scripts stored in the Cloud Pipeline environment have the following lifecycle:
* Created/Uploaded into the Filesystem storage
* For some time, development occurs in this kind of a "plain" filesystem
* At some stage, users want to put the data under a version control

It is possible to perform this operation manually (e.g. mount a filesystem and push its contents to the VS manually). But this is a complicated task. It shall be fully automated.

**Approach**
* `OWNER` of the Filesystem storage shall have a `Convert to Versioned Storage` button in the storage details GUI
* Once this button is clicked - user shall be notified, that all the data will be moved under a version control and this operation cannot be undone
* API shall support a new operation to:
  * Create a new VS:
    * Shall be named as the original filesystem storage
    * Shall be stored in the same parent folder
    * Shall have the same description/metadata/permissions/attributes as the original storage
  * Copy all the data from the filesystem to the VS in a single commit
  * Delete the original filesystem storage according to the new preference:
    * Either unregister a storage
    * Fully delete

rodichenko:
61206964e363e64e64aba0d73d5a636750058cfb implements GUI part: "Convert to Versioned Storage" action is available under "Edit" menu for *non-empty* NFS storages (see https://github.com/epam/cloud-pipeline/pull/2107#issue-701657231)

PR #2107, opened 2021-08-02
Support data storage to versioned storage conversion

tcibinan:
Resolves issue #2086.

The pull request brings support for data storage to versioned storage conversion. *Notice* that currently only nfs storages are supported.

Single system preference was added in order to configure default conversion behaviour:

- `storage.convert.source.action` specifies default action to perform on source data storage once conversion is done. Defaults to `LEAVE`. Supported values are the following:
  - `LEAVE` keeps the source data storage intact,
  - `UNREGISTER` keeps the source data storage in cloud but deletes it from Cloud Pipeline,
  - `DELETE` deletes the source data storage from both cloud and Cloud Pipeline.

The following API request with no body can be used to convert a data storage to a versioned storage.

```
POST /datastorage/{id}/convert
```


mzueva (approved):

tcibinan:
Cherry-picked `NFSStorageMounter` class to `release/0.16` via 54ad7bf2ea9d8674723a76eaf62dca6b82f6f566.

Issue #2110, opened 2021-08-04
Ability to restrict docker images list to which the storage can be mounted


NShaforostov:
### Background

Currently, the list of storages that will be mounted to the running instance can be restricted via "limit mounts" option - it can be performed via the tool settings or exactly before the run launch via custom settings.
But in some cases, it is necessary to restrict the list of tools (docker images) to which the specific storage will be mounted.
For example, we want the specific user home storage can be mounted to only single image - `library/ubuntu`, e.g. It would be quite difficult and time consuming to tweak the "limit mounts" option for all the other tools in the Docker registry to exclude the desired storage from their allowable mounts.

We should add the ability to set the "mount-allowable" docker images list for any storage. If this list is set for the storage, it can be mounted to only docker images from that list.

### Approach
 
Add to the storage settings pop-up (to the "Info" tab) a new control "**Allow mount to**", e.g.:
  ![image](https://user-images.githubusercontent.com/45459424/128167595-4040314e-67c2-42e7-8971-a920b0c545c5.png)

In this field, the admin or storage owner can specify to which images the mount of the storage is allowable.
By default, the mount shall be allowable to all available images.

To restrict the list of images allowable to mount:
- user should click the field near the "**Allow mount to**" label
- a new pop-up shall appear that contains a list of all docker images available to the user, e.g.:
  ![image](https://user-images.githubusercontent.com/45459424/128177294-4dc12c67-9346-4867-9971-c0eb6ba790d3.png)
  This pop-up shall contain:
  - checkbox to allow mounting the storage to any docker images (enabled by default)
  - checkbox to enable the form where user can select docker images to which only current storage can be mounted (these checkboxes are mutually exclusive)
  - buttons to save/cancel changes
- user should enabled "**Limit mounts**" checkbox, the form with docker images selection becomes enabled too:
  ![image](https://user-images.githubusercontent.com/45459424/128197700-855d31ed-f9f5-4a5f-ada5-811b10e8e241.png)
  This form shall contain:
  - list of all docker images available to the user
  - checkbox near each docker image
  - search bar to search a specific image
  - button "Select all" - to select all images in the list
  - button "Clear selection" - to unselect all selected images
  - near each selected image, there should be a dropdown list with all image versions
- user should select any desired docker image(s) (or none)
- for each selected docker image, user can select version(s) (at least one shall be selected), e.g.:
[...]

PR #2116, opened 2021-08-09
Issue 2110 restrict mount of storage for tools

SilinPavel:
Relates to #2110.

Adds possibility to restrict for the mount storage to specific image or even version of an image

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/datastorage/DataStorageDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/datastorage/DataStorageDao.java
@@ -26,10 +26,14 @@ import com.epam.pipeline.entity.datastorage.aws.S3bucketDataStorage;
 import com.epam.pipeline.entity.datastorage.azure.AzureBlobStorage;
 import com.epam.pipeline.entity.datastorage.gcp.GSBucketStorage;
 import com.epam.pipeline.entity.pipeline.Folder;
+import com.epam.pipeline.entity.pipeline.ToolFingerprint;
+import com.epam.pipeline.entity.pipeline.ToolVersionFingerprint;
 import com.epam.pipeline.entity.utils.DateUtils;
+import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Required;
+import org.springframework.data.util.Pair;
 import org.springframework.jdbc.core.ResultSetExtractor;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
@@ -39,14 +43,17 @@ import org.springframework.transaction.annotation.Transactional;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.epam.pipeline.dao.DaoHelper.POSTGRES_LIKE_CHARACTER;
 
@@ -75,6 +82,9 @@ public class DataStorageDao extends NamedParameterJdbcDaoSupport {
     private String loadDataStoragesFileShareId;
     private String loadDataStorageRootQuery;
     private String createDataStorageRootQuery;
+    private String loadToolsToMountQuery;
+    private String deleteToolsToMountQuery;
+    private String addToolVersionToMountQuery;
 
     @Autowired
     private DaoHelper daoHelper;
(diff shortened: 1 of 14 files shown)


api/src/main/java/com/epam/pipeline/dao/tool/ToolVersionDao.java
```diff
@@ -121,6 +125,14 @@ public void updateToolVersionWithSettings(ToolVersion toolVersion) {
                 .collect(Collectors.toMap(ToolVersion::getVersion, Function.identity()));
     }
 
+    public void setDeleteDatastorageToolVersionQuery(String deleteDatastorageToolVersionQuery) {
```
tcibinan: Could you please move these setters closer to other ones?

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -1025,6 +1045,32 @@ private void checkDataStorageObjectExists(final AbstractDataStorage dataStorage,
                         path, dataStorage.getRoot()));
     }
 
+    private void assertToolsToMount(DataStorageVO dataStorageVO) {
+        if (!CollectionUtils.emptyIfNull(dataStorageVO.getToolsToMount()).isEmpty()) {
```
tcibinan: Let's just use `CollectionUtils.isNotEmpty(dataStorageVO.getToolsToMount())` here. It's null safe as well.

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -1025,6 +1045,32 @@ private void checkDataStorageObjectExists(final AbstractDataStorage dataStorage,
...
+            for (ToolFingerprint tool : dataStorageVO.getToolsToMount()) {
+                Assert.notNull(tool.getId(),
+                        "Tool id is not provided when specifying to which tools storage should be mounted");
+                Assert.notNull(toolManager.load(tool.getId()),
+                        messageHelper.getMessage(MessageConstants.ERROR_TOOL_NOT_FOUND, tool.getId()));
+                if (!tool.isAllVersions()) {
+                    Assert.isTrue(CollectionUtils.isNotEmpty(tool.getVersions()),
+                            "Tool version is not specified but isAllVersion is false");
```
tcibinan: Let's consider no versions specified as all versions allowed case.

api/src/main/resources/dao/datastorage-dao.xml
```diff
@@ -597,6 +597,39 @@
...
+                        tool_id,
+                        all_tool_versions,
+                        tool_version_id,
+                        tool_version,
+                        t.image AS TOOL_IMAGE,
+                        dr.path AS TOOL_REGISTRY
+                    FROM pipeline.datastorage_tool_version AS dtv
+                    INNER JOIN pipeline.tool t ON t.id = dtv.tool_id
+                    INNER JOIN pipeline.docker_registry dr ON t.registry_id = dr.id
+                    WHERE dtv.datastorage_id = ?
```
tcibinan: Tool symlinks should be supported in this request as well. Please check similar queries in `api/src/main/resources/dao/tool-dao.xml`.

api/src/main/resources/db/migration/v2021.08.05_15.00__issue_2110_storage_to_docker_mount_restriction.sql
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS pipeline.datastorage_tool_version (
+    datastorage_id INT NOT NULL REFERENCES pipeline.DATASTORAGE (DATASTORAGE_ID) ON DELETE CASCADE,
+    tool_id INT NOT NULL REFERENCES pipeline.TOOL (ID) ON DELETE CASCADE,
+    tool_version_id INT DEFAULT NULL,
+    tool_version TEXT DEFAULT NULL,
+    all_tool_versions BOOLEAN DEFAULT TRUE NOT NULL,
```
tcibinan: I would recommend getting rid of `all_tool_versions` and considering `tool_version_id=null` as all versions case.

api/src/main/resources/db/migration/v2021.08.05_15.00__issue_2110_storage_to_docker_mount_restriction.sql
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS pipeline.datastorage_tool_version (
+    datastorage_id INT NOT NULL REFERENCES pipeline.DATASTORAGE (DATASTORAGE_ID) ON DELETE CASCADE,
+    tool_id INT NOT NULL REFERENCES pipeline.TOOL (ID) ON DELETE CASCADE,
+    tool_version_id INT DEFAULT NULL,
+    tool_version TEXT DEFAULT NULL,
```
tcibinan: Why do we need to store tool version name here? I believe we can just join the tool versions table in select requests.

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -1025,6 +1045,32 @@ private void checkDataStorageObjectExists(final AbstractDataStorage dataStorage,
...
+                    Map<String, ToolVersion> stringToolVersionMap = toolVersionManager
+                            .loadToolVersions(tool.getId(), tool.getVersions().stream()
+                                    .map(ToolVersionFingerprint::getVersion)
+                                    .filter(StringUtils::isNotBlank)
+                                    .collect(Collectors.toList()));
+                    for (ToolVersionFingerprint version : tool.getVersions()) {
+                        Assert.isTrue(StringUtils.isNotBlank(version.getVersion()),
+                                "Version could not be empty when configure tools to mount");
```
tcibinan: Let's also consider this case as all versions allowed.

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -136,7 +167,9 @@ def run(self, mount_root, tmp_dir):
                 return
             initialized_mounters = []
             for storage_and_mount in available_storages_with_mounts:
-                if not PermissionHelper.is_storage_readable(storage_and_mount.storage):
+                if not PermissionHelper.is_storage_readable(storage_and_mount.storage)
+                    or not PermissionHelper.is_storage_available_for_mount(storage_and_mount.storage, run):
+                    Logger.info('Storage is not readable or it is not allowed to be mount to this image', task_name=self.task_name)
                     continue
```
tcibinan: It would be nice to have two separate ifs here:

```python
if a:
    log('a')
    continue
if b:
    log('b')
    continue

```

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -58,6 +59,27 @@ def __init__(self):
     def is_storage_readable(cls, storage):
         return cls.is_permission_set(storage, READ_MASK)
 
+    # Checks that tool with its version for the current run is allowed to mount this storage
+    # or there is no configured restriction for this storage
+    @classmethod
+    def is_storage_available_for_mount(cls, storage, run):
+        if storage.tools_to_mount is None or len(storage.tools_to_mount) == 0:
```
tcibinan: You can use the following expression here:

```python
if storage.tools_to_mount:
    ...
```

api/src/main/java/com/epam/pipeline/dao/datastorage/DataStorageDao.java
```diff
@@ -117,17 +168,33 @@ public void deleteDataStorage(Long id) {
                 DataStorageParameters.getRowMapper());
     }
 
+    public List<AbstractDataStorage> loadAllDataStoragesWithToolsToMount() {
+        final List<AbstractDataStorage> storages = getNamedParameterJdbcTemplate().query(loadAllDataStoragesQuery,
+                DataStorageParameters.getRowMapper());
+        storages.forEach(storage -> storage.setToolsToMount(loadToolsToMountForStorage(storage.getId())));
+        return storages;
+    }
```
tcibinan: This method performs 1+n requests where n is a number of data storages. It seems quite an easy improvement to just load all datastorage-toolversion mappings by a single request here.

workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -58,6 +59,27 @@ def __init__(self):
...
+
+        tool = run["actualDockerImage"]
+        re_result = re.search(r"([^/]+)/([^:]+):?(:?.*)", tool)
+        registry, image, version = re_result.groups()
+        for tool_to_mount in storage.tools_to_mount:
+            if registry == tool_to_mount["registry"] and image == tool_to_mount["image"]:
+                if not v["version"]:
+                    return True
+                else if tool_to_mount["versions"]:
+                    return version and version in [v["version"] for v in tool_to_mount["versions"]]
```
tcibinan: Looks like this piece of code is not correct. Let's use something like the following:

```python
tool_versions_to_mount = tool_to_mount.get('versions', [])
return not tool_versions_to_mount or version in [v["version"] for v in tool_versions_to_mount]
```

api/src/main/resources/dao/datastorage-dao.xml
tcibinan: On second thought tool symlinks should be managed as an independent tools which basically means that the original sql request was fine.

tcibinan (approved):

PR #2121, opened 2021-08-11
Set original owner run parameter for "run as" runs

tcibinan:
Relates to #1949.

The pull request brings support for automatically set original owner pipeline run parameter in case pipeline was configured with `run as` option enabled. Such pipeline runs will have an additional run parameter which is set to a username of a person who launched the pipeline.

The name of this run parameter can be configured using the following system preference `launch.original.owner.parameter` which defaults to `ORIGINAL_OWNER`.


tcibinan:
Cherry-picked to `release/0.16` via e67ffe4b1852d8f997de9fee4d9c8752ef7c60c8.

PR #2158, opened 2021-09-13
Issue 2155 NFS observer implementation

Wedds:
This PR is related to issue #2155 

It brings some 'observer' service implementation, which performs tracking on all mounted NFS (or specific paths passed given in `CP_CAP_NFS_OBSERVER_TARGET_PATHS` env var) - filesystem events, which might change disk consumptions are being saved and dumped to the certain object storage.

PR contains the script itself, along with changes to `launch.sh` and init scripts, performing installation and configuration of that observer in the runs, __api__ and __dav__` instances.

Issue #2155, opened 2021-09-09
Improve NFS search synchronization 


Wedds:
**Background**
At the moment, NFS shares scanning is a time-consumable procedure, so synchronization of NFS storage is less frequent, compared to object storage.
As a result:
1. It is hard to control quotas on such type of storage
2. Its search relevance is low, which also affects the accuracy of billing reports.

It would be nice to implement a mechanism to reduce synchronization latency for NFS storage.

**Approach**
1. Let's implement a service, observing changes, affecting disk consumption of mounted shares.
To track all of the changes in a fileshares, that service should be active in all the runs and following core services:`cp-api-srv`, `cp-dav`. 
It seems, that [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html) API suits building such component.
2. Service gathers target FS events and persists them to configurable object storage.
3. Along with the regular full-synchronization of NFS, `elasticsearch-agent` performs more frequent resynchronization of files, affected by events collected.

PR #2160, opened 2021-09-15
Issue 2155 NFS events synchronization

Wedds:
This PR is related to issue #2155

It brings a new service into `elasticsearch-agent`, which performs more frequent (compared to common NFS) synchronization based on events, gathered by "observer" introduced in #2158.

It pulls files from a configurable bucket in chunks, extracts, and merges events. After that, it checks the corresponding NFS state and creates/updates/removes required documents in ES.


mzueva (approved):

Wedds:
Cherry-picked to `release/0.16` via 28ded9a6318becdc060293869ad28068383a41b2

PR #2171, opened 2021-09-28
NFS quotas: monitor implementation

Wedds:
This PR is related to NFS quotas control.

It brings the implementation of the 'quotas monitor' into API.  
This new service performs polls and checks if any quotas are applied to NFS storage. If so - it calculates the actual size, based on search documents stored in ES, and changes storage status according to action configured for the matching quota.

Currently `GB` and `PERCENTS` quotas are supported for `NFS` and `Lustre` shares (be conscious, that usage of `PERCENTS` quota makes no sense with NFS shares since it has really no total size limit).

PR contains:
- changes into DB migrations, SQL queries, DAOs and manager to support `mount_status` attribute of `NFSDataStorage`
- LustreFS improvements: new `capacityGb` field, the method in the related manager to find Lustre corresponding to a CP fileshare
- `NFSQuotasMonitor` service


api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSQuotasMonitor.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    private boolean hasRestrictingActions(final NFSQuotaNotificationEntry notification) {
+        final Set<String> actions = notification.getActions();
+        return CollectionUtils.size(actions) > 1
+               || (CollectionUtils.size(actions) == 1 && !actions.contains(emailNotificationAction));
+    }
+
+    private boolean excessLimit(final NFSDataStorage storage, final NFSQuotaNotificationEntry notification) {
```
mzueva: Better name is `exceedsLimit`, same for `exceedsAbsoluteLimit ` and similar

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSQuotasMonitor.java
```diff
@@ -0,0 +1,209 @@
...
+    private final String disableMountAction;
+    private final String notificationsKey;
+
+    public NFSQuotasMonitor(final DataStorageManager dataStorageManager,
+                            final SearchManager searchManager,
+                            final MetadataManager metadataManager,
+                            final FileShareMountManager fileShareMountManager,
+                            final LustreFSManager lustreManager,
+                            final MessageHelper messageHelper,
+                            final @Value("${data.storage.nfs.quota.action.email:EMAIL}")
```
mzueva: Can't we use enum for actions?

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSQuotasMonitor.java
Wedds: Done

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSQuotasMonitor.java
Wedds: Done, new `StorageQuotaAction` enum is introduced

mzueva (approved):

Issue #2181, opened 2021-10-04
SGE Autoscale cluster can't be scaled if master node cores is set to 0


SilinPavel:
For now if we will set `CP_CAP_SGE_MASTER_CORES` = `0` `autoscale_sge` will fail to scale because can't parse sge output

PR #2187, opened 2021-10-04
issue 2181 fix qstat parsing while scaling up

SilinPavel:
This PR fixes bug when `autoscale_sge` could crash while scaling up, when muster is not configured as a worker


tcibinan (approved):
Looks fine.

Nevertheless `qsub` command will exit with non zero code if a user tries to submit a job in a cluster with a single main node and `CP_CAP_SGE_MASTER_CORES=0`. The issue is resolved in #1622.

tcibinan:
Partially resolves issue #2181.

PR #2201, opened 2021-10-08
Support manager node users and groups sharing within a cluster

### LINKED ISSUE #2195: Support users and groups sharing within clusters
**Background**

Relates to #2197.

Currently all nodes in both static and autoscaled clusters manage users and groups independently. For certain use cases (f.e. per user SGE jobs submission) it can be a good idea to share cluster manager node users and groups with all its worker nodes.

**Approach**

The simplest yet tricky approach would be to share manager node local users and groups management files (_/etc/passwd, /etc/shadow, /etc/group, /etc/gshadow_ and etc.) between all nodes in cluster.

**Other options**

Commonly tools like LDAP or NIS are used for users and groups management between multiple nodes. Nevertheless it seems like an overhead for Cloud Pipeline clusters atm.

tcibinan:
Resolves #2195.

The pull request brings support for cluster-wide manager node users and groups sharing.

The following list of run parameters can be used to configure cluster users and groups sharing:
- `CP_CAP_SHARE_USERS` enables cluster users and groups sharing. Defaults to `false`.
- `CP_CAP_SHARE_USERS_LOGGING_LEVEL` configures logging level in a GUI log. Defaults to `ERROR`. 
- `CP_CAP_SHARE_USERS_LOGGING_LEVEL_LOCAL` configures logging level in a local log file. Defaults to `DEBUG`.

Local log file can be found here `$LOG_DIR/configure_shared_users.log` or `/runs/pipeline-12345/logs/configure_shared_users.log`.

### DIFF
--- a/workflows/pipe-common/scripts/configure_shared_users.py
+++ b/workflows/pipe-common/scripts/configure_shared_users.py
@@ -89,7 +89,7 @@ def configure_shared_users():
         logger.success('Shared users and groups management was successfully configured.')
     except BaseException as e:
         traceback.print_exc()
-        stacktrace = traceback.format_stack()
+        stacktrace = traceback.format_exc()
         logger.error('Shared users and groups management configuration has failed: {} {}'.format(e, stacktrace))
         raise
 
--- a/workflows/pipe-common/scripts/configure_shared_users.py
+++ b/workflows/pipe-common/scripts/configure_shared_users.py
@@ -1,0 +1,74 @@
+# Copyright 2017-2021 EPAM Systems, Inc. (https://www.epam.com/)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+import traceback
+
+from pipeline.api import PipelineAPI
+from pipeline.common.common import execute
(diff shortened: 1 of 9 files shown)


workflows/pipe-common/scripts/configure_shared_users.py
```diff
@@ -0,0 +1,74 @@
...
+    try:
+        if cluster_role != _MANAGER_CLUSTER_ROLE:
+            logger.info('Configuring shared users and groups management...')
+
+            logger.info('Installing sshfs...')
+            install_package(rpm='fuse-sshfs', nonrpm='sshfs', logger=logger)
+
+            logger.info('Mounting etc directory from manager node...')
+            mkdir(_ETC_DIRECTORY_SHARED)
+            execute('sshfs -o ro,allow_other root@lizardfs-master:{etc_directory} {etc_directory_shared}'
```
mzueva: We need to check that lizard fs is used as shared fs or support NFS & Lustre masters as well

workflows/pipe-common/scripts/configure_shared_users.py
```diff
@@ -0,0 +1,74 @@
...
+from pipeline.utils.path import mkdir
+
+_MANAGER_CLUSTER_ROLE = 'master'
+_ETC_DIRECTORY = '/etc'
+_ETC_DIRECTORY_SHARED = '/etc-shared'
+_ETC_FILE_PATHS = ['passwd', 'shadow', 'group', 'gshadow']
+
+if __name__ == '__main__':
+    api_url = os.environ['API']
+    api_token = os.environ['API_TOKEN']
```
mzueva: Unused variable?

mzueva (changes requested):

workflows/pipe-common/scripts/configure_shared_users.py
tcibinan: Thanks. From now on parent run pod ip will be used rather than some shared fs specific host name.

workflows/pipe-common/scripts/configure_shared_users.py
tcibinan: Yeap.

mzueva (approved):

PR #2203, opened 2021-10-08
Support local user accounts creation for shared runs

### LINKED ISSUE #2197: Support local users synchronisation for shared runs
**Background**

Relates to #2195.

Currently any user can share her/his run with some other user or a group. In this case run endpoints are accessible by users the run was shared with. Nevertheless such users don't have corresponding local accounts in the shared run and therefore they can only connect using local root account. As a result there is no way to strictly limit certain user permissions within a run.

**Approach**

Users which a run was shared with should be automatically created within a run. Some daemon process can constantly check run sharing configuration and create missing local accounts.

**Other options**

Differently local accounts can be automatically created when some user tries to connect to a run using SSH using either GUI or CLI.

tcibinan:
Resolves #2197 and depends on #2201.

The pull request brings support shared user local accounts synchronisation. 

The following list of run parameters can be used to configure cluster users and groups sharing:
- `CP_CAP_SYNC_USERS` enables users synchronisation. Defaults to `false`.
- `CP_CAP_SYNC_USERS_HOME_DIR` configures root directory for user account home directories. Defaults to `$SHARED_ROOT/home` or `/common/home`.
- `CP_CAP_SYNC_USERS_TIMEOUT_SECONDS` configures timeout between two continuous synchronisations in seconds. Defaults to `60` seconds.
- `CP_CAP_SYNC_USERS_LOGGING_LEVEL` configures logging level in a GUI log. Defaults to `ERROR`. 
- `CP_CAP_SYNC_USERS_LOGGING_LEVEL_LOCAL` configures logging level in a local log file. Defaults to `DEBUG`.

Local log file can be found here `$LOG_DIR/sync_users.log` or `/runs/pipeline-12345/logs/sync_users.log`.

### DIFF
--- a/deploy/contents/install/preferences/launch.system.parameters.json
+++ b/deploy/contents/install/preferences/launch.system.parameters.json
@@ -362,5 +362,12 @@
         "description": "Enables cluster users sharing.",
         "defaultValue": "true",
         "passToWorkers": true
+    },
+    {
+        "name": "CP_CAP_SYNC_USERS",
+        "type": "boolean",
+        "description": "Enables users synchronization.",
+        "defaultValue": "true",
+        "passToWorkers": false
     }
 ]
--- a/scripts/pipeline-launch/launch.sh
+++ b/scripts/pipeline-launch/launch.sh
@@ -1401,6 +1401,25 @@ echo
 
 
 
+######################################################
+# Setup users synchronization if required
+######################################################
+
+echo "Setup users synchronization"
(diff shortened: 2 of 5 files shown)


mzueva (approved):

tcibinan:
Merged to a wrong branch but cherry-picked to `develop` via d5af0057ea0a7dd56c37638e04c280402d498414.

PR #2251, opened 2021-10-27
Issue 2233 Request DAV mount for storage

SilinPavel:
This PR provides new possibility to request and call off DAV mount for storage.

For now when dav mount is requested by user metadata tab `dav-mount` will be set with value of timestamp when this mount should be invalidated.

`syncmount` script from `dav service` now checks if storage timestamp is expired and if so - call of dav mount for this storage and umount it


mzueva (changes requested):
Could you please check checkstyle?

mzueva (approved):

PR #2256, opened 2021-10-28
Issue 2244 Release Notes Agent: Cloud-Pipeline Integration

kamyshova:
The PR provides CP integration implementation for Release Notes Agent:

- get the current version
- compare two version
- store the current version if needed


release-notes-agent/src/main/java/com/epam/release/notes/agent/entity/version/Version.java
```diff
@@ -26,4 +31,24 @@
     String buildNumber;
     String sha;
 
+    public static Version buildVersion(final String version) {
+        final String versionPattern = "^0.\\d+.0.\\d+.\\w+$";
```
SilinPavel: should be a constant

release-notes-agent/src/main/java/com/epam/release/notes/agent/entity/version/Version.java
```diff
@@ -26,4 +31,24 @@
...
+        final Pattern pattern = Pattern.compile(versionPattern);
```
SilinPavel: also can be a constant value

release-notes-agent/src/main/java/com/epam/release/notes/agent/entity/version/Version.java
```diff
@@ -26,4 +31,24 @@
...
+        Matcher matcher = pattern.matcher(version);
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException(format("The application version %s doesn't match the pattern %s ", version,
+                    versionPattern));
+        }
+        final String[] versionParts = version.split("\\.");
```
SilinPavel: we can reuse matcher here if will update regexp with regexpt groups to be able to do the next:
```
Matcher matcher = pattern.matcher(version);
Version.builder()
                .major(matcher.group(1)))
                .buildNumber(matcher.group(2))
                .sha(matcher.group(3))
                .build();
```

release-notes-agent/src/main/java/com/epam/release/notes/agent/entity/version/Version.java
SilinPavel: lets use the next version pattern:
`^(\\d+\\.\\d+\\.\\d+)\\.(\\d+)\\.(\\w+)$`

cloud-pipeline-common/model/src/main/java/com/epam/pipeline/entity/app/ApplicationInfo.java
```diff
@@ -0,0 +1,28 @@
...
+
+import lombok.Data;
+
+@Data
+public class ApplicationInfo {
+
+    private final String version;
+
+    public ApplicationInfo() {
+        this.version = getClass().getPackage().getImplementationVersion();
```
SilinPavel: I believe we don't need to fill it here, we need to fill it with response from cloud pipeline API
lets use `@Builder` and `@Value` here to be able to fill it with jackson and also leave it final

SilinPavel: I would prefer to add this functionality with separate PR actually

release-notes-agent/src/main/java/com/epam/release/notes/agent/utils/VersionUtils.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static com.epam.release.notes.agent.entity.version.Version.buildVersion;
+import static java.lang.String.format;
+
+public final class VersionUtils {
+
+    private VersionUtils() {
+    }
+
+    public static Version readVersionFromFile(final String versionFilePath) {
```
SilinPavel: I don't see any reason to store this methods it Util class, I would prefer to see it in ApplicationVersionServiceImpl actually

SilinPavel (changes requested):

release-notes-agent/src/main/java/com/epam/release/notes/agent/entity/version/Version.java
```diff
@@ -18,12 +18,37 @@
...
 
+    private static final String VERSION_PATTERN = "^(\\d+\\.\\d+\\.\\d+)\\.(\\d+)\\.(\\w+)$";
+    private static final Pattern PATTERN = Pattern.compile(VERSION_PATTERN);
+
     String major;
     String buildNumber;
     String sha;
 
+    public static Version buildVersion(final String version) {
+        Matcher matcher = PATTERN.matcher(version);
```
SilinPavel: final

Issue #2242, opened 2021-10-26
Release Notes Agent


SilinPavel:
**Approach**
Main idea:

Release Note Agent works in the manner of daemon application.

Each n sec (configurable) application checks if followed deployment `app version` is changed. If so - agent will do next:

If `version head number` (e.g. 0.16) isn't change:
            Get from an `app version` commit sha
            Goes to GitHub fetch all commits between old version and new version 
            Agent will parse commit messages of this commits and get issue numbers (it is required that such commit messages will have the next structure `Issue #<issue id> ....`), configure and store issue links and info.
            Agent goes to Jira and search for all issue with tag `release/version`(could be configured)  that equal to new `app version`, collects required links and info from those issues

Else if version number is changed:
           If `version head number` is changed there could be two possible ways (configurable):
                   - Send an email for `admin` person that `head` version is changed and he/she needs to prepare release notes manually
                   - Prepare release notes only for last commit

After all issues will be fetched, email template should be populated and sent to pre-configured list of subscribers

This idea could be represented by the next diagram:

![Release Notes Agent drawio](https://user-images.githubusercontent.com/11179595/138879194-4991c480-bca4-42a0-bb36-e4279bc1955f.png)

**Other options**

PR #2263, opened 2021-10-28
Issue #2242 Release Notes Notifier

kamyshova:
The PR provides ReleaseNotesNotifier for Release Notes Agent.

### DIFF
--- a/release-notes-agent/src/main/java/com/epam/release/notes/agent/cli/ReleaseNotesNotifier.java
+++ b/release-notes-agent/src/main/java/com/epam/release/notes/agent/cli/ReleaseNotesNotifier.java
@@ -56,7 +56,10 @@ public class ReleaseNotesNotifier {
         }
 
         final VersionStatus versionStatus = applicationVersionService.getVersionStatus(old, current);
-        if (versionStatus == VersionStatus.NOT_CHANGED) {
+        if (versionStatus == VersionStatus.NOT_FOUND) {
+            applicationVersionService.storeVersion(current);
+            return;
+        } else if (versionStatus == VersionStatus.NOT_CHANGED) {
             log.info(format(
                     "The current API version %s has not changed.", current.toString()));
             return;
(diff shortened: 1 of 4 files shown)


release-notes-agent/src/main/java/com/epam/release/notes/agent/service/version/ApplicationVersionServiceImpl.java
```diff
@@ -56,13 +57,16 @@ public Version fetchCurrentVersion() {
     @Override
     public Version loadPreviousVersion() {
         try {
-            final String savedVersion = Files.lines(Paths.get(versionFilePath))
+            final Path path = Paths.get(versionFilePath);
```
SilinPavel, line 60: it is bad idea to change something in load method. Instead that lets move this code (about creation of a file) to `storeVersion` method, I believe it would be more naturally

SilinPavel: and here we can return null (or some predefined constant object) if there is no file or it is empty

release-notes-agent/src/main/java/com/epam/release/notes/agent/cli/ReleaseNotesNotifier.java
```diff
@@ -0,0 +1,79 @@
...
+    public void sendReleaseNotes(@ShellOption(defaultValue = ShellOption.NULL) List<String> emails) {
+
+        final Version old = applicationVersionService.loadPreviousVersion();
+        final Version current = applicationVersionService.fetchCurrentVersion();
+
+        if (emails == null) {
+            emails = subscribers;
+        }
+
+        final VersionStatus versionStatus = applicationVersionService.getVersionStatus(old, current);
```
SilinPavel, line 58: after we refactor a code in loadversion and now we return there null or predefined constant - we need to change logic here to do the next:
```
if version is not changed or prev version doesn't exist
   skip all
else
   do all stuff 

storing new version
```

SilinPavel: better to call `checkVersionStatus`

SilinPavel (changes requested):

SilinPavel (approved):

PR #2267, opened 2021-10-29
Issue 2242 release notes agent GitHub integration tests

Aleksei-Golovenchik:
This pull request implements test for `GitHubServiceImpl` class


release-notes-agent/src/main/java/com/epam/release/notes/agent/service/RestApiClient.java
```diff
@@ -35,4 +42,19 @@
             throw new IllegalStateException(e);
         }
     }
+
+    default <T> T createApi(final Class<T> clazz, final String gitHubBaseUrl, final Interceptor interceptor,
```
SilinPavel: please add docs describing all params especially interseptor

SilinPavel: `gitHubBaseUrl` -> `baseUrl` since it will be used not only by Github Api

SilinPavel (changes requested):

Issue #2259, opened 2021-10-28
`pipe` shall respect the ui.hidden.objects preference


sidoruka:
GUI now uses `ui.hidden.objects` to hide any object, which is listed in the preference (#1906).
`pipe` CLI shall respect this preference and skip the same objects from any list commands. E.g.:
* `pipe storage ls`
* `pipe view-...` commands
* Any other, which lists the Cloud Pipeline objects

Please backport to `release/0.16`

Issue #1906, opened 2021-04-05
Allow to hide certain objects from the general user's UI


sidoruka:
**Approach**
* Introduce `ui.hidden.objects` preference 
  * Section: `User Interfact`
  * Type: `JSON`
* Preference shall contain a list of the objects IDs, which are not displayed to the non-admin users in the GUI (whatever permissions are set)
* Format:
```
{
  "pipeline": [ 1, 2, 3, 4 ],
  "folder": [ 5, 6 ],
  "data_storage": [ 7, 8, 9 ],
  "tool": [ 10, 11 ],
  "tool_group": [ 12, 13 ],
  "configuration": [ 14 ]
}
```

mzueva:
@sidoruka done for `develop` & `release/0.16`

rodichenko:
@sidoruka @mzueva @NShaforostov @maryvictol @kamyshova GUI part implemented (#1912).

UI can handle the following properties of `ui.hidden.objects` preference:
```json
{
  "pipeline": [ 1, 2, 3, 4 ],
  "folder": [ 5, 6 ],
  "data_storage": [ 7, 8, 9 ],
  "tool": [ 10, 11 ],
  "tool_group": [ 12, 13 ],
  "configuration": [ 14 ],

```
and
```json
  "version": ["<pipeline-id>/<version-name>"],
  "tool_registry": [15, 16],
  "metadata_folder": [17, 18, "<metadata-folder-ID>"],
  "metadata_class": ["<parent-folder-id>/ClassName", "ClassName"]
}
```

rodichenko:
@sidoruka backported to `release/0.16` (2681fcf442f8684884997a02dc8c43b7e9230dfc)

PR #2279, opened 2021-11-02
issue 2259 pipe cli add filtering for hidden objects

SilinPavel:
Resolves issue #2259.

This PR adds filtering based on preference `ui.hidden.objects` for command such `storage ls` `view-pipes` etc


pipe-cli/src/utilities/cp_preference_utilities.py
```diff
@@ -0,0 +1,36 @@
...
+import json
+
+from src.api.preferenceapi import PreferenceAPI
+
+UI_HIDDEN_OBJECTS_PREFERENCE = 'ui.hidden.objects'
+hidden_objects = None
+
+
+def is_object_hidden(category, identifier):
+    global hidden_objects
```
tcibinan: Let's avoid the usage of global variables since it is not a good practice. What do you think if we rather use some stateful object instead?

```python
hidden_objects_manager = HiddenObjectsManager()
for pipeline in pipelines:
    if not hidden_objects_manager.is_hidden_pipeline(pipeline):
        yield pipeline
```

pipe-cli/pipe.py
```diff
@@ -350,8 +350,9 @@ def echo_title(title, line=True):
 @click.option('-p', '--parameters', help='List parameters of a pipeline', is_flag=True)
 @click.option('-s', '--storage-rules', help='List storage rules of a pipeline', is_flag=True)
 @click.option('-r', '--permissions', help='List user permissions for a pipeline', is_flag=True)
+@click.option('-h', '--show-hidden', help='List all pipelines, even if it is hidden', is_flag=True)
```
tcibinan: Let's show hidden objects to admins by default. Please use the similar approach as in [this piece of code](https://github.com/epam/cloud-pipeline/blob/develop/pipe-cli/src/utilities/pipeline_run_operations.py#L57-L64) from `pipe run` command implementation where we perform command differently if the user is admin.

We can put this admins-only logic to `HiddenObjectsManager`.

tcibinan: In GUI users don't have an option to see hidden objects. What do you think if we stick with the GUI's approach and remove this additional argument? We can still provide support for this feature to advanced users via f.e. `CP_SHOW_HIDDEN` environment variable.

pipe-cli/pipe.py
```diff
@@ -360,14 +361,19 @@ def view_pipes(pipeline, versions, parameters, storage_rules, permissions):
...
     pipes_table = prettytable.PrettyTable()
     pipes_table.field_names = ["ID", "Name", "Latest version", "Created", "Source repo"]
     pipes_table.align = "r"
-    pipelines = Pipeline.list()
+    pipelines = []
+    if show_hidden:
+        pipelines = Pipeline.list()
+    else:
+        pipelines = [p for p in Pipeline.list() if not cp_preference_utilities.is_object_hidden('pipeline', p.identifier)]
+
```
tcibinan, line 381: Let's also hide pipeline versions if specified as described here https://github.com/epam/cloud-pipeline/issues/1906#issuecomment-814278737.

pipe-cli/src/utilities/cp_preference_utilities.py
SilinPavel: since we need to know size of a collection we can't use generators here, but I added specific method to manager for the future

SilinPavel:
@tcibinan I have refactored the code, please check

pipe-cli/src/utilities/cp_preference_utilities.py
tcibinan: It was just an example ;). I believe you can use the same thing as before in a call site.

```python
[p for p in Pipeline.list() if not hidden_objects_manager.is_hidden_pipeline(p.identifier)]
```

pipe-cli/src/utilities/hidden_object_manager.py
```diff
@@ -0,0 +1,45 @@
...
+    def is_object_hidden(self, category, identifier):
+        if not self.user_operation_manager.is_admin() and not self.show_hidden \
+                and self.hidden_objects and category in self.hidden_objects:
+            return identifier in self.hidden_objects[category]
+        return False
+
+    def filter_hidden(self, category, elements, id_mapper):
+        for element in elements:
+            if not self.is_object_hidden(category, id_mapper(element)):
+                yield element
```
tcibinan: Seems like this method is not used anywhere.

Issue #2283, opened 2021-11-03
Release Notes Agent: Implement approach for support several actions


SilinPavel:
It would be convenient to have some kind of `ActionServiceProvider` that will store all available `ActionService`s and be able to run `perform` method for it.

The main idea of this provider is:

 - We have some `command` for spring-shell
 -  This command has number of actions that should be performed 
 -  We pass this actions to the `ActionServiceProvider` and its run `perform` method of specific ActionServices

PR #2288, opened 2021-11-08
issue #2283: Implement approach for support several actions for Release Notes Agent

kamyshova:
The PR provides an implementation of the approach to support several actions for Release Notes Agent through ActionServiceProvider.


release-notes-agent/src/main/java/com/epam/release/notes/agent/entity/action/Action.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Getter
+@RequiredArgsConstructor
+public enum Action {
+    POST("post"),
+    PUBLICATION("publish");
```
SilinPavel: we do not need now any additional action, lets leave only POST

release-notes-agent/src/main/java/com/epam/release/notes/agent/service/action/ActionServiceProvider.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.epam.release.notes.agent.entity.action.Action;
+import com.epam.release.notes.agent.service.action.mail.TemplateNotificationService;
+import org.springframework.mail.javamail.JavaMailSender;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ActionServiceProvider {
+
+    private final JavaMailSender javaMailSender;
```
SilinPavel: those fields shouldn't be there, `ActionServiceProvider` doesn't need to know any information about internal staff of `ActionNotificationService`s. Actually this class should be a simple storage for all kind of `ActionNotificationService`s and be able to return specific object of specific `ActionNotificationService` by Action

release-notes-agent/src/main/java/com/epam/release/notes/agent/service/action/PublishActionNotificationService.java
```diff
@@ -0,0 +1,31 @@
...
+package com.epam.release.notes.agent.service.action;
+
+import com.epam.release.notes.agent.entity.github.GitHubIssue;
+import com.epam.release.notes.agent.entity.jira.JiraIssue;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+@Service
+public class PublishActionNotificationService implements ActionNotificationService {
```
SilinPavel: remove this class since it is not used right now

SilinPavel (changes requested):

Issue #2249, opened 2021-10-26
Release Notes Agent: Pre-commit hooks


SilinPavel:
Since we decide to implement Release Note Agent #2242  and this software relies on commit messages conventions we need to provide convenient way to remind for the developer team to commit all changes with appropriate messages.

Lets use https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks for this

The main idea is to define Pre-Commit or Pre-Push hook to be able to restrict such action if messages of this commits is not complaint with our conventions.

For now we can see the next helpful conventions:
- Each commit should have constriction that target specific issue ( [ Issue #<issuenumber>, issue #<issuenumber> ] )
- If You need to commit anything without specifying an issue you need to use pre-defined construction `Issue #undefined#`

**NOTE** It is also would be helpful to automatically change PR merge commits messages

PR #2298, opened 2021-11-11
Issue #2249: Pre-commit hooks

kamyshova:
The PR provides git hooks for checking the commit message and preparing a commit message template in respecting format conventions.
The commit message should include an issue number in the `Issue # / issue #` format or `#no issue#` format, otherwise, the script rejects the commit. The `Issue #undefined#` format was introduced for exceptional cases and should be avoided.
To apply hooks in the project:
`git config core.hooksPath $PROJECT_DIR/.hooks`

### DIFF
--- a/.hooks/commit-msg
+++ b/.hooks/commit-msg
@@ -1,0 +1,19 @@
+#!/bin/bash
+#
+# This hook script checks the message commit format.
+# If the message does not include an issue number in the 'Issue # / issue #' format, the script rejects the commit.
+# It is also possible to use the 'Issue #undefined#' format, but this should be avoided.
+#
+
+MESSAGE="$1"
+
+matched=$(grep -E "^.*(((I|i)ssue( |_)#)[1-9]+).*" "$MESSAGE")
+if [ -z "$matched" ] ; then
+    matched_undefined=$(grep -E "#(N|n)o( |_)(I|i)ssue#" "$MESSAGE")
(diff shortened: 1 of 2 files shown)


.hooks/commit-msg
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# This hook script checks the message commit format.
+# If the message does not include an issue number in the 'Issue # / issue #' format, the script rejects the commit.
+# It is also possible to use the 'Issue #undefined#' format, but this should be avoided.
+#
+
+MESSAGE="$1"
+
+cat "$MESSAGE"
```
SilinPavel: why we need to cat it here?

.hooks/commit-msg
```diff
@@ -0,0 +1,21 @@
...
+matched=$(grep -E "^.*(((I|i)ssue( |_)#)[1-9]+).*" "$MESSAGE")
+cat "$matched"
+if [ -z "$matched" ] ; then
+    matched_undefined=$(grep -E "(I|i)ssue( |_)#undefined#" "$MESSAGE")
```
SilinPavel: Lets use shorten message if no issue is provided, to make it more comfortable
what about `#(N|n)o( |_)(I|i)ssue#` to get something like `#no issue#`

SilinPavel (changes requested):

PR #2310, opened 2021-11-16
Refactor release notes notifier

GrKolychev:
This PR brings some refactor of code as add validation of emails list and default emails subjects


GrKolychev:
-

Issue #2308, opened 2021-11-15
Release Notes Notification Agent: Deployment


SilinPavel:
We need to have mechanism to deploy Release Notes Notification Agent #2242 

Lets consider the next approach:

- It should be deployed as a kube CronJob
- Schedule should be configured when overall deploy process is started
- This service should be optional to deploy
- application properties and logs dir should be mounted to the pod

PR #2311, opened 2021-11-16
Issue #2308: Release Notes Agent Deployment

kamyshova:
The PR provides deployment for Release Notes Notification Agent. The Release Notes Agent will be deployed as a Kubernetes CronJob.


deploy/contents/install/app/install.sh
```diff
@@ -1271,5 +1276,19 @@ if is_service_requested cp-run-policy-manager; then
   fi
 fi
 
+# Release notes agent - send email with api changes to subscribers
+if is_service_requested cp-release-notes; then
+  print_ok "[Starting release-notes agent deployment]"
+  print_info "-> Deleting existing instance of release-notes agent"
+  delete_deployment_and_service "cp-release-notes" \
```
SilinPavel: better to call in `cp-release-notes-agent`

deploy/contents/install/app/install.sh
```diff
@@ -1271,5 +1276,19 @@ if is_service_requested cp-run-policy-manager; then
...
+                                  "/opt/release-notes"
```
SilinPavel: `/opt/release-notes-agent`

deploy/contents/k8s/cp-release-notes/cp-release-notes-dpl.yaml
```diff
@@ -0,0 +1,52 @@
...
+            - name: release-notes-logs
+              hostPath:
+                path: /opt/release-notes/logs
+            - name: release-notes-template
+              hostPath:
+                path: /opt/release-notes/templates
+            - name: release-notes-data
+              hostPath:
+                path: /opt/release-notes/data
+          restartPolicy: OnFailure
```
SilinPavel: I believe that actually we don't need to restart again and again the job if it is fails, because it isn't a critical service and with `restartPolicy: OnFailure` we have a risk to stack in a infinity loop if something goes wrong

SilinPavel: lets explicitly disable it

deploy/docker/cp-release-notes/config/application.properties
```diff
@@ -0,0 +1,40 @@
...
+github.issue.regex=(?i)\\(?issue #.+
+github.issue.number.regex=.+#(\\d+).*
+github.token=${CP_RELEASE_NOTES_GITHUB_TOKEN}
+
+jira.base.url=${CP_RELEASE_NOTES_JIRA_URL}
+jira.auth.token=Basic ${CP_RELEASE_NOTES_JIRA_BASE_64_TOKEN}
+jira.version.custom.field.id=${CP_RELEASE_NOTES_JIRA_VERSION_FIELD_ID:}
+jira.github.custom.field.id=${CP_RELEASE_NOTES_JIRA_GITHUB_FIELD_ID:}
+
+release.notes.agent.admin.emails=${CP_RELEASE_NOTES_ADMIN_EMAIL:}
```
SilinPavel: now I see that this two case (admin/subscribers) actually very similar cases and it differ from each other only in matter of version change type (minor/major)

so actually I thing it would be much better to refactor this names to:
```
release.notes.agent.templates.dir=

# Major version changed case's settings
release.notes.agent.major.version.changed.subscriber.emails=
release.notes.agent.major.version.changed.email.template=
release.notes.agent.major.version.changed.subject=

# Minor version changed case's settings
release.notes.agent.minor.version.changed.subscriber.emails=
release.notes.agent.minor.version.changed.email.template=
release.notes.agent.minor.version.changed.subject=
```

SilinPavel: the same changes i believe better to do with env vars

SilinPavel (changes requested):

PR #2320, opened 2021-11-18
Add support for run billings exporting

### LINKED ISSUE #1716: Billing reports: full-size & raw data export
### Background

At the moment in Cloud Pipeline, billing reports allow CSV export of only limited quantity of data (only `TOP-N` rows, displaying at GUI). In some cases users may want to export the whole billing reports data - e.g. to perform the further analysis. So, it would be useful to allow users export the full tables of billing reports.

Another moment, in billing CSV export only the aggregated data attends, which is shown in the GUI.
It would be nice to have an option to export the raw (non-aggregated) information for the compute and data consumption.

### Approach

1. Existing CSV export feature shall produce a CSV with all the entries, not only `TOP-N` rows.
   This should be applied for the export:
   - from the "**Storages**" section (lists of _Storages_, _File storages_, _Object storages_)
   - from the "**Compute instances**" section (lists of _Instances_, _Pipelines_, _Tools_)
   Only exporting files should be changed, no changes are required for the UI.
2. GUI shall expose a new export option - `Export raw information`.
   This export shall produce a CSV with the per-run information, respecting any billing filters configured by the user.
   This option should be available only for the export from the "**Compute instances**" section.
   The exported data should include the following columns:
   - RunID
   - User name (run owner)
   - Pipeline name (if applicable)
   - Docker image
   - Instance type
   - Run start date and time
   - Run finish date and time
   - Run duration
   - Cost
   Data shall be sorted by the cost (descending)

tcibinan:
Resolves second task of #1716.

The pull request brings support for run billings export. The export produces a csv file where each row represents a single run with aggregated cost and duration within the specified time period.

*Notice* that *pipeline name*, *started* and *finished* column values will be empty for previously indexed runs. The limitation won't affect any runs which will be active once the billing service is redeployed.

Examples and more details can be found in #2358.


api/src/main/java/com/epam/pipeline/entity/billing/RunBilling.java
```diff
@@ -0,0 +1,20 @@
+package com.epam.pipeline.entity.billing;
+
+import lombok.Builder;
+import lombok.Value;
+
+import java.time.LocalDateTime;
+
+@Value
+@Builder
+public class RunBilling {
```
mzueva, line 10: I think it would be useful to add `computeType` and `billingCenter` fields to the report fields

billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/service/impl/mapper/RunBillingMapper.java
```diff
@@ -84,4 +91,8 @@ private long scaled(final BigDecimal price) {
                 .map(BigInteger::longValue)
                 .orElse(0L);
     }
+
+    private String asString(final Date date) {
+        return date != null ? SIMPLE_DATE_FORMAT.format(date) : null;
```
mzueva: Maybe use `Optional` here?

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/entity/billing/RunBilling.java
tcibinan: Added.

billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/service/impl/mapper/RunBillingMapper.java
tcibinan: Done.

Issue #2338, opened 2021-11-23
It shall be possible to disable `COMMIT` function for the tool


sidoruka:
**Background**
Some of the tools contain applications with a restrictive licensing. Cloud Pipeline shall offer functionality to keep the user away from copying the docker contents from the image.

One of such features - restrict `commit` operation for a tool. So the user won't be able to save a copy into a personal tool group. 

**Approach**
* Tool/version settings shall contain `Allow commit of the tool` bool parameter. It shall be set to `true` by default.
* If it is set to `true` - commit operation shall work as now
* If it is set to `false`:
  * GUI shall not display `COMMIT` hyperlink for any job, which uses that tool
  * API shall throw error if a `COMMIT` operation is requested for such tool

sidoruka:
@rodichenko please take care of the GUI part in the Tools settings forms

sidoruka:
Please backport to `release/0.16`

rodichenko:
@sidoruka backported to `release/0.16` (6dbfc7e46a158dbe9324912dd1df9184425290d6)

maryvictol:
**Test case** was created by #2370 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/testcases/UI/Run_tools_in_sandbox/2338.md).

maryvictol:
e2e-test was implemented via #2370 .

maryvictol:
Daily e2e-tests results are located and updated [here](https://github.com/epam/cloud-pipeline/blob/e2e-reports/e2e/reports/AWS/test/classes/com.epam.pipeline.autotests.DockerCommitTest.html) (for `AWS`).

PR #2341, opened 2021-11-23
Issue #2338 Tool commit restrictions

Wedds:
This PR is related to issue #2338

It brings the implementation for tool commit restrictions processing at the server-side:
- add `allowCommit` field to the `Tool` and `ToolVersion` allow persisting it in the DB
- validate `allowCommit` status during commit-related operations in the corresponding managers: if a commit is not allowed on the tool level or for a given version - the operation will be aborted

Configuration details:
| Scope| Endpoint|Description|Correponding UI page|
| ----- | :---: | :---: | :---: |
| Restrictions could be configured for all of the versions at once by changing the flag of the tool (it will not change the status of versions, but since commit validation resolving starts from the tool all of its versions are forbidden for commit). | `POST` **/tool/update** | `allowCommit` field in the request body<br>(_true_ by default) | https://cp-deployment.com/pipeline/#/tool/1/settings |
| Commit operation shouldn't be allowed for a specific version only | `POST`<br> **/tool/{toolId}/settings** | `allowCommit` request parameter<br>(optional, _true_ by default) | https://cp-deployment.com/pipeline/#/tool/1/info/latest/settings |


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -762,6 +762,8 @@ public PipelineRun updatePrettyUrl(Long runId, String url) {
     @Transactional(propagation = Propagation.REQUIRED)
     public PipelineRun updateCommitRunStatus(Long id, CommitStatus commitStatus) {
         PipelineRun pipelineRun = pipelineRunDao.loadPipelineRun(id);
+        toolManager.validateCommitOperationAllowed(null, pipelineRun.getActualDockerImage());
```
tcibinan: Let's add an overloaded method `toolManager.validateCommitOperationAllowed(final String image)`.

Wedds: Done

tcibinan (approved):

mzueva:
@sidoruka backported to `release/0.16`

Issue #2342, opened 2021-11-24
Allow to restrict "system" parameters only to specific users groups


sidoruka:
* API shall support `roles` field (array) for the parameters, defined in `launch.system.parameters`. E.g.:
```
{
        "name": "CP_FSBROWSER_ENABLED",
        "type": "string",
        "description": "Disables BROWSE button for the job",
        "defaultValue": "false",
        "passToWorkers": true,
        "roles": [
                "ROLE_ADMIN",
                "ROLE_ADVANCED_USER"
        ]
}
```
* Only members of the roles, defined by `roles` field, shall be able to use those parameters
* If `roles` is empty or not set - consider it as "allow to all the users"
* This new field shall be respected by the GUI and CLI
* GUI part:
  * Do not show parameters in the "Add system parameters" popup
  * If the parameter is added manually via "Add parameter" menu - show an alert, that this parameter is not allowed for use
  * This shall be enforced for all the forms: Launch, Tool settings, Pipeline settings, Detached configuration
* CLI part:
  * If the user launches a job with a parameter, which is not allowed - a meaningful error shall be thrown and job shall NOT be started
  * E.g. using an example from the (1):
    * A member of `ROLE_USER` tries to run `pipe run -di library/centos:7 --CP_FSBROWSER_ENABLED true`
    * Job is not started
    * An error message is shown
```
An error has occurred while starting a job: "CP_FSBROWSER_ENABLED" parameter is not permitted for use
```

sidoruka:
Please backport to `release/0.16`

rodichenko:
@sidoruka GUI part implemented (d4315df7e2c947df263e3fa72a344784fee2df5a ), backported to `release/0.16` (bbe7cd2348f20f0244dcd2f4161f70e2cb92412d)

Wedds:
CLI and API parts are merged into `develop` (6c679cbec693e9391e3ee10ba75dbe83da6c95f7) and backported to `release/0.16`(a46f787507706e12ce712a0ddf91fc719e6056cc)

maryvictol:
Test cases were created by #2379 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/Run_tools_in_sandbox/Launch_parameters/2342).

maryvictol:
e2e-tests were implemented via #2422.

maryvictol:
[...]

PR #2348, opened 2021-11-26
GUI: Allow to restrict "system" parameters only to specific users groups

TimSPb89:
GUI for #2342


rodichenko:
Merged with d4315df7e2c947df263e3fa72a344784fee2df5a

Issue #2354, opened 2021-11-29
Cloud Pipeline shall allow to disable mounting of specific data storages


sidoruka:
**Background**
At the moment, Cloud Pipeline allows to restrict data storages mounts based on the docker images selection (introduced by #2110).
But some of the data storages shall never be mounted. E.g. system storages ("etc", "fsbrowser", "fsobserver").

**Approach**
* Introduce an option for the data storage to disable the mounting at all:
  * From the GUI - this shall be a checkbox `Disable mount`. If it is set - `Mount point`, `Mount options`, `Allow mounts to` controls shall be hidden
  * From the API - just another option in the data storage object
* When any job is launched:
  * `mount_storage.py` shall check for this option and always skip such storages
  * GUI shall hide such storages from the `Limit mounts` dialogs (all forms)

maryvictol:
Test cases were created by #2447 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/Run_tools_in_sandbox/Limit_mounts/2354).

maryvictol:
e2e-tests were implemented via #2449.

maryvictol:
Daily e2e-tests results are located and updated [here](https://github.com/epam/cloud-pipeline/blob/e2e-reports/e2e/reports/AWS/test/classes/com.epam.pipeline.autotests.DataStoragesFeaturesTest.html) (for `AWS`).

PR #2359, opened 2021-11-30
Issue 2354 disable mounting of some storages

SilinPavel:
This PR related to #2354 
It adds additional field for `AbstractDatastorage` (default false)
if this field is true - `mount_storage.py` will skip such storage while mounting process


api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -1031,6 +1031,10 @@ private AbstractDataStorage updateDataStorageObject(AbstractDataStorage dataStor
                 storageProviderManager.getStorageProvider(dataStorage).getDefaultMountOptions(dataStorage));
         }
 
+        if (dataStorageVO.getMountDisabled() != null) {
```
mzueva, line 1034: Maybe use the same logic as in `AbstractDataStorageFactory` with `BooleanUtils` to prevent null values

mzueva (changes requested):

api/src/main/resources/db/migration/v2021.11.29_15.00__issue_2354_disable_mount_for_storage.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pipeline.datastorage ADD mount_disabled BOOLEAN DEFAULT FALSE NOT NULL;
```
mzueva: Please move migration to current date

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
SilinPavel: I believe here we need explicitly check for null, because there could be a case, when storage is updated and this value is not provided, it means that we don't need to change current value. But if we will use here BooleanUtils.isTrue, we will not be able to update value from True to False for example.

mzueva (approved):

PR #2361, opened 2021-11-30
GUI xmas tree: management

AlfiyaRF:
This PR includes previews for appearance settings


client/src/components/settings/appearance-management/element-preview.js
```diff
@@ -16,13 +16,24 @@
...
+    super(props);
+    this.state = {
+      navigationSelected: {
+        home: false,
+        run: false,
+        search: false
+      }
+    };
+  }
+
```
rodichenko: Though this is OK in terms of [`constructor` usage in React component](https://reactjs.org/docs/react-component.html#constructor), we do not have any specific initialization logic here (like binding methods). Let's change this to general state initialization:
```javascript
class ElementPreview extends React.Component {
  state = {
    navigationSelected: {
      home: false,
      run: false,
      search: false
    }
  };
}
```
as we do so in other components of the application.

rodichenko (changes requested):

rodichenko (approved):

PR #2368, opened 2021-12-01
GUI Xmas tree management: image uploader

ZMaratovna:


client/src/components/settings/appearance-management/element-preview.js
```diff
@@ -131,6 +130,8 @@ class ElementPreview extends React.Component {
   renderPreviews () {
     const {section} = this.props;
     switch (section) {
+      case sectionNames.backgroundImage:
+        return this.renderBackgroundImage();
```
rodichenko: This component does not have `renderBackgroundImage` method

rodichenko (changes requested):

client/src/components/settings/appearance-management/element-preview.js
ZMaratovna: fixed

Issue #2232, opened 2021-10-20
NAT Gateway configuration


NShaforostov:
## Background

Currently, if the Cloud Pipeline Platform is deployed in some private subnet, it could be quite difficult for the admin to expose a network endpoint for some service. This requires manual execution of a number of tasks on the Platform Core instance and, accordingly, may lead to errors.
It would be convenient to allow the management (creating/removing) network routes from the GUI - this would save from wasting time and allow to avoid possible errors in configurations.

### GUI

Add a new form to the **System Settings** that will allow admins to create/remove network routes from private subnet(s) to the external network resources.
By default, this form shall display:
- the list of existing routes (port forwarding map). Each route record should contain:
  - info about external resources:
    - server name
    - IP
    - port
  - info about corresponding internal config (mapping details):
    - service name
    - service IP
    - port
  - button to remove the route
- button to add a new route
- disabled button to save changes

Approximate view of that form:
![image](https://user-images.githubusercontent.com/45459424/138140678-cc9dae56-b305-40d8-bc41-f279cadf9ca0.png)

When admin clicks the **Add route** button - new field shall appear:
- text field to specify a server name of the adding service
- control to specify an IP of the adding service:
  - text field to specify IP manually
  - button to autoresolve IP by specified server name. This button should be disabled until the server name is not specified.
  Admin can specify IP manually or use this button - for the last case, we should implement the mechanism to resolve an IP and output it to the IP field
- control to specify a port of the adding service:
  - text field to specify a port
  - button to add another port. Admin shall have the ability to add any count of ports for one server name.
  By click this button - additional text field should appear under the first port field and a control to remove that additional port
- control to remove these fields

Approximate view of that form:
![image](https://user-images.githubusercontent.com/45459424/138140335-74a3ddac-0780-4af0-8218-da96550f4ec5.png)

Admin should specify all fields. The **Save** button should become enabled, e.g.:
![image](https://user-images.githubusercontent.com/45459424/138140138-9044026c-344a-4278-a55e-c22ce98f2acb.png)
Admin should click it to save changes. After the saving, the routing list shall be refreshed. If the form will be closed without click that button - all changes should be canceled (the corresponding warning message should be shown).

In case, when the route record was added - the new `Kubernetes` service, that handles requests from the private subnet and passes it to the "Port forwarding service" shall be created, the corresponding records shall be added to mapping configs. See details [below](#route-creation).

[...]

PR #2383, opened 2021-12-07
Issue #2232 Force usage of the system-default DNS 

Wedds:
This PR is related to issue #2232 

System default DNS should be used when checking a new cluster DNS record because `defaultCustomDnsIP` could be configured to some external resource, which is not aware of cluster-level records.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/NatGatewayManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/NatGatewayManager.java
@@ -174,8 +174,8 @@ public class NatGatewayManager {
         return updateRoutesInDatabase(request, NatRouteStatus.TERMINATION_SCHEDULED);
     }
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/NatGatewayManager.java
```diff
@@ -174,8 +174,8 @@ public void updateStatus() {
         return updateRoutesInDatabase(request, NatRouteStatus.TERMINATION_SCHEDULED);
     }
 
-    private Set<String> tryResolveAddress(final String hostname) {
-        return tryResolveAddress(hostname, null);
+    private Set<String> tryResolveAddressUsingSystemDns(final String hostname) {
```
mzueva: Isn't it more simple just or call `resolveNameUsingSystemDefaultDns` ?

mzueva (approved):

Issue #762, opened 2019-10-21
Cloud Pipeline shall allow users to configure spending quotas


NewOlya:
## Background

The Cloud-Pipeline Billing (spending reports) is implemented and users can view their spendings for different periods.
Now, it would be convenient for administrators to have the ability to configure spending quotas - to manage money limits of the platform usage and define different restrictions for users when these limits are achieved that may allow to reduce global spending of the Platform usage.

## Approach

1. We should add to the Billing page two tabs - _Spendings_ and _Quotas_. To the _Spending_ tab, all currently displayed forms with the Platform billing reports should be placed. In the _Quotas_, forms with quotas configuring described below should be placed.
2. Only admins and users with the **ROLE\_BILLING\_MANAGER** shall have the ability to configure quotas and have the access to the _Quotas_ tab
3. There shall be the ability to configure quotas on the following entities:
   - **whole Platform** - summary spendings on compute instances and data storages
   - separately on **compute instances** - spendings on jobs running
   - separately on **storages** - spendings on data storing
4. Additionally for each entity (_except whole Platform_), there shall be the ability to specify the following types of quotas:
   - _User_ - quota that should be applied to a specific user
   - _User group/role_ - quota that should be applied to a specific group/role
   - _Billing Center_ - quota that should be applied to a specific billing center. This is a different dimension of the users grouping. Typically this is a meta-group, which is not used to apply security permissions
   - _Overall_ - the overall budget for the current entity
   Example of _Quotas_ tab view (when no quotas is configured yet):
   ![image](https://user-images.githubusercontent.com/45459424/139245311-8ab1d687-0d7e-4c75-ba1f-95e53d849332.png)
5. There shall be the ability to configure time period for the quotas setting.
  Possible values should be: _month_/_quarter_/_year_. The default time period should be one calendar month.
  There shall not be the ability to set different periods for different quotas. Only one period type can be set at once - it should be applied for all existing quotas
6. For each specific configured quota, there shall be the ability to specify the thresholds in percent (e.g. `50%` / `75%` / `100%`), upon reaching which specific actions shall be performed by the platform.
  Thresholds shall be specified manually, value can be more than `100%`.
  For each quota, any count of thresholds can be set.
7. For each quota and threshold, there shall be the ability to configure the action that will be performed when the specified limit (threshold) will be reached.
   - For **Compute instances** entity, there should be:
[...]

PR #2401, opened 2021-12-14
Cloud Pipeline shall allow users to configure spending quotas

ekazachkova:
The current PR provides implementation for issue #762 

- a new API method `POST /quotas` that creates a new quota entity
```
{
    "quotaGroup": "STORAGE",
    "type": "USER",
    "subject": "user1",
    "value": 1000,
    "actions": [
      {
        "threshold": 146,
        "actions": [
          "NOTIFY"
        ]
      }
}
// OR whole Platform:
{
    "quotaGroup": "GLOBAL",
    "period": "QUARTER",
    "value": 1000
}
```
- a new API method `GET /quotas/{id}` that loads quota by ID
- a new API method `PUT /quotas/{id}` that updates quota by ID
Note: it is not possible to update quota's group and type
- a new API method `DELETE /quotas/{id}` that deletes quota by ID
- a new API method `GET /quotas` that loads all quotas as list


api/src/main/java/com/epam/pipeline/entity/quota/QuotaEntity.java
```diff
@@ -0,0 +1,78 @@
...
+    private Long value;
+
+    @ManyToMany
+    @JoinTable(
+            name = "quotas_users",
+            schema = "pipeline",
+            inverseJoinColumns = { @JoinColumn(name = "user_id") },
+            joinColumns = { @JoinColumn(name = "quota_id") }
+    )
+    private List<PipelineUser> informedUsers;
```
mzueva: It should be an object similar to `NFSQuotaNotificationRecipient` or any SID entity (`name` and `isPrincipal` flag), as we need to support group/role notifications as well

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/entity/quota/QuotaActionEntity.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @ManyToOne
+    @JoinColumn(name = "quota_id", referencedColumnName = "id")
+    private QuotaEntity quota;
+
+    private Integer threshold;
```
mzueva: Let's make it double

api/src/main/java/com/epam/pipeline/entity/quota/QuotaEntity.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Enumerated(EnumType.STRING)
+    private QuotaType type;
+
+    @Enumerated(EnumType.STRING)
+    private QuotaPeriod period;
+
+    private String name;
+
+    private Long value;
```
mzueva: Let's make it double

api/src/main/java/com/epam/pipeline/entity/quota/QuotaEntity.java
```diff
@@ -0,0 +1,78 @@
...
+    @Enumerated(EnumType.STRING)
+    private QuotaGroup quotaGroup;
+
+    @Enumerated(EnumType.STRING)
+    private QuotaType type;
+
+    @Enumerated(EnumType.STRING)
+    private QuotaPeriod period;
+
+    private String name;
```
mzueva: `subject` or `subjectName`

Issue #2399, opened 2021-12-13
Allow to skip certain files/folders when applying quotas to the filesystems


sidoruka:
**Background**
Certain apps (like RStudio) use home directories, which are EFS volumes, to store session data. When the session is suspended - all the current environment is dumped into the home directory (it is common to have 50Gb+ session snapshot).
If there are quotas set for the home directory - it gets reached quickly and the users are not aware of the reasons.
We need to skip some of the "well-known" paths when indexing the EFS quotas to mitigate this issue.

**Approach**
* Introduce a preference to list the paths, which shall be skipped during filesystem volume computation
* When the size is computed we need to store two values:
  * "Real" size - same as a current behavior, all the files volume
  * "Effective" size - exclude the the files/folders, which fall into the preference
* Quotas shall be now applied based on the "Effective" size. "Real" size shall be used for billing reports.
* API shall expose both sizes in the data storage details, "Real" size is provided for the information.
* GUI shall display both sizes. "Effective" shall be shown by default. "Real" size shall be displayed in the brackets or in the tooltip.
* Preference shall support setting the list of "skipped" folders/files for a single storage or a list of storages. Each storage entry shall support basic "wildcards" (as we have e.g. for the allowed instance types). So it is possible to defined an "absolute" storage name (e.g. `my_storage1`) or a set of storages in a single shot (e.g. `my_storage*` or even all storages with `*`).

sidoruka:
Shall be backported to `release/0.16`

rodichenko:
@sidoruka GUI part implemented (4d64e609197e6c5eeea2a773cd42043395075972)

rodichenko:
@sidoruka backported to `release/0.16` (595515d0f2a4750acd061988423f915690bda149)

sidoruka:
@mzueva @Wedds please implement "wildcards" for the `storageName` field. See last point from the original issue.

mzueva:
> @mzueva @Wedds please implement "wildcards" for the `storageName` field. See last point from the original issue.

Done and backported to `release/0.16`

NShaforostov:
Docs were added via #3228 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/08_Manage_Data_Storage/8.7._Create_shared_file_system.md#exclude-certain-directories-from-monitoring).

PR #2402, opened 2021-12-15
Issue #2399 Skip certain files during storage size calculation

Wedds:
This PR is related to issue #2399

It allows specifying masking rules for storage volume computation - files matching those masks are not affecting 'effective' storage size, which is used for quotas control from now on.
- volume calculation masks have the same format, as search hiding ones (`StorageFileSearchMask`), and are specified via system preferences as well
- `StorageUsage` received a new `effectiveSize` attribute
- masks are used in a separate request for 'effective' size retrieval (`MultiSearchRequest` is performed to receive both 'total' and 'effective' values at one request to ES)

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
+++ b/api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
@@ -823,6 +823,13 @@ public class SystemPreferences {
         SEARCH_GROUP,
         isNullOrValidJson(new TypeReference<List<StorageFileSearchMask>>() {}));
 
+    public static final ObjectPreference<List<StorageFileSearchMask>> STORAGE_SIZE_MASK_RULES = new ObjectPreference<>(
+        "storage.size.calculation.mask",
+        Collections.emptyList(),
+        new TypeReference<List<StorageFileSearchMask>>() {},
+        SEARCH_GROUP,
+        isNullOrValidJson(new TypeReference<List<StorageFileSearchMask>>() {}));
+
     // Grid engine autoscaling
     public static final IntPreference GE_AUTOSCALING_SCALE_UP_TIMEOUT =
             new IntPreference("ge.autoscaling.scale.up.timeout", 30,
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/epam/pipeline/manager/search/SearchResultConverter.java
```diff
@@ -72,19 +73,27 @@ public SearchResult buildResult(final SearchResponse searchResult,
...
     }
 
-    public StorageUsage buildStorageUsageResponse(final SearchResponse searchResponse,
+    public StorageUsage buildStorageUsageResponse(final MultiSearchResponse searchResponse,
                                                   final AbstractDataStorage dataStorage, final String path) {
-        final Long size = Optional.ofNullable(searchResponse.getAggregations())
-                .map(aggregations -> aggregations.<ParsedSum>get(STORAGE_SIZE_AGG_NAME))
-                .map(result -> new Double(result.getValue()).longValue())
-                .orElse(0L);
+        final MultiSearchResponse.Item[] responses = searchResponse.getResponses();
```
mzueva: Maybe we need to check whether we have expected number of successful requests and add some NPE & size checks?

api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
```diff
@@ -823,6 +823,13 @@
         SEARCH_GROUP,
         isNullOrValidJson(new TypeReference<List<StorageFileSearchMask>>() {}));
 
+    public static final ObjectPreference<List<StorageFileSearchMask>> STORAGE_SIZE_MASK_RULES = new ObjectPreference<>(
```
mzueva: Please name this preference `STORAGE_QUOTAS_SKIPPED_PATHS` `storage.quotas.skipped.paths` and let's move it to storage group

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
Wedds: Done

api/src/main/java/com/epam/pipeline/manager/search/SearchResultConverter.java
Wedds: Done, checking MultiSearchResponse structure and each of the expected nested responses

mzueva (approved):

PR #2413, opened 2021-12-17
Issue #2399 gui quotas size tooltip

AlfiyaRF:
This PR implements tooltip for quotas size

### DIFF
--- a/client/src/components/special/storage-size/index.js
+++ b/client/src/components/special/storage-size/index.js
@@ -19,7 +19,6 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import {message, Tooltip, Icon} from 'antd';
 import {inject, observer} from 'mobx-react';
-import Markdown from '../markdown';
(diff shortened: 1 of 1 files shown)


client/src/components/special/storage-size/index.js
```diff
@@ -27,11 +28,35 @@ import styles from './storage-size.css';
...
 
+function InfoTooltip ({size}) {
+  const {realSize, effectiveSize} = size;
+
+  if (!effectiveSize || effectiveSize === realSize) {
+    return null;
+  }
+
+  const tooltip = `Effective size: ${displaySize(effectiveSize, effectiveSize > 1024)}  
+    Real size: ${displaySize(realSize, realSize > 1024)}`;
```
rodichenko: We don't need to create *markdown* here; we can build a simple component:
```js
const tooltip = (
  <div>
    <div>Effective size: ...</div>
    <div>Real size: ...</div>
  </div
);
```

rodichenko (changes requested):

rodichenko (approved):

Issue #2423, opened 2021-12-21
Maintenance mode support


sidoruka:
Cloud Pipeline shall be capable of setting a so called "Maintenance mode" to limit access to certain features, while platform is being updated of other maintenance operations are happening.

* Introduce a preference `system.maintenance.mode`
  * Type: `boolean`
  * Default value: `false`
* If this preference is set to `true`, the following behaviour changes shall happen:
  * `PAUSE`, `RESUME`, `COMMIT` API calls shall return an error `Platform is in a maintenance mode, operation is temporary unavailable`
  * Any running `PAUSE`, `RESUME`, `COMMIT` processes shall proceed (do not stop them if the preference is set to `true`)
  * Hot node pools autoscaling shall be disabled. I.e. pools shall be just ignored while maintenance mode is on. Any running scale job shall shall proceed. "General" autoscaling (for the new runs) shall work as usual.
  * GUI shall show the `PAUSE`, `RESUME`, `COMMIT` as "disabled". Hovering these buttons shall show a tooltip with a message `Platform is in a maintenance mode, operation is temporary unavailable`
* Introduce a preference `system.maintenance.mode.banner`
  * Type: `string`
  * Default value: `Platform is in a maintenance mode. Some of the features are disabled.`
* If `system.maintenance.mode.banner` is set to a non-empty value and a `system.maintenance.mode` is set to true, then  a "system notification" in the top-right corner with a "info" icon. Content shall be taken from `system.maintenance.mode.banner` preference.

sidoruka:
Please backport to `release/0.16`

rodichenko:
@sidoruka 
GUI part implemented: 
* Disable `PAUSE`, `RESUME`, `COMMIT` buttons: 4462dbff5a72fcc7c98d5a8cb7318c7e6842a1b2
* Show banner: 3554f1a99a94f3257be0a398c887914c2f0af0a3

maryvictol:
@rodichenko ,
**Bug:** Message `Platform is in a maintenance mode, operation is temporary unavailable` shouldn't be shown on *Stop run confirmation pop-up* in Maintenance mode.

![image](https://user-images.githubusercontent.com/47791878/153405017-35ce2ef4-e63c-4178-95b0-bd295021229c.png)

rodichenko:
[...]

PR #2430, opened 2021-12-22
Issue #2423: Maintenance mode support

ekazachkova:
The current PR provides implementation for issue #2423


mzueva (approved):

api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
```diff
@@ -750,6 +750,12 @@
     public static final BooleanPreference SYSTEM_DISABLE_NAT_SYNC = new BooleanPreference(
             "system.disable.nat.sync", true, SYSTEM_GROUP, pass);
 
+    public static final BooleanPreference SYSTEM_MAINTENANCE_MODE = new BooleanPreference(
+            "system.maintenance.mode", true, SYSTEM_GROUP, pass);
```
mzueva: Default value shall be `false`

ekazachkova:
```suggestion
            "system.maintenance.mode", false, SYSTEM_GROUP, pass);
```

Issue #2433, opened 2021-12-23
[Platform usage] Show user statuses


NShaforostov:
**Background**

Currently, admins can monitor Platform usage, for example, by list of **ACTIVE RUNS** or via **CLUSTER STATE** pages.
But for some cases, it can be useful for admins to know which users do utilize the Platform in the current moment.
So, we should add the displaying of user states in the "User management" system tab.

**Approach**

In the "**User management**" system tab, for "**Users**" sub-tab add the displaying of users' states.
Possible states should be:
- _Online_ - for users who are logged in and use the Platform in the moment
- _Offline_ - for users who are not logged in at the moment/do not use the Platform for some time

Details:
- the state should be shown as simple icon near the user name, e.g.:
  ![image](https://user-images.githubusercontent.com/45459424/147242554-1ec33e31-bc42-49d4-a7fd-e6e07e9a90f0.png)
- by hover over the _Offline_ icons - the tooltip shall appear that contains the info when the specific user has utilized the Platform the last time, e.g.:
  ![image](https://user-images.githubusercontent.com/45459424/147243964-c5884c44-11d5-4e23-a245-83971fee866f.png)

Additionally:
- Add two new items to the users dropdown filter:
  - ability to show only users who are _online_ at the moment
  - ability to show only users who are _offline_ at the moment
  ![image](https://user-images.githubusercontent.com/45459424/147244604-a6ff61b0-f2a2-4c98-a868-ec692f4c4c5b.png)

All described functionality shall be accessible and shown only for admins.

sidoruka:
Shall be backported to `release/0.16`

maryvictol:
@rodichenko 
**Bug**: User with `ROLE_USER_READER` role shouldn't see statuses icons for users in the *Users Management* list.

maryvictol:
@rodichenko
**Bug:** *Show online users* and *Show offline users* filters should be hidden (or disabled) for User with `ROLE_USER_READER` role.

maryvictol:
@ekazachkova,
User session in impersonation mode should be displayed in **Last visited:** info as well as normal login.

maryvictol:
> @ekazachkova, User session in impersonation mode should be displayed in **Last visited:** info as well as normal login.

Verified as fixed

maryvictol:
> @rodichenko **Bug:** _Show online users_ and _Show offline users_ filters should be hidden (or disabled) for User with `ROLE_USER_READER` role.

Verified as fixed

[...]

PR #2437, opened 2021-12-24
Issue #2433: [Platform usage] Show user statuses

ekazachkova:
This PR provides implementation for issue #2433

- a new boolean request parameter `activity` added to `GET /users` API method (Default: false). If `activity=true` and current user is admin a new field `online` will be present at the pipeline user object.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/acl/user/UserApiService.java
+++ b/api/src/main/java/com/epam/pipeline/acl/user/UserApiService.java
@@ -132,6 +132,11 @@ public class UserApiService {
     }
 
     @PreAuthorize(ADMIN_OR_GENERAL_USER + OR_USER_READER)
+    public List<PipelineUser> loadUsersWithActivityStatus() {
+        return new ArrayList<>(userManager.loadUsersWithActivityStatus());
+    }
+
+    @PreAuthorize(ADMIN_OR_GENERAL_USER + OR_USER_READER)
     public List<UserInfo> loadUsersInfo(final List<String> userNames) {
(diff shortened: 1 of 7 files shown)


api/src/main/java/com/epam/pipeline/controller/user/UserController.java
```diff
@@ -223,8 +223,10 @@ public Result deleteUser(@PathVariable Long id) {
     @ApiResponses(
             value = {@ApiResponse(code = HTTP_STATUS_OK, message = API_STATUS_DESCRIPTION)
             })
-    public Result<List<PipelineUser>> loadUsers() {
-        return Result.success(userApiService.loadUsers());
+    public Result<List<PipelineUser>> loadUsers(@RequestParam(defaultValue = "false") final boolean activities) {
```
mzueva: Let's rename this parameter to `activity`

api/src/main/java/com/epam/pipeline/acl/user/UserApiService.java
```diff
@@ -131,6 +131,11 @@ public PipelineUser updateUserRoles(Long id, List<Long> roles) {
         return new ArrayList<>(userManager.loadAllUsers());
     }
 
+    @PreAuthorize(ADMIN_OR_GENERAL_USER + OR_USER_READER)
```
mzueva: This method shall have the access as `loadUsers` - `ADMIN_ONLY + OR_USER_READER` (according to the task)

mzueva (changes requested):

mzueva (approved):

Issue #2444, opened 2021-12-28
Support UDP traffic for the NAT Gateway


sidoruka:
**Background**
At the moment, Cloud Pipeline allows to configure NAT Gateway via #2232 for the TCP traffic/ports. We need to implement the same for the UDP ports, as some of the workloads require UDP for the bulk data transfers.

**Approach**
* GUI: 
  * Add `Protocol` dropdown list to the `Add Route` form:
![image](https://user-images.githubusercontent.com/20998291/147560562-19ec4fae-5185-4c82-9625-b47d4163ebd8.png)
  * It shall contain two items: `TCP` (default) and `UDP`
  * This dropdown list shall be shown for each port. I.e. if one clicks `Add port` - another `Protocol` selector shall be shown
  * Add `Protocol` column to the `EXTERNAL RESOURCES` section of the routes list
* Backend: 
  * For UDP protocol - use `CP_TP_UDP_DEST` parameter in the `cp-config-global` map
  * For backward compatibility, please consider any route, without an explicit protocol specification, as a TCP

maryvictol:
**Test cases** were created by #2546 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/System_Management/2232-2444_NAT_Gateway).

maryvictol:
**e2e-tests** were implemented via https://github.com/epam/cloud-pipeline/pull/2546.

maryvictol:
**Daily e2e-tests** results are located and updated [here](https://github.com/epam/cloud-pipeline/blob/e2e-reports/e2e/reports/AWS/test/classes/com.epam.pipeline.autotests.NATGatewayTest.html) (for `AWS`).

NShaforostov:
Docs added via https://github.com/epam/cloud-pipeline/pull/3387 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/12_Manage_Settings/12.14._NAT_gateway.md).

PR #2448, opened 2021-12-29
GUI NAT: support UDP (WIP)

ZMaratovna:
This PR implements feature for #2444


client/src/components/settings/system-management/nat-getaway-configuration/add-route-modal/add-route-modal.js
```diff
@@ -43,7 +43,7 @@ export default class AddRouteForm extends React.Component {
   };
 
   state = {
-    ports: {'port': undefined},
+    ports: {'port': {value: undefined, protocol: 'tcp'}},
```
rodichenko: Please define `const protocols = {TCP: 'tcp', UDP: 'udp'}` enumeration and use it.

rodichenko (changes requested):
We should use enumerations for predefined values like `TCP` and `UDP`

PR #2477, opened 2022-01-20
GUI user reports: rename controls, handle clicks

ZMaratovna:


client/src/components/settings/user-management/usage-report/charts/users-activity-chart.js
```diff
@@ -340,7 +343,8 @@ class UsersActivityChart extends React.Component {
       period,
       range,
       runner,
-      tooltip: undefined
+      tooltip: undefined,
+      mode: modes[period][0]
```
rodichenko: `modes[period]` may be undefined - we should check that.

client/src/components/settings/user-management/usage-report/charts/users-activity-chart.js
```diff
@@ -486,7 +495,7 @@ class UsersActivityChart extends React.Component {
         <div className={styles.modeSwitcher}>
           <Button.Group>
             {
-              AllModes.map(mode => (
+              modes[period].length > 1 && modes[period].map(mode => (
```
rodichenko: We should check if `modes[period]` is not `undefined`

rodichenko (changes requested):

rodichenko (approved):

PR #2535, opened 2022-02-25
Initial pipe-common sge cluster support 

kamyshova:
The PR provides initial SGE cluster support to pipe-common.

### DIFF
--- a/workflows/pipe-common/pipeline/cluster/cluster.py
+++ b/workflows/pipe-common/pipeline/cluster/cluster.py
@@ -1,0 +1,63 @@
+# Copyright 2022 EPAM Systems, Inc. (https://www.epam.com/)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import subprocess
+from random import getrandbits
+from threading import RLock
+from . import utils
+
+lock = RLock()
+
+
+class AbstractCluster(object):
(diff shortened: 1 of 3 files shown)


workflows/pipe-common/pipeline/cluster/utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+def merge_log(common_logfile, job_logfile):
+    piece_size = 4096
+    with open(common_logfile, 'ab+') as common_logfile, open(job_logfile, 'rb') as job_logfile:
+        while True:
+            piece = job_logfile.read(piece_size)
+            if piece == b'':
+                common_logfile.write("\n".encode())
+                common_logfile.seek(0)
+                job_logfile.seek(0)
```
tcibinan: I'm not sure if we need to do this. Is there any reason to seek to the file beginnings?

workflows/pipe-common/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import subprocess
+from random import getrandbits
+from threading import RLock
+from . import utils
+
+lock = RLock()
```
tcibinan: Lets rather create `_lock` field in `SGECluster` class and then pass it to `create_directory` method. The usage of global variables with common names can be pretty harmful.

workflows/pipe-common/pipeline/cluster/utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def get_log_filename(work_directory, job_name):
+    logs_directory = create_directory(work_directory, "logs")
+    random_number = getrandbits(64)
+    return os.path.join(logs_directory, "{job_name}_{number}_out.log"
+                        .format(number=random_number, job_name=job_name if job_name else random_number))
+
+
+def merge_log(common_logfile, job_logfile):
```
tcibinan: This method can be called from multiple threads. I imagine that the current implementation can lead to damaged log files where some symbols are not fully written. What do you think if we synchronize while merging log files?

tcibinan (approved):

Issue #2552, opened 2022-03-11
VM-Monitor service shall notify about the "core" filesystem free space


sidoruka:
**Background**
At the moment, Cloud Pipeline (via `vm-monitor` service) allows to monitor the core services health and certificates expiration dates.
But the "core" filesystem (`/opt`) is not covered by monitoring. Any issue with this filesystem may cause a global platform outage.

**Approach**
* VM-Monitor shall run a scheduled analysis of the "core" filesystem(s)
* List of the filesystems shall be configurable
* Monitoring shall be done via a filesystem stats (i.e. like `df` command does). We shall not compute everything on the fly (i.e. do not use `du` behavior)
* List of the thresholds shall be configurable as well
* Once the threshold is reached - fire an email

PR #2553, opened 2022-03-11
Issue #2552 Implement core FS monitoring

Wedds:
This PR is related to issue #2552

It extends `vm-monitor` service, allowing scanning of the core file system.

### DIFF
--- a/deploy/contents/k8s/cp-vm-monitor/cp-vm-monitor-dpl.yaml
+++ b/deploy/contents/k8s/cp-vm-monitor/cp-vm-monitor-dpl.yaml
@@ -56,6 +56,9 @@ spec:
             - mountPath: /opt/common/pki
               name: common-pki
               readOnly: true
+            - mountPath: /opt/cp-core-fs
+              name: cp-core-fs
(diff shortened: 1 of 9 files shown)


deploy/contents/k8s/cp-vm-monitor/cp-vm-monitor-dpl.yaml
```diff
@@ -56,6 +56,9 @@ spec:
             - mountPath: /opt/common/pki
               name: common-pki
               readOnly: true
+            - mountPath: /opt/cp-core-fs
```
mzueva: Let's change path not to overlap with other mounts, e.g. `/fs/core-fs`

mzueva (changes requested):

deploy/contents/k8s/cp-vm-monitor/cp-vm-monitor-dpl.yaml
Wedds: Moved default mount-point to `/fs`

PR #2561, opened 2022-03-15
Added cp-idp deployment

madmongoose:
Deployment solves the problem of deleting credentials after pods rebooted.

### DIFF
--- a/deploy/contents/k8s/cp-idp/cp-idp-dpl.yaml
+++ b/deploy/contents/k8s/cp-idp/cp-idp-dpl.yaml
@@ -1,0 +1,71 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: cp-idp
+  namespace: default
+spec:
+  replicas: 1
+  template:
+    metadata:
+      namespace: default
+      labels:
+        cloud-pipeline/cp-idp: "true"
+    spec:
+      nodeSelector:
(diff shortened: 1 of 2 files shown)


deploy/contents/k8s/cp-idp/cp-idp-dpl.yaml
```diff
@@ -0,0 +1,75 @@
...
+            path: /opt/gitlab/pki
+        - name: share-pki
+          hostPath:
+            path: /opt/share-srv/pki
+        - name: common-pki
+          hostPath:
+            path: /opt/common/pki
+        - name: idp-profiles-folder
+          hostPath:
+            path: /opt/idp/basic
```
SilinPavel: do we really need it here?  I believe we can just mount file from `/opt/idp/`

deploy/docker/cp-idp/init
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+# Check profiles file is not empty
+if [ -s saml-idp-profiles.json ]; then
+    echo "The profiles file is not-empty"
+else
+    echo "The profiles file is empty"
```
SilinPavel: please expand a comment with information what this script will do with this script in this case

SilinPavel (changes requested):

SilinPavel (approved):

Issue #2581, opened 2022-03-31
Show additional data storages info via pipe CLI


sidoruka:
**Background**
At the moment, Cloud Pipeline does not allow to view the data storage information easily. `pipe storage ls` shows only a list of them. 
It would be beneficial to show (optionally) the following information:
- Tags (key=value)
- Quota status: ACTIVE/READONLY/MOUNT_DISABLED
- Mount limits (if a data storage is limited to a specific docker image)

**Approach**
Let's add this functionality to the `pipe storage ls` as an addition option: `-o|--output full/compact`
* Not set or `compact` - current bahavior, only a list of storages
* `full` - new behavior with the extended information in the list

sidoruka:
Please backport to `release/0.16`

PR #2586, opened 2022-04-01
Issue #2581 CLI extended storage listing

Wedds:
This PR is related to the issue #2581 

It allows configuring 'extended' storage listing mode via` -o|--output full/compact` parameter. 
When `full` mode selected, storage metadata, status and mount limits are shown. Note, that `full` format works for storage summary listing only and takes effect with `-l|--show_details` flag only.

Ex:
- `extended` mode
```
pipe storage ls -l -o full

Type    Labels    Modified               Size    Name          Mount status      Mount limits                                   Metadata
S3                2022-01-13 20:00:11            bucket1       ACTIVE            cp-deployment.com:443/library/ubuntu                                                                                                                                                 
S3                2021-12-29 03:11:34            bucket2       DISABLED                                                         key_1=value_1, key_2=value_2                                                                                          
NFS               2022-03-09 16:44:16            nfs-storage   READ_ONLY         cp-deployment.com:443/library/centos:latest                                                                                                                                                          

```
- `compact` mode keeps the current behavior

```
pipe storage ls -l

Type    Labels    Modified               Size    Name          
S3                2022-01-13 20:00:11            bucket1       
S3                2021-12-29 03:11:34            bucket2                                                                                            
NFS               2022-03-09 16:44:16            nfs-storage    
```

### DIFF
--- a/pipe-cli/pipe.py
+++ b/pipe-cli/pipe.py
@@ -1064,11 +1064,25 @@ def mvtodir(name, directory):
 @click.option('-r', '--recursive', is_flag=True, help='Recursive listing')
 @click.option('-p', '--page', type=int, help='Maximum number of records to show')
 @click.option('-a', '--all', is_flag=True, help='Show all results at once ignoring page settings')
+@click.option('-o', '--output', default='compact', show_default=False,
+              help="Option for configuring storage summary details listing mode. Possible values:"
+                   "compact - brief summary only"                   
+                   "full - show extended details")
 @common_options
-def storage_list(path, show_details, show_versions, recursive, page, all):
+def storage_list(path, show_details, show_versions, recursive, page, all, output):
     """Lists storage contents
     """
-    DataStorageOperations.storage_list(path, show_details, show_versions, recursive, page, all)
+    show_extended = False
+    if output != 'compact':
+        if output != 'full':
+            click.echo('Invalid listing mode: `{}`, possible values are [full, compact]!'.format(output), err=True)
+            sys.exit(1)
+        elif path is not None and output == 'full' or not show_details:
+            click.echo('Extended output could be configured for the storage summary listing only!', err=True)
+            sys.exit(1)
+        else:
+            show_extended = True if output == 'full' else False
+    DataStorageOperations.storage_list(path, show_details, show_versions, recursive, page, all, show_extended)
 
 
 @storage.command(name='mkdir')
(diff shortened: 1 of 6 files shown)


pipe-cli/pipe.py
```diff
@@ -1064,11 +1064,25 @@ def mvtodir(name, directory):
 @click.option('-r', '--recursive', is_flag=True, help='Recursive listing')
 @click.option('-p', '--page', type=int, help='Maximum number of records to show')
 @click.option('-a', '--all', is_flag=True, help='Show all results at once ignoring page settings')
+@click.option('-o', '--output', default='compact', show_default=False,
```
mzueva: We can use `Choice` to limit number of allowed options instead of manual check

pipe-cli/src/api/metadata.py
```diff
@@ -39,22 +39,44 @@ def find(cls, identifier, entity_class):
...
         else:
             raise RuntimeError("Failed to load metadata.")
 
+    @classmethod
+    def load_metadata_mapping(cls, entity_ids, entity_class):
+        metadata_list = Metadata.load_all_for_ids(entity_ids, entity_class)
+        metadata_mapping = dict()
+        for metadata_entry in metadata_list:
+            metadata_data_dict = {}
+            for key, data in metadata_entry.data.iteritems():
```
mzueva: We need to use `from future.utils import iteritems` to make it compatible with Python 2 & 3

mzueva (changes requested):

pipe-cli/pipe.py
Wedds: Done, thanks for the tip!

pipe-cli/src/api/metadata.py
Wedds: Done

pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -481,6 +496,27 @@ def __print_data_storage_contents(cls, bucket_model, relative_path, show_details
...
+            item.metadata = metadata_mapping.get(item.identifier, {})
+
+    @classmethod
+    def load_metadata_mapping(cls, items):
+        storage_ids = [item.identifier for item in items]
+        metadata_list = Metadata.load_metadata_mapping(storage_ids, 'DATA_STORAGE')
+        metadata_mapping = dict()
+        for metadata_entry in metadata_list:
+            metadata_data_dict = {}
+            for key, data in metadata_entry.data.iteritems():
```
mzueva: Please, fix iteritems here as well

mzueva (changes requested):

PR #2593, opened 2022-04-08
Issue 2577: Change Cloud Data app executable icon

kamyshova:
The PR contains changes of the pre-built Windows distribution icon for the Cloud Data application.


deploy/docker/cp-api-preconfigure/preconfigure-api
```diff
@@ -0,0 +1,87 @@
...
+    (( ${#dir_contents[*]} )) && result=1 || result=0
+    return $result
+}
+
+function configure_icon_win_distribution() {
+    # Check and change windows icon if exists
+    CLOUD_DATA_WIN_ICO_NAME=${CLOUD_DATA_WIN_ICO_NAME:-"favicon.ico"}
+    if [ "$CP_PREF_API_STATIC_DIRECTORY_EXT" ] && ! is_dir_empty "$CP_PREF_API_STATIC_DIRECTORY_EXT"; then
+        if [ "$CP_PREF_API_STATIC_DIRECTORY_EXT/$CLOUD_DATA_WIN_ICO_NAME" ]; then
+            xvfb-run -f $JENKINS_HOME/.Xauthority -a wine $JENKINS_HOME/reshack/ResourceHacker.exe -open $CP_COMMON/cloud-data-win32-x64/cloud-data.exe -save $CP_COMMON/cloud-data-win32-x64/cloud-data.exe -action addoverwrite -res $CP_PREF_API_STATIC_DIRECTORY_EXT/favicon.ico -mask ICONGROUP,MAINICON,
```
Wedds: Seems, like
> ... -res $CP_PREF_API_STATIC_DIRECTORY_EXT/favicon.ico  ..

Should be changed to:
> ... -res $CP_PREF_API_STATIC_DIRECTORY_EXT/$CLOUD_DATA_WIN_ICO_NAME ...

in case the icon name is overridden.

It could be extracted to some variable

> CLOUD_DATA_WIN_ICO_PATH="$CP_PREF_API_STATIC_DIRECTORY_EXT/$CLOUD_DATA_WIN_ICO_NAME"

to be re-used in `if` condition and both `xvfb-run` calls

deploy/docker/cp-api-preconfigure/preconfigure-api
```diff
@@ -0,0 +1,87 @@
...
+
+    shopt -u nullglob
+    shopt -u dotglob
+
+    local result=0
+    (( ${#dir_contents[*]} )) && result=1 || result=0
+    return $result
+}
+
+function configure_icon_win_distribution() {
```
Wedds, line 70: Right now, this method will be executed every time during `cp-api-srv` pod startup. But, the `cloud-data` app itself is not changing frequently, besides for the HA deployments such operation seems necessary during the first pod boot up only. It's not that crucial, but would be nice to reduce startup time, where possible.

It could be implemented using the following approach:
1. Create `cloud_data_win_checksum.json`  during `.exe` processing
```json
{
    "executable_sum": "<CHECKSUM_1>",
    "icon_sum": "<CHECKSUM_2>"
}
```
2. Compare values from there with the ones, calculated for the current executable and icon
3. In case JSON file exists, checksum values are matching and updated executable is present - skip icon replacement

Wedds (changes requested):

deploy/docker/cp-api-preconfigure/preconfigure-api
kamyshova: Done

deploy/docker/cp-api-preconfigure/preconfigure-api
```diff
@@ -71,18 +103,46 @@ EOF
...
+    CLOUD_DATA_WIN_ICO_PATH="$CP_PREF_API_STATIC_DIRECTORY_EXT/$CLOUD_DATA_WIN_ICO_NAME"
+
+    CLOUD_DATA_EXE_CHECKSUM=($(sha1sum "${CP_COMMON}/cloud-data-win32-x64/cloud-data.exe"))
+    CLOUD_DATA_ISO_CHECKSUM=($(sha1sum "${CLOUD_DATA_WIN_ICO_PATH}"))
+    CLOUD_DATA_WIN_CHECKSUM_JSON="cloud_data_win_checksum.json"
+
+    check_win_distribution_updates "$CLOUD_DATA_WIN_CHECKSUM_JSON" "$CLOUD_DATA_EXE_CHECKSUM" "$CLOUD_DATA_ISO_CHECKSUM" "Windows"
+    if [ $? -ne 0 ]; then
+        echo "Windows Cloud Data executable will not be changed. Skip configuring."
+        return 0
```
Wedds, line 128: I believe, that `TEMP_PROCESSING_FILE` should be removed before return operation

deploy/docker/cp-api-preconfigure/preconfigure-api
```diff
@@ -71,18 +103,46 @@ EOF
     rm -rf ${_cloud_data_name}-win32-x64
 }
 
+function configure_cloud_data_win_distribution() {
+    TEMP_PROCESSING_FILE="$CP_COMMON/win_processing.txt"
```
Wedds: As I can see now, `$CP_COMMON` by default points to `/opt/api/cp-common`. 
It is a container level directory now, but to synchronize multiple pods' startup in HA mode, `TEMP_PROCESSIG_FILE` should be stored in some shared persistent (host-based) location.

Please, modify deployment file, for `$CP_COMMON` to point onto some host volume.
And maybe:
- some more specific name, like `CP_CLOUD_DATA_PATCHING_DIR`, might be used, because `CP_COMMON` looks similar to `CP_API_COMMON_DIR`
- subdir in `CP_PREF_API_ETC_DIRECTORY` for those purpose could be created

Issue #2574, opened 2022-03-29
`pipe storage cp` shall start data upload before traversing full source hierarchy


sidoruka:
**Background**
At the moment, `pipe storage cp/mv` CLI command for `--recursive` operations runs (at least) in two phases:
* Scan the `source` directory recursively and collect all the files/folders
* Transfer the data collected

For huge filesystem hierarchies the `scan` process may take a lot of time (hours) and consume a lot of memory.

**Approach**
It would be great to change the scanning procedure in a more asynchronous fashion, e.g.:
* Change the scanning process to collect files/folders in batches (e.g. 1000 paths)
* Once we have a batch collected - start upload to the `destination`
* In the meantime (while upload is in progress) collect the next batch of files
* And so on...

sidoruka:
Backport to `release/0.16`

PR #2597, opened 2022-04-11
Issue #2574: pipe storage cp shall start data upload before traversing full source hierarchy (S3 provider)

ekazachkova:
The current PR provides implementation for issue #2574

- `pipe storage cp/mv` operations - this feature enables for:
  - S3 provider only
  - `--verify-destination` shall not be enabled (since this option required full source listing)
  - `--file-list` shall not be enabled
- `pipe storage ls` operation - full listing enables for:
  -  S3 provider only
  - `--recursive` options shall be specified
  - `--show_versions` shall not be specified 
  - `--page` size limit shall not be specified
- `pipe storage rm` operations - added page size support for S3 provider

- batch size can be tuned via `CP_CLI_STORAGE_BATCH_SIZE` environment variable (default: 1000)
- `CP_CLI_STORAGE_ASYNC_BATCH_ENABLE=[True|False] ` - enables asynchronous batch transfer - listing operation performs in separate thread (default: False)


pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -97,8 +99,47 @@ def cp(cls, source, destination, recursive, force, exclude, include, quiet, tags
...
+            cls._transfer_batch_items(source_wrapper, destination_wrapper, items_batch, manager, permission_to_check,
+                                      include, exclude, force, quiet, skip_existing, verify_destination, threads,
+                                      clean, tags, io_threads, next_token)
+
+    @classmethod
+    def _transfer_batch_items(cls, source_wrapper, destination_wrapper, items_batch, manager, permission_to_check,
+                              include, exclude, force, quiet, skip_existing, verify_destination, threads, clean, tags,
+                              io_threads, next_token):
+        while True:
+            workers = []
```
tcibinan: The list is pretty confusing. We accumulate background processes here over time but we only really need to wait for one or two processes in `cls._handle_keyboard_interrupt` calls:
- transfer
- transfer + listing

What do you think if we get rid of the list and just wait for specific processes?

```python
cls._handle_keyboard_interrupt([transfer_process])
# or
cls._handle_keyboard_interrupt([transfer_process, listing_process])
```

pipe-cli/src/utilities/storage/s3.py
```diff
@@ -625,6 +631,22 @@ def list_items(self, relative_path=None, recursive=False, page_size=StorageOpera
         else:
             return self.list_objects(client, prefix, operation_parameters, recursive, page_size)
 
+    def list_paging_items(self, relative_path=None, recursive=False, page_size=S3_DEFAULT_BATCH_SIZE,
```
tcibinan: This method seems very similar to `self.list_items`. What do you think if we merge these two methods and always return `items, next_token` from `self.list_items`?

pipe-cli/src/utilities/storage/s3.py
```diff
@@ -932,6 +978,42 @@ def get_items(cls, storage_wrapper, session=None):
                         continue
                     yield ('File', file['Key'], cls.get_prefix(delimiter, name), file['Size'])
 
+    @classmethod
+    def get_paging_items(cls, storage_wrapper, page_size, next_token=None, session=None, results=None):
```
tcibinan: I think that we can merge the method with `cls.get_items`. The `results` argument can be removed and used in the call site in the following fashion.

```python
queue = multiprocessing.Queue()
def call():
    queue.put(source_wrapper.get_items())
process = multiprocessing.Process(target=call)
output = queue.get()
```

tcibinan (approved):

PR #2612, opened 2022-04-18
Support source deletion in data transfer service

tcibinan:
Relates to #2044.

The pull request brings support for optional source deletion to data transfer service autonomous synchronization. It allows to delete already transferred local files and directories.

The following command can be used to enable autonomous transfer source deletion for all syncing paths.

```bash
pipe dts preferences update dtsname -p 'dts.source.deletion.enabled=true'
```

The following command can be used to enable autonomous transfer source deletion for specific syncing paths.

```bash
pipe dts preferences update "$DTS_NAME" -p 'dts.local.sync.rules=[{
                                                "source": "c:\\local\\path\\to\\source\\directory",
                                                "destination": "s3://data/storage/path/to/destination/directory",
                                                "cron": "0 0/1 * ? * *",
                                                "deleteSource": true
                                            }, {
                                                "source": "c:\\Program Files\\CloudPipeline\\DTS\\logs",
                                                "destination": "s3://data/storage/path/to/logs/directory",
                                                "cron": "0 0/1 * ? * *"
                                            }]'
```

### DIFF
--- a/cloud-pipeline-common/model/src/main/java/com/epam/pipeline/cmd/PipelineCLIImpl.java
+++ b/cloud-pipeline-common/model/src/main/java/com/epam/pipeline/cmd/PipelineCLIImpl.java
@@ -73,9 +73,7 @@ public class PipelineCLIImpl implements PipelineCLI {
 
             while (attempts < retryCount) {
                 try {
-                    final String command = deleteSource
-                            ? pipeMV(source, destination, include)
-                            : pipeCP(source, destination, include);
+                    final String command = buildPipeTransferCommand(source, destination, include, deleteSource);
                     cmdExecutor.executeCommand(command, username);
                     log.info(String.format("Successfully uploaded from %s to %s", source, destination));
                     return destination;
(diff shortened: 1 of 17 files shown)


cloud-pipeline-common/model/src/main/java/com/epam/pipeline/cmd/PipelineCLIImpl.java
```diff
@@ -71,7 +73,10 @@ public String uploadData(String source, String destination, List<String> include
 
             while (attempts < retryCount) {
                 try {
-                    cmdExecutor.executeCommand(pipeCP(source, destination, include), username);
+                    final String command = deleteSource
```
Wedds: Both `PIPE_MV_TEMPLATE` and `PIPE_CP_TEMPLATE` has the similar structure, so `pipeMV` and `pipeCP` methods are similar, as well. 
Maybe we can generalize them to `buildPipeTransferCommand` and pass template as an additional argument?

Wedds (changes requested):

cloud-pipeline-common/model/src/main/java/com/epam/pipeline/cmd/PipelineCLIImpl.java
tcibinan: Done.

Wedds (approved):

PR #2613, opened 2022-04-20
Make ACL expression for storage items' tags update operations less strict

Wedds:
This PR slightly changes security constraints for 2 operations: storage items' tags update and removal.

For the majority of secured entities, tags update available for admins or owner only. 
In case of storage items, it makes sense to allow such modification for users with write access to the storage, since if a user is allowed to modify storage items - its tagging should be supported as well.


mzueva (approved):

tcibinan:
Cherry-picked to `release/0.16` via 1a43b332919daae8bec2e8c3ad14bd060acb53dd.

PR #2618, opened 2022-04-26
Implement transferring unsafe paths management in pipe cli

### LINKED ISSUE #2617: Manage transferring unicode storage paths explicitly in pipe cli
**Background**

Currently pipe storage `cp/mv` commands fail on linux if some of the transferring files have non ascii symbols in their paths. The issue is related to #1334 where the same problem is partly resolved by changing the default encoding in python. At the same time more common approach would be to manage unicode paths explicitly in order to avoid multiple issues with subsequent processing.

**Approach**

Multiple strategies should be introduced for unicode paths management in pipe storage `cp/mv` commands.

tcibinan:
Resolves #2617.

The pull request brings support for unsafe paths management to pipe storage `cp/mv` commands. From now on all unsafe (non-ascii) paths can be managed explicitly via multiple pipe storage `cp/mv` command arguments.

```
  --on-unsafe-chars [fail|skip|replace|remove|allow]
                                  Configure how unsafe characters in file paths should be handled.
                                  By default only ascii characters are safe.
                                  Allowed values:
                                  [fail] fails immediately; 
                                  [skip] skips paths with unsafe characters (default);
                                  [replace] replaces unsafe characters in paths; 
                                  [remove] removes unsafe characters from paths; 
                                  [allow] allows unsafe characters in paths.
  --on-unsafe-chars-replacement TEXT
                                  Specify a string to replace unsafe characters with.
                                  The option has effect only if --unsafe-chars option is set to replace value.
```

Same arguments can be configured via the following environment variables respectively:
- `CP_CLI_TRANSFER_UNSAFE_CHARS`
- `CP_CLI_TRANSFER_UNSAFE_CHARS_REPLACEMENT`

### Additional changes

Additionally the following pipe storage `cp/mv` commands argument has been added.

```
  --on-failures [fail|fail-after|skip]
                                  Configure how singular file processing failures should affect overall command execution.
                                  Allowed values: 
                                  [fail] fails immediately (default); 
                                  [fail-after] fails only after all files are processed; 
                                  [skip] skips all failures.
```

Same argument can be configured via the following environment variable:
- `CP_CLI_TRANSFER_FAILURES`

### DIFF
--- a/pipe-cli/src/utilities/datastorage_operations.py
+++ b/pipe-cli/src/utilities/datastorage_operations.py
@@ -193,6 +193,7 @@ class DataStorageOperations(object):
                     relative_path = to_ascii(relative_path, removing=True)
                 else:
                     logging.warn(u'Ignoring unsafe characters in path {}...'.format(full_path))
+                item = (item[0], full_path, relative_path, source_size)
 
             if relative_path.endswith(FOLDER_MARKER):
                 filtered_items.append(item)
--- a/pipe-cli/src/utilities/datastorage_operations.py
+++ b/pipe-cli/src/utilities/datastorage_operations.py
@@ -12,14 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import datetime
+import logging
 import multiprocessing
 import os
 import platform
-import sys
 
 import click
+import datetime
 import prettytable
+import sys
 from future.utils import iteritems
 from operator import itemgetter
 
@@ -30,9 +31,10 @@ from src.model.data_storage_wrapper import DataStorageWrapper, S3BucketWrapper
 from src.model.data_storage_wrapper_type import WrapperType
 from src.utilities.du import DataUsageHelper
 from src.utilities.du_format_type import DuFormatType
+from src.utilities.encoding_utilities import to_string, is_safe_chars, to_ascii
 from src.utilities.hidden_object_manager import HiddenObjectManager
 from src.utilities.patterns import PatternMatcher
-from src.utilities.storage.common import TransferResult, UploadResult
+from src.utilities.storage.common import TransferResult
 from src.utilities.storage.mount import Mount
 from src.utilities.storage.umount import Umount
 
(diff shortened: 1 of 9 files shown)


NShaforostov:
I've tried to verify.
Currently, for unsafe paths `--on-unsafe-chars` option works fine (as described) with [`fail`|`skip`|`allow`] values.
With [`replace`|`remove`] values works fully the same as with `allow` option - i.e. all files despite of unsafe chars are being copied/moved with the original names, unsafe chars are not being removed or replaced. Setting of the `--on-unsafe-chars-replacement` option applies nothing.

Was tested for `pipe cp/mv` commands on Ubuntu:18.04, Centos:7, Windows, WSL

pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -99,46 +116,90 @@ def cp(cls, source, destination, recursive, force, exclude, include, quiet, tags
...
-                      include, exclude, force, quiet, skip_existing, verify_destination):
+                      include, exclude, force, quiet, skip_existing, verify_destination,
+                      unsafe_chars, unsafe_chars_replacement):
         filtered_items = []
         for item in items:
             full_path = item[1]
             relative_path = item[2]
             source_size = item[3]
 
+            if not is_safe_chars(relative_path):
```
ekazachkova: Maybe we can extract this logic to separate method? just to increase readability?

pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -588,28 +649,52 @@ def _multiprocess_transfer_items(cls, sorted_items, threads, manager, source_wra
...
+            try:
+                transfer_result = manager.transfer(source_wrapper, destination_wrapper, path=full_path,
+                                                   relative_path=relative_path, clean=clean, quiet=quiet, size=size,
+                                                   tags=tags, io_threads=io_threads, lock=lock)
+                if not destination_wrapper.is_local() and transfer_result:
+                    transfer_results.append(transfer_result)
+                    transfer_results = cls._flush_transfer_results(source_wrapper, destination_wrapper,
+                                                                   transfer_results, clean=clean)
+            except Exception as e:
+                if on_failures == AllowedFailuresValues.FAIL:
```
ekazachkova: Is it possible to unify this pice of code and [this](https://github.com/epam/cloud-pipeline/pull/2618/files#diff-e4e3d5936cbd7e5d07fbdc6e346208d3677b3910de8a7e1bed2d65a5984e2b37R140) and extract separate method. As far as I can see there are diffs with logging but I'm not sure how critical that are.

pipe-cli/src/utilities/storage/s3.py
```diff
@@ -270,8 +275,10 @@ def transfer(self, source_wrapper, destination_wrapper, path=None, relative_path
         source_key = self.get_source_key(source_wrapper, path)
         destination_key = self.get_destination_key(destination_wrapper, relative_path)
 
-        tags += ("CP_SOURCE={}".format(source_key),)
-        tags += ("CP_OWNER={}".format(self._get_user()),)
+        tags += ("{}={}".format(StorageOperations.CP_SOURCE_TAG,
+                                source_key if is_safe_chars(source_key)
```
ekazachkova: it looks like such logic repeats in this [line](https://github.com/epam/cloud-pipeline/pull/2618/files#diff-cbdb173c35573ef86081cbc46790e9435e6dcba6620b07e389585715cd321f92R191) and [here](https://github.com/epam/cloud-pipeline/pull/2618/files#diff-35839fd2c17f3300a68f380dddae3145cbc10439b2aa9de2045ca0794b7d51b8R228)  
Maybe we can extract common method?

pipe-cli/src/utilities/datastorage_operations.py
tcibinan: Done

pipe-cli/src/utilities/datastorage_operations.py
tcibinan: Done

pipe-cli/src/utilities/storage/s3.py
tcibinan: Done

ekazachkova (approved):

PR #2621, opened 2022-04-28
GUI Library (Folders): render JSON-type attributes in the table the s…

AleksandrGorodetskii:
…ame way as in the split-panel


rodichenko:
Merged with 7896eb4c

PR #2626, opened 2022-05-13
Add NFS events synchronisation deployment

tcibinan:
Relates to #2155.

The pull request brings a separate search service kubernetes deployment solely for nfs events synchronisation. Additionally the request brings support for utf-8 paths processing to nfs events synchronisation.

It is also recommended to merge #2236 along with the current one as long as it removes configuration parameters differences between multiple Cloud Pipeline services.

### DIFF
--- a/deploy/contents/k8s/cp-search/cp-search-srv-dpl.yaml
+++ b/deploy/contents/k8s/cp-search/cp-search-srv-dpl.yaml
@@ -26,6 +26,9 @@ spec:
           command: ["/init"]
           ports:
             - containerPort: 8080
+          env:
+            - name: CP_SEARCH_DISABLE_NFS_OBSERVER_SYNC
+              value: "true"
           envFrom:
           - configMapRef:
               name: cp-config-global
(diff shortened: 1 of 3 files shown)


deploy/contents/k8s/cp-search/cp-search-srv-dpl.yaml
```diff
@@ -26,6 +26,9 @@ spec:
           command: ["/init"]
           ports:
             - containerPort: 8080
+          env:
+            - name: CP_SEARCH_DISABLE_NFS_OBSERVER_SYNC
+              value: "true"
```
Wedds, line 33: 1. Since we are changing this deployment file anyway, let's add the following entry:
```yaml
- name: CP_SEARCH_DISABLE_NFS_FILE
  value: "false"
```
It is a default value, that might be handy to have in place, so we can easily change the deployment file, when the environment should have a separate NFS synchronizer.

2. Maybe we should add `LANG` env var as well

Wedds (changes requested):

deploy/contents/k8s/cp-search/cp-search-srv-dpl.yaml
tcibinan: Both variables were added to the deployment file.

Wedds (approved):

mzueva (approved):

PR #2637, opened 2022-05-23
Remove repository configurations in launch.sh

kamyshova:
The PR contains removing `cuda.list` and `nvidia-ml.list` repositories configurations for Ubuntu distribution in `launch.sh`.
`CP_REMOVE_REP_CONFIGURATION` environment variable was added to enable/disable removal.


sidoruka:
This looks, but a bit rigid. Let's make it a bit more flexible.
Instead of being just a boolean, let's rename the parameter to `CP_CAP_REMOVE_FILES` and make it a list (comma-delimited) of the paths to be dropped.
`CP_CAP_REMOVE_FILES` shall be set to `/etc/apt/sources.list.d/cuda.list,/etc/apt/sources.list.d/nvidia-ml.list` by default.
In this case - `launch.sh` will delete everything listed in `CP_CAP_REMOVE_FILES`. If it is not set or it is empty - do nothing

Issue #2641, opened 2022-05-25
Allow to limit number of active runs for a user/group


mzueva:
**Background**
It would be helpful to limit number of active runs for a user/group using contextual preferences.

PR #2645, opened 2022-05-30
Issue #2641 Implement user/group-level run launch limits 

Wedds:
This PR is related to issue #2641

It allows configuring launch limits for a specific user or group.
In case "static" part of submitted run (master + workers count) potentially exceeds any limits - an exception with limit details is thrown and launch is being aborted.


api/src/main/java/com/epam/pipeline/manager/quota/RunLimitsService.java
```diff
@@ -0,0 +1,176 @@
...
+    private final MessageHelper messageHelper;
+    private final AuthManager authManager;
+
+    public void checkRunLaunchLimits(final Integer newInstancesCount) {
+        if (authManager.isAdmin()) {
+            return;
+        }
+        final PipelineUser user = authManager.getCurrentUser();
+        final String userName = user.getUserName();
+        final List<ContextualPreference> allPreferences = contextualPreferenceManager.loadAll();
```
mzueva: I'd suggest to load only group/user related preferences, even if it will be two requests. We can have thousands of preferences and it seems to be overwhelming to load everything for such check.

mzueva (changes requested):

mzueva (approved):

PR #2650, opened 2022-06-02
Allow specifying availability zone for AWS runs

Wedds:
This PR brings new system `CP_CAP_TARGET_AVAILABILITY_ZONE` parameter, that allows specifying target availability zone at startup.

If no matches for target zone occurred during comparing with the loaded ones in node up script - the current logic is being used.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2021 EPAM Systems, Inc. (https://www.epam.com/)
+ * Copyright 2017-2022 EPAM Systems, Inc. (https://www.epam.com/)
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
```diff
@@ -477,6 +480,7 @@ private void createNodeForRun(List<CompletableFuture<Void>> tasks, String runId,
                                       InstanceRequest requiredInstance) {
             long longId = Long.parseLong(runId);
             addNodeUpTask(longId);
+            setAvailabilityZoneIfSpecified(requiredInstance, longId);
```
mzueva: Maybe we can set AZ in method getNewRunInstance where PipelineRun is already loaded instead of setting it separately?

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
Wedds: Done

mzueva (approved):

Issue #2582, opened 2022-03-31
Cannot delete user who is an OWNER of any object


sidoruka:
**Describe the bug**
Cannot delete user who is an OWNER of any object

**To Reproduce**
1. Create `USER_X`
2. Grant this user OWNER access to any Cloud Pipeline object
3. Delete user
4. `SQL Error` message is shown, user is not deleted

**Expected behavior**
User is deleted and the object keeps track of the ownership

sidoruka:
Please backport to `release/0.16`

PR #2670, opened 2022-06-19
Issue 2582: Cannot delete user who is an OWNER of any object

okolesn:
**Background**
[Cannot delete user who is an OWNER of any object #2582](https://github.com/epam/cloud-pipeline/issues/2582)

**Changes**
delete corresponding acl_object_identity record when delete user


mzueva:
Fixed

PR #2731, opened 2022-07-13
Add support for regional desktop connections

tcibinan:
Relates to #1715.

The pull request brings support for regional nomachine/nicedcv desktop connections to both linux and windows runs. Depending on the chosen region the desktop connections will use regional proxies from now on. Previously default region proxy was used no matter what.


workflows/pipe-common/scripts/serve_desktop.py
```diff
@@ -0,0 +1,230 @@
...
+app = Flask(__name__)
+
+
+@app.route('/')
+def get_desktop_file():
+    proxy_host, proxy_port = _resolve_edge(Config.api)
+
+    template_data, template_type = _read_template_file(Config.template_path)
+
+    if Config.personal and is_windows():
```
SilinPavel: I would move it to separate (username resolving) method similar to `_read_template_file` and `_resolve_edge` just to make it more clean

SilinPavel (changes requested):

workflows/pipe-common/scripts/serve_desktop.py
tcibinan: Done

SilinPavel (approved):

tcibinan:
Cherry-picked to `release/0.16` via 59f14fb4a963c23c50a3aa21f8538270becc1ee4.

PR #2795, opened 2022-08-26
Issue 2721 restore datastorage folder from 'archived' storage class

SilinPavel:
This PR adds backend part of functionality to restore folder in datastorage and brings such things like:

 - Initiate restore action for `folder` in `datastorage`
 - Update restore action with new status and metadata
 - Filter restore actions by `path` (could match all `path`s that include the given one, or match all `path`s that included in the given one) and `status`es
 - Load `effective` restore action. Effective action is an action that has `active` status (`INITIATED`, `RUNNING`, `SUCCEEDED`) and is the last one


api/src/main/java/com/epam/pipeline/controller/datastorage/lifecycle/DataStorageLifecycleController.java
```diff
@@ -221,4 +225,88 @@ public Result<List<StorageLifecycleRuleExecution>> listStorageLifecyclePolicyRul
                 dataStorageLifecycleApiService.listStorageLifecyclePolicyRuleExecutions(
                         datastorageId, ruleId, path, status));
     }
+
+    @PostMapping(value = "/datastorage/{datastorageId}/lifecycle/restore")
+    @ResponseBody
```
mzueva: I missed this in the initial request - let's use `@RestController` to get rid of multi`@ResponseBody` annotations

api/src/main/java/com/epam/pipeline/manager/datastorage/lifecycle/DataStorageLifecycleRestoreManager.java
```diff
@@ -0,0 +1,254 @@
...
+        return ListUtils.union(
+                rootActionIsActive ? Collections.singletonList(root) : Collections.emptyList(),
+                relatedActions
+        );
+    }
+
+    @Transactional
+    public void deleteRestoreActions(final Long datastorageId) {
+        dataStoragePathRestoreActionRepository.deleteByDatastorageId(datastorageId);
+        dataStoragePathRestoreActionRepository.flush();
```
mzueva, line 183: Do we really need to manually flush?

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/controller/datastorage/lifecycle/DataStorageLifecycleController.java
SilinPavel: Fixed

api/src/main/java/com/epam/pipeline/manager/datastorage/lifecycle/DataStorageLifecycleRestoreManager.java
SilinPavel: Yes, it's because we need to force hibernate to propagate changes that were done to DB, because this method is used by StorageManager when it deletes some storage and StorageManager uses jdbc to connect to db, so as I see there are two ways:
 - `flush` hibernate changes to DB, in this case this changes will still under common transaction 
 - provide for this method `propogation = REQUIRED_NEW` - but here we will commit all changes to db
 
 Maybe I miss something and there is another convenient way to do that

Issue #2721, opened 2022-07-07
S3 objects lifecycle management


mzueva:
**Background**

Requirements to the S3 objects lifecycle management :
- Allow to specify multiple transition rules for a bucket based on file prefix and/or glob pattern
- All available S3 storage archive classes shall be supported (`S3 Glacier Instant Retrieval`, `S3 Glacier Flexible Retrieval`, `S3 Glacier Deep Archive`) as well as deletion of objects
- User shall receive a notification on lifecycle events:
  - User shall receive a notification when data is close up to N days to transition. Notification shall be sent with configurable delay
  - User shall have an option to delay data transition using link from notification email
  - User shall receive a notification once data transition is initiated
  - Default template shall be available for such notifications, also notifications (text, subject, delays) shall be configurable per bucket
- User shall be able to restore archived files and folders using CLI and GUI

### Approach

- A new object `LifecyclePolicy` shall be attached to an S3 bucket. It is similar to AWS `BucketLifecycleConfiguration` object and contains set of transition rules including filters (prefix, tags), storage class and transition days. This object shall also contain notification settings. E.g. in this object we specify that files with `StorageType=bcl` tag shall be deleted after 90 days and files with `StorageType=fastq` tag shall be moved to `S3 Glacier Deep Archive` after 180 days and deleted after 5 years.
- When files are uploaded to storage after pipeline execution or from GUI/CLI files shall be automatically tagged. Tagging preferences shall be defined in a System Preference, e.g. `[{"*.bcl" : "StorageType=bcl"}]` or as a json file for a pipeline. Note no more than 10 tags are allowed. Upload API including generate URL api methods shall support tagging from SystemPreference as well.
- Pipe CLI shall provide command for object native tagging using S3 batch requests

**Object Lifecycle Monitor**

A new daemon service shall monitor buckets with `LifecyclePolicy` attached using the following algorithm:
  - Monitoring is done for a configured folder in the bucket. Each folder under this prefix is considered to be a `dataset`. For a file structure below we have tree datasets (`run1`, `run2`, `run3`) for prefix `data/`
```
 bucket/
     data/
         run1/
         run2/
         run3/
```
   - Each dataset is processed individually. When a new dataset is detected a new `LifecyclePolicy` shall be created for this dataset from default template attached to the storage. A dataset without assigned `LifecyclePolicy` is considered to be new
   - Each file in the dataset is checked against the lifecycle policy. File is considered eligible when:
[...]

PR #2811, opened 2022-09-05
2721 Storage Lifecycle Service: Restore objects

SilinPavel:
Related to #2721 
Depends on #2795 

This PR adds to `Storage Lifecycle Service` functionality for restoring files and folders from `archive` storage class.

See https://github.com/epam/cloud-pipeline/issues/2721#issuecomment-1237035186 for restore process description


SilinPavel:
Should be merged after #2795

mzueva (approved):

Issue #2787, opened 2022-08-17
[HCS Analysis] Enhancements


maryvictol:
It would be helpful for users

- [x] to fill ***Parent Objects*** (*Nuclei objects*,  *Cells objects*, etc) where it 's possible when user adds *Main* modules.
- [x] to disable ***Run analysis*** button if the required *Parent Objects* (*Nuclei objects*,  *Cells objects*, etc) aren't filled.
- [x] to give any *Warning message* when *Analysis* return no results.

---
- [x] At case of single analysis ***Run analysis*** button shall be disabled when no modules added.
- [x] *Objects names* shall be checked for uniqueness.
- [x] User shall have the ability to unselect time point, well, field by second click (with Shift) at multi time points (wells, fields) selection.

maryvictol:
@AleksandrGorodetskii, @rodichenko 
**Bug:** 
HCS Analysis: On the *Evaluations* tab tooltips for *job status icon* and *job owner* are cut off by the browser edge.
![image](https://user-images.githubusercontent.com/47791878/188419963-de88e2e6-5833-4cec-a7aa-2e003adc128a.png)

rodichenko:
> * to disable _**Run analysis**_ button if the required _Parent Objects_ (_Nuclei objects_,  _Cells objects_, etc) aren't filled.

@maryvictol should be fixed with 14a10775cc7cf56225276cf5efe5d3de03da077f

rodichenko:
> * to give any _Warning message_ when _Analysis_ return no results.

@maryvictol should be fixed with d917184eb5b5aa61ece18d66260b020970f7dc0c

rodichenko:
> * User shall have the ability to unselect time point, well, field by second click (with Shift) at multi time points (wells, fields) selection.
> * _Objects names_ shall be checked for uniqueness.
> * At case of single analysis _**Run analysis**_ button shall be disabled when no modules added.

@maryvictol shall be fixed with 9b84d78e3ea3c526e36f52d3751afdb3f8f08ec6

maryvictol:
> > * User shall have the ability to unselect time point, well, field by second click (with Shift) at multi time points (wells, fields) selection.
> > *...
> 
> @maryvictol shall be fixed with [9b84d78](https://github.com/epam/cloud-pipeline/commit/9b84d78e3ea3c526e36f52d3751afdb3f8f08ec6)

Verified as fixed

rodichenko:
[...]

PR #2816, opened 2022-09-07
[HCS Analysis] Enhancements (#2787) - disable run button if required …

AleksandrGorodetskii:
…parameters fields are empty


rodichenko:
Merged with 14a10775

Issue #2859, opened 2022-10-05
Introduce read only storages support in webdav


tcibinan:
**Background**

Currently a user can access storages if given read/write permissions using Cloud Data or pipe fuse tools which both use Cloud Pipeline Webdav endpoint to access its data. It would be helpful to also support storages with read only permission in Cloud Pipeline Webdav. This way users will be able to access read only storages through Cloud Data and pipe fuse.

**Approach**

Introduce read only storages support in Cloud Pipeline Webdav.

tcibinan:
Cherry-picked to `release/0.16` via `ef26c73bd66e0c5e238a834fa75e0be960af7092`, `d7b380bd91050fc5185584fe10a2b0e1360fcc3a`.

tcibinan:
Cherry-picked to `release/0.16` via ef26c73bd66e0c5e238a834fa75e0be960af7092, d7b380bd91050fc5185584fe10a2b0e1360fcc3a, 46ac86a05274e18f31b79643743a317e3d5c932f and b8ce0f41c34af1dbab375e2af527dcab9e4350e8.

PR #2891, opened 2022-10-20
Add support for users blocking/unblocking in webdav

tcibinan:
Relates to #2859.

The pull request brings support for users blocking/unblocking in webdav.

### DIFF
--- a/scripts/nfs-roles-management/internal/synchronization/synchronization.py
+++ b/scripts/nfs-roles-management/internal/synchronization/synchronization.py
@@ -18,6 +18,7 @@ import re
 import subprocess
 
 from internal.api.storages_api import Storages
+from internal.api.users_api import Users
 from internal.config import Config
 from internal.model.mask import Mask, FullMask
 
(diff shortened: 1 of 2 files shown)


scripts/nfs-roles-management/internal/synchronization/synchronization.py
```diff
@@ -122,6 +127,10 @@ def synchronize_user(self, user, use_symlinks=False):
                     if storage_user.username.strip().lower() == user.strip().lower():
                         syncing_storages[storage] = storage_user_mask
 
+            if not details or details.blocked:
```
SilinPavel, line 130: just minor comment on naming, as for me `user_details` would be easier to follow here.

scripts/nfs-roles-management/internal/synchronization/synchronization.py
```diff
@@ -122,6 +127,10 @@ def synchronize_user(self, user, use_symlinks=False):
...
+                logging.warning('Skipping user {} storages linking because the user is blocked...'.format(user))
```
SilinPavel, line 131: How about to add `details` dict to log message or some how mark in a log message what was the case here, user was blocked or details isn't provided?

Issue #2912, opened 2022-11-03
Support folder browsing for static resources


mzueva:
**Background**
At the moment Cloud Pipeline provides `/restapi/static-resources/${path}` endpoint for browsing files from S3 objects in a browser. For folders this API at the moment returns 404, even if folder exists.

**Approach**
API `/restapi/static-resources/${path}` shall return simple html with files list for folders.

![image](https://user-images.githubusercontent.com/12640002/199772842-a5033cab-b04b-40a3-b2cd-1ce759c18e09.png)

PR #2917, opened 2022-11-08
Issue 2912: Support folder browsing for static resources

okolesn:
**Background**
[Support folder browsing for static resources #2912](https://github.com/epam/cloud-pipeline/issues/2912)

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -249,6 +249,9 @@ dependencies {
     compile('io.reflectoring.diffparser:diffparser:1.4')
 
     compile 'dnsjava:dnsjava:3.4.3'
+
+    //Template engine
+    compile group: 'org.freemarker', name: 'freemarker', version: '2.3.23'
 }
 
 // >>>>> processes profiles
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -46,7 +46,7 @@ task copyLocalDtsStartupScripts(type: Copy) {
 }
 
 task generateComponentsVersions(type: Exec) {
-    commandLine "bash", 
+    commandLine "bash",
                 "$project.rootDir/scripts/pipeline-scripts/generate_components_versions.sh",
                 "$project.rootDir",
                 "$rootDir/api/src/main/resources/static/components_versions.json"
@@ -251,7 +251,7 @@ dependencies {
     compile 'dnsjava:dnsjava:3.4.3'
 
     //Template engine
-    compile group: 'org.freemarker', name: 'freemarker', version: '2.3.23'
+    compile group: 'org.apache.velocity', name: 'velocity', version: '1.7'
 }
 
 // >>>>> processes profiles
--- a/api/src/main/java/com/epam/pipeline/manager/resource/StaticResourcesService.java
+++ b/api/src/main/java/com/epam/pipeline/manager/resource/StaticResourcesService.java
@@ -40,8 +40,6 @@ import org.springframework.util.Assert;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.StringWriter;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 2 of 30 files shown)


api/build.gradle
```diff
@@ -249,6 +249,9 @@ dependencies {
     compile('io.reflectoring.diffparser:diffparser:1.4')
 
     compile 'dnsjava:dnsjava:3.4.3'
+
+    //Template engine
+    compile group: 'org.freemarker', name: 'freemarker', version: '2.3.23'
```
mzueva: Could you please try to use `Velocity` - we already have in the dependencies, so I'd prefer not to add one more template manager here.

api/src/main/java/com/epam/pipeline/manager/resource/StaticResourcesService.java
```diff
@@ -19,22 +19,49 @@
...
+import java.util.stream.Collectors;
+
+import static com.epam.pipeline.controller.resource.StaticResourcesController.STATIC_RESOURCES;
 
 @Service
 @RequiredArgsConstructor
 public class StaticResourcesService {
 
     private static final String DELIMITER = "/";
+    private static final String TEMPLATE = "folder.ftlh";
```
mzueva: Could you please move the settings (template folder, template name) to `SystemPreferences`?

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/resource/StaticResourcesService.java
```diff
@@ -46,7 +68,54 @@ public DataStorageStreamingContent getContent(final String path) {
         Assert.isTrue(StringUtils.isNotBlank(filePath),
                 messageHelper.getMessage(MessageConstants.ERROR_STATIC_RESOURCES_INVALID_PATH));
         final AbstractDataStorage storage = dataStorageManager.loadByNameOrId(bucketName);
+        if (Files.isDirectory(Paths.get(path))) {
+            final List<AbstractDataStorageItem> items = dataStorageManager.getDataStorageItems(storage.getId(),
+                    path, false, null, null).getResults();
```
mzueva: Fix `path` -> `filePath`

api/src/main/java/com/epam/pipeline/manager/resource/StaticResourcesService.java
```diff
@@ -46,7 +68,54 @@ public DataStorageStreamingContent getContent(final String path) {
...
         dataStorageManager.checkDataStorageObjectExists(storage, filePath, null);
         return dataStorageManager.getStreamingContent(storage.getId(), filePath, null);
     }
+
+    public static String buildHtml(final List<AbstractDataStorageItem> items,
+                            final String templatePath,
+                            final String templateName,
+                            final String staticResourcesPrefix) throws IOException{
+        final VelocityEngine engine = new VelocityEngine();
+        engine.setProperty(RuntimeConstants.FILE_RESOURCE_LOADER_PATH, templatePath);
```
mzueva: With this configuration, classpath template is not loaded. We need to support both classpath and filesystem resources. Maybe it is easier to use an approach from `SMTPNotificationManager`  where template is passed as a string to `Velocity.evaluate` method? In this case we can read template using `ResourceUtils` class or similar. In this case we can leave only one preference `STATIC_RESOURCES_FOLDER_TEMPLATE_PATH ` with full path to template, e.g. `classpath/views/folder.vm` or `/opt/api/templates/folder.html`

api/src/main/java/com/epam/pipeline/manager/resource/StaticResourcesService.java
```diff
@@ -46,7 +68,54 @@ public DataStorageStreamingContent getContent(final String path) {
         Assert.isTrue(StringUtils.isNotBlank(filePath),
                 messageHelper.getMessage(MessageConstants.ERROR_STATIC_RESOURCES_INVALID_PATH));
         final AbstractDataStorage storage = dataStorageManager.loadByNameOrId(bucketName);
+        if (Files.isDirectory(Paths.get(path))) {
```
mzueva: This check won't work for cloud resource, I will add a new method to check cloud object type

mzueva (changes requested):

mzueva (approved):

PR #2925, opened 2022-11-11
Search for storage objects by specific tag

SilinPavel:
This PR provides functionality to search Data Storage Items by tag

### DIFF
--- a/api/src/main/resources/dao/datastorage-tag-dao.xml
+++ b/api/src/main/resources/dao/datastorage-tag-dao.xml
@@ -178,24 +178,13 @@
         </constructor-arg>
         <constructor-arg>
             <value>
-                <!--searchDatastorageObjectsByTagQuery-->
-                <![CDATA[
-                    SELECT *
-                    FROM pipeline.datastorage_tag t
-                    WHERE t.tag_key = :TAG_KEY
-                    AND (:TAG_VALUE::text IS NULL OR t.tag_value = :TAG_VALUE)
-                ]]>
-            </value>
-        </constructor-arg>
-        <constructor-arg>
-            <value>
                 <!--searchDatastorageObjectsByTagInSpecificStoragesQuery-->
                 <![CDATA[
                     SELECT *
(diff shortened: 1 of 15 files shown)


api/src/main/resources/dao/datastorage-tag-dao.xml
```diff
@@ -176,5 +176,28 @@
...
+                <!--searchDatastorageObjectsByTagInSpecificStoragesQuery-->
+                <![CDATA[
+                    SELECT *
+                    FROM pipeline.datastorage_tag t
+                    WHERE t.tag_key = :TAG_KEY
+                    AND (:TAG_VALUE::text IS NULL OR t.tag_value = :TAG_VALUE)
+                    AND t.datastorage_root_id IN (:DATASTORAGE_ROOT_ID)
+                ]]>
+            </value>
+        </constructor-arg>
```
tcibinan: Is it possible to use just a single sql query for both cases? By checking *DATASTORAGE_ROOT_ID* nullability similarly to *TAG_VALUE*.

workflows/pipe-common/pipeline/api/api.py
```diff
@@ -1140,6 +1147,76 @@ def prolong_lifecycle_rule(self, datastorage_id, rule_id, path, days, force=Fals
             raise RuntimeError("Failed to prolong lifecycle rule '{}'.",
                                "Error message: {}".format(str(rule_id), str(e.message)))
 
+    def load_datastorage_item_content(self, storage_id, path, version=None):
+        try:
+            formed_url = self.LOAD_STORAGE_ITEM_CONTENT_URL.format(id=storage_id, path=path)
+            if version:
+                formed_url = formed_url + "&version={version}".format(version=version)
+            return self.execute_request(str(self.api_url) + formed_url)
```
tcibinan: It is better to use `self._request` method rather than `self.execute_request` because it provides more efficient error handling. The same applies for all the methods below.

PR #2944, opened 2022-11-28
Datastorage object: Search by tag refactoring

SilinPavel:
This PR includes:
 - Changes to load bounded number of entries in NFSProvider if pageSize isn't specified
 - Link to issue about enchantment regarding NFS storage object tagging
 - Changes regarding permission checking (should increase performance because we don't load storage several times)
 - Fix bug with NFSSynchronizer in elastichsearch-agent - now tags are indexed in a right way

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
@@ -146,6 +146,7 @@ public class DataStorageManager implements SecuredEntityManager {
     private static final String DEFAULT_USER_STORAGE_DESCRIPTION_TEMPLATE = "Home folder for user @@";
     public static final String DAV_MOUNT_TAG = "dav-mount";
     private static final String SHARED_STORAGE_SUFFIX = "share";
+    private static final String READ_PERMISSION = "READ";
 
     @Autowired
     private MessageHelper messageHelper;
@@ -571,7 +572,7 @@ public class DataStorageManager implements SecuredEntityManager {
     private List<String> adjustPermissions(final List<String> permissions) {
         return Optional.ofNullable(permissions)
                 .filter(CollectionUtils::isNotEmpty)
-                .orElseGet(() -> Collections.singletonList("READ"));
+                .orElseGet(() -> Collections.singletonList(READ_PERMISSION));
     }
 
     private Duration resolveDuration(final long hours) {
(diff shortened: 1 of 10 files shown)


api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -778,16 +779,19 @@ public List<DataStorageTagSearchResult> searchDataStorageItemByTag(
             final Map<Long, List<DataStorageTag>> results = new HashMap<>();
             final Map<Long, List<AbstractDataStorage>> storagesByRootId = dataStorageDao.loadDataStoragesByRootIds(
                     rootSearchResult.keySet()
-            ).stream().filter(
-                ds -> CollectionUtils.isEmpty(storageIdsToFilter) || storageIdsToFilter.contains(ds.getId())
-            ).collect(Collectors.groupingBy(AbstractDataStorage::getRootId));
+            ).stream()
+                    .filter(dataStorage -> permissionManager.storagePermission(dataStorage, READ_PERMISSION))
```
tcibinan: I think we should avoid managing permissions on data storage manager layer if possible. It seems like something that `DataStorageApiService` should be doing f.e. by using `PostFilter`.

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
```diff
@@ -778,16 +779,19 @@ public List<DataStorageTagSearchResult> searchDataStorageItemByTag(
...
 
             rootSearchResult.forEach((rootId, rootStorageTags) ->
                 rootStorageTags.forEach(dataStorageTag -> {
-                    final List<AbstractDataStorage> storages = storagesByRootId.get(rootId);
+                    final List<AbstractDataStorage> storages = storagesByRootId
+                            .getOrDefault(rootId, Collections.emptyList());
                     storages.stream().filter(
                         dataStorage -> dataStorageTag.getObject().getPath().startsWith(dataStorage.getPrefix())
-                    ).findFirst().ifPresent(dataStorage -> {
+                    ).max(Comparator.comparingInt(s -> s.getPrefix().length())).ifPresent(dataStorage -> {
```
tcibinan: I think that the chosen mechanism of selecting only a single data storage per root is very fragile. We should either select a closest data storage on per tag basic or return tags associated with all storages.

Let's look at the following example

```
# we have two storages
/storage1
/storage1/storage2

# we have two objects
/storage1/file1
/storage1/storage2/file2

# current implementation will associate all found objects with storage2 
# even if they are actually a part of storage1
/storage1/storage2/file1
/storage1/storage2/file2
```

tcibinan (changes requested):

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
tcibinan: Nah. What I've described is not entirely correct. The mechanism should work fine.

At the same time I would propose to keep only a single chain call per line otherwise it can be a little confusing :)

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
SilinPavel: I put it here because we need to filter available to read storage first, and that use that storage in search results. So If we move it to upper level (on api service) we may lose some of the results, because during the search will peek one that is not available.

api/src/main/java/com/epam/pipeline/manager/datastorage/DataStorageManager.java
SilinPavel: In fact this works in other way:
In your example we will have results as:

/storage1/file1
/storage1/storage2/file2 

because on step
```
.filter(dataStorage -> dataStorageTag.getObject().getPath().startsWith(dataStorage.getPrefix()))
```
we will peek `storage1` for `file1` and `storage2` for `file2`

tcibinan (approved):

mzueva (approved):

Issue #3003, opened 2023-01-11
Add new fields to storage file object in Data Catalog


SilinPavel:
**Background**
`Data Catalog` is a great tool to perform search on a big amount of indexed data stored in Cloud-Pipeline such as pipeline, datastorage or file inside a datastorage, etc.
To make this tool more powerful and user friendly it would be good to expand number of fields to index.
Currently `Data Catalog` already provides information on `path` for files stored in datastorages, however this path is relative to `datastorage path` and for some cases it can be not enough.

**Approach**
Let's in addition to `path` field also index fields `cloud_path` and `mount_path`:
 - `cloud_path` - path to file with information of a datastorage: `s3://storage-name/path/to/file` `nfs://lustre.storage.address@tcp:/mount-id/path/to/file`
 - `mount_path` - path to file with information of local mount point (how it will be mounted into a run): `/cloud-data/storage-name/path/to/file` `/custom-mount-point/path/to/file`

These fields will be sent on `/facet` and `/export` request, together with other fields like `id`, `name`, `path`, etc.

PR #3004, opened 2023-01-11
Expand number of fields of elasticsearch index for datastorage files

SilinPavel:
Related to #3003 
This PR adds couple of new fields for to document scheme for elasticsearch index for datastorage files.
Also it includes these fields to search result for `Data Catalog` and to export csv report

### DIFF
--- a/elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/converter/storage/StorageFileMapper.java
+++ b/elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/converter/storage/StorageFileMapper.java
@@ -118,7 +118,7 @@ public class StorageFileMapper {
     }
 
     private static String constructCloudPath(AbstractDataStorage dataStorage, DataStorageFile dataStorageFile) {
-        return String.join(dataStorage.getDelimiter(), dataStorage.getPath(), dataStorageFile.getPath());
+        return String.join(dataStorage.getDelimiter(), dataStorage.getPathMask(), dataStorageFile.getPath());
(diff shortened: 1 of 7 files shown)


elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/converter/storage/StorageFileMapper.java
```diff
@@ -65,6 +69,8 @@ public XContentBuilder fileToDocument(final DataStorageFile dataStorageFile,
                     .field("lastModified", dataStorageFile.getChanged())
                     .field("size", dataStorageFile.getSize())
                     .field("path", dataStorageFile.getPath())
+                    .field("cloud_path", constructCloudPath(dataStorage, dataStorageFile))
+                    .field("mount_path", constructMountPath(dataStorage, dataStorageFile))
```
tcibinan, line 73: As long as we add new fields to elastic documents we also add these fields to the corresponding index mapping. It helps avoiding dynamic index mapping. Could you please add these fields to `elasticsearch-agent/src/main/resources/templates/storage_file.json`?

tcibinan (changes requested):

elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/converter/storage/StorageFileMapper.java
SilinPavel: Nice catch, will do that

Issue #3015, opened 2023-01-17
Show notifications in GUI


mzueva:
**Background**
At the moment Cloud Pipeline platform sends notifications to users via email. For many cases it would be useful to show such notifications in the GUI as well.

**Approach**
- A new API methods shall be added to manage `UserNotification` entities (`userId`, `subject`, `text`, `dateCreated`, `(un)read flag`, `dateRead`). CRUD operation shall be implemented to create, delete, get list of notifications for a user.
- All email notifications shall be duplicated into this new object. I'd suggest to do this somewhere at `NotificationManager` level.
- Notifications shall be deleted by a scheduled service after a period, configured via SystemPreferences.

maryvictol:
@AleksandrGorodetskii 

**Bug**
 **Notifications table**:  when the user changes the notifications type from _New messages_ to _Read messages_ (or vice versa) the application tries to open the same table page number for _Read messages_  table that was opened in the _New messages_ table. If there are fewer pages in the _Read messages_  table, then an empty table is shown.

NShaforostov:
Docs were added via #2919 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/12_Manage_Settings/12.9._Change_email_notification.md#push-notifications).

PR #3023, opened 2023-01-20
Issue 3015: Show notifications in GUI

okolesn:
[Show notifications in GUI #3015](https://github.com/epam/cloud-pipeline/issues/3015)


api/src/main/java/com/epam/pipeline/acl/notification/UserNotificationApiService.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.List;
+
+@Service
+@RequiredArgsConstructor
+public class UserNotificationApiService {
+
+    private final UserNotificationManager notificationManager;
+
+    @PreAuthorize(AclExpressions.ADMIN_ONLY)
```
mzueva: All methods shall be available for admin and user himself, so that user may change or delete own notifications

api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
```diff
@@ -949,4 +965,31 @@ private NotificationSettings getNotificationSettings(final NotificationType type
         }
         return settings;
     }
+
+    @Transactional(propagation = Propagation.REQUIRED)
```
mzueva: As far ad I remember `@Transactional` doesn't work on private methods

api/src/main/java/com/epam/pipeline/controller/notification/UserNotificationController.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+
+@RestController
+@RequestMapping("/user-notification")
+@RequiredArgsConstructor
+public class UserNotificationController extends AbstractRestController {
+
+    private final UserNotificationApiService notificationApiService;
+
+    @PostMapping("/message")
```
mzueva, line 45: I'd add method `GET /message/my` to get messages for currently authorised used as well

api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
```diff
@@ -168,6 +172,7 @@ public void notifyLongRunningTask(PipelineRun run, Long duration, NotificationSe
 
         notificationMessage.setTemplateParameters(PipelineRunMapper.map(run, settings.getThreshold(), duration));
         monitoringNotificationDao.createMonitoringNotification(notificationMessage);
+        saveUserNotification(notificationMessage);
```
mzueva: At the moment each method calls dao class directly, but with new functionality we need to add a second method to each call. I'd suggest to add new private method `saveNotification` with call to saveUserNotification & dao and use it in all other methods

mzueva (changes requested):

mzueva (approved):

Issue #1615, opened 2020-11-27
Special type of endpoints to be able to configure and deploy independent DNS record for the tool


SilinPavel:
**Background**
At the moment, jobs will get endpoint URLs in the following format: https://< cloud-pipeline-edge >/pipeline-RUNID-PORTNUMBER
But this does not work well for the Web applications which the root of the site hard coded in the application’s codebase. E.g.:
Application is started and is available via https://< cloud-pipeline-edge >/pipeline-1111-8080/
But it may contains an html page, which has a link pointing to e.g. “/auth” page
This will make the web-browser load the https://< cloud-pipeline-edge >/auth instead of https://< cloud-pipeline-edge >/pipeline-1111-8080/auth
Such configuration results into 404 Not Found error
This is not possible to handle via the Proxies/Network routing techniques in a general way (it will have to create own nginx configuration to rewrite the javascript contents on the fly)
What is suggested is to have (optionally) the “Pipeline ID” (this part of the URL: /pipeline-RUNID-PORTNUMBER) as a part of the domain. E.g.: pipeline-RUNID-PORTNUMBER.< cloud-pipeline-edge >
This will make all such Web applications happy, as they will be served from the root of the site, as they expect

So it is proposed to consider as a development item for future is:

If the job is launched with some “special” option – Cloud-Pipeline requests creation of the new DNS entry in Route53
This DNS entry is then automatically used as a “friendly url” option for the job 

User gets an endpoint in a “domain-based” fashion (https://pipeline-1111-8080.< cloud-pipeline-edge >), instead of “suffix-based” (https://< cloud-pipeline-edge >/pipeline-1111-8080/)

SilinPavel:
Regarding AWS route53 permission, we will need the next list of permissions:
```
[
 {
      "Sid": "ChangeResourceRecordSets",
      "Effect": "Allow",
      "Action": "route53:ChangeResourceRecordSets",
      "Resource": "arn:aws:route53:::hostedzone/hosted zone ID"
  },
   {
      "Sid": "GetChange",
      "Effect": "Allow",
      "Action": "route53:GetChange",
      "Resource": "*"
  },
 {
      "Sid": "ListResourceRecordSets",
      "Effect": "Allow",
      "Action": "route53:ListResourceRecordSets",
      "Resource": "arn:aws:route53:::hostedzone/hosted zone ID"
  }
]
```

mzueva:
@sidoruka @SilinPavel 
[...]

PR #3024, opened 2023-01-23
Support regional custom dns records

tcibinan:
Relates #1615.

The pull request improves custom dns records support to allow multiple edge/region-specific dns custom records. 

The following additional optional fields are added to cloud region entities:
- _dnsHostedZoneId_ configures dns hosted zone id
- _dnsHostedZoneBase_ configures dns hosted zone base

Previously used system preferences are still used as fallback values:
- _instance.dns.hosted.zone.id_ configures fallback dns hosted zone id
- _instance.dns.hosted.zone.base_ configures fallback dns hosted zone base


api/src/main/java/com/epam/pipeline/manager/cloud/aws/Route53Helper.java
```diff
@@ -131,7 +131,7 @@ private InstanceDNSRecord.DNSRecordStatus getStatus(final String status) {
             case "INSYNC":
                 return InstanceDNSRecord.DNSRecordStatus.INSYNC;
             default:
-                return InstanceDNSRecord.DNSRecordStatus.NO_OP;
```
SilinPavel, line 134: I think that this place should be also changed just for consistency, see: [line 94](https://github.com/epam/cloud-pipeline/pull/3024/files#diff-a2772bae2f3ffc210ed129e3c24c68d6a0333b34836c435abb732fbb73105f9bL94)
from `INSYNC` to `NOOP`, because actual work wasn't done. What do you think? As I can see we don't check result of this method but still it can be useful some day.

api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSInstanceService.java
```diff
@@ -289,29 +285,51 @@ public CloudInstanceState getInstanceState(final AwsRegion region, final String
...
-                                                          final InstanceDNSRecord dnsRecord) {
-        if (dnsRecord.getDnsRecord().contains(
-                preferenceManager.getPreference(SystemPreferences.INSTANCE_DNS_HOSTED_ZONE_BASE))) {
-            return route53Helper
-                    .createDNSRecord(awsRegion, preferenceManager.getPreference(
-                            SystemPreferences.INSTANCE_DNS_HOSTED_ZONE_ID), dnsRecord);
-        } else {
-            return NO_OP_INSTANCE_DNS_RECORD;
-        }
+    public InstanceDNSRecord getOrCreateInstanceDNSRecord(final AwsRegion region,
```
SilinPavel, line 289: Overall these changes are great improvements in working with `DNS` records. 
The only one thing that I'm worry about is format of `InstanceDNSRecord`, right now the contract for this functionality:
 - When you create a record you should use `relative` format
 - When you delete a record you should use `full` format

For me it seems little bit inconsistent, at least without explicitly specification what type of record we use.
What do you think if we will introduce new field for `InstanceDNSRecord` - `format` (`RELATIVE`/`ABSOLUTE`) and will work with respect to this value. Such approach will be easy to implement for current functionality (we can set `RELATIVE` as default) and also will look more consistent and flexible for further use if any.

SilinPavel (changes requested):

api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSInstanceService.java
tcibinan: Done

api/src/main/java/com/epam/pipeline/manager/cloud/aws/Route53Helper.java
tcibinan: Done

SilinPavel (approved):

mzueva (approved):

Issue #3020, opened 2023-01-19
Archived S3 files shall be excluded from pipe storage & pipe mount operations


mzueva:
**Background**
With implementation of #2721 S3 files can be transferred to the different archive storage classed, such as Deep Archive. Any operation of such files is possible only after process of file restoring, that's why we need to exclude such files from listing and any operations available from `pipe storage` and `pipe mount` commands.
This behaviour shall be used by default, but a new flag shall be added `--show-archive` to both commands which enables archive files browsing.

NShaforostov:
Docs were added via #2919 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/08_Manage_Data_Storage/8.10._Storage_lifecycle.md#listing-via-cli).

PR #3039, opened 2023-01-31
Archived S3 files shall be excluded from pipe storage & pipe mount operations

ekazachkova:
The current PR provides additional implementation for issue #3020 

 - a new command line option `--show-archived (Default: false)` added to `pipe storage mount` command. If this flag enabled archived files (GLACIER, DEEP_ARCHIVE storage classes) shall be shown. Otherwise, not restored archived files shall be filtered out. 
 - a new command line option `--storage-class-exclude` added to pipe fuse. This option specifies a list of storage classes that shall be excluded from listing.
 - added storage class info for get attribute operation

### DIFF
--- a/pipe-cli/mount/pipe-fuse.py
+++ b/pipe-cli/mount/pipe-fuse.py
@@ -67,7 +67,7 @@ from pipefuse.trunc import CopyOnDownTruncateFileSystemClient, \
 from pipefuse.webdav import WebDavClient, ResilientWebDavFileSystemClient
 from pipefuse.xattr import ExtendedAttributesCache, ThreadSafeExtendedAttributesCache, \
     ExtendedAttributesCachingFileSystemClient, RestrictingExtendedAttributesFS
-from pipefuse.archived import ArchivedFilesFilterFileSystemClient
+from pipefuse.archived import ArchivedFilesFilterFileSystemClient, ArchivedAttributesFileSystemClient
 from pipefuse.storageclassfilter import StorageClassFilterFileSystemClient
 
 _allowed_logging_level_names = ['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG', 'NOTSET']
@@ -124,6 +124,7 @@ def start(mountpoint, webdav, bucket,
             client = S3StorageLowLevelClient(bucket_name, pipe=pipe, chunk_size=chunk_size, storage_path=bucket)
             if not show_archived:
                 client = ArchivedFilesFilterFileSystemClient(client, pipe=pipe, bucket=client.bucket_object)
+            client = ArchivedAttributesFileSystemClient(client, pipe=pipe, bucket=client.bucket_object)
         elif bucket_type == CloudType.GS:
             client = GoogleStorageLowLevelFileSystemClient(bucket_name, pipe=pipe, chunk_size=chunk_size,
                                                            storage_path=bucket)
--- a/pipe-cli/mount/pipe-fuse.py
+++ b/pipe-cli/mount/pipe-fuse.py
@@ -67,6 +67,7 @@ from pipefuse.trunc import CopyOnDownTruncateFileSystemClient, \
 from pipefuse.webdav import WebDavClient, ResilientWebDavFileSystemClient
 from pipefuse.xattr import ExtendedAttributesCache, ThreadSafeExtendedAttributesCache, \
     ExtendedAttributesCachingFileSystemClient, RestrictingExtendedAttributesFS
+from pipefuse.archived import ArchivedFilesFilterFileSystemClient
 
 _allowed_logging_level_names = ['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG', 'NOTSET']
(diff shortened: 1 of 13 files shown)


pipe-cli/pipe.py
```diff
@@ -1340,9 +1340,10 @@ def storage_delete_object_tags(path, tags, version):
 @click.option('-m', '--mode', required=False, help='Default file permissions',  default=700, type=int)
 @click.option('-w', '--timeout', required=False, help='Waiting time in ms to check whether mount was successful',
               default=1000, type=int)
+@click.option('-g', '--show-archive', is_flag=True, help='Show archived files.')
```
tcibinan, line 1343: In `pipe storage mount` we have a special mechanism for pushing `pipe fuse` additional arguments. For example in order to push `show-archive` and `write-buffer-size` arguments we can use the following command:

```
pipe storage mount --custom-options show-archive=True,write-buffer-size=100 ...
```

So I would suggest getting rid of explicit argument `--show-archive` from `pipe storage mount`.

pipe-cli/mount/pipe-fuse.py
```diff
@@ -342,6 +346,7 @@ def parse_mount_options(options_string):
     parser.add_argument("-th", "--threads", action='store_true', help="Enables multithreading.")
     parser.add_argument("-d", "--monitoring-delay", type=int, required=False, default=600,
                         help="Delay between path lock monitoring cycles.")
+    parser.add_argument("--show-archived", action='store_true', help="Show archived files.")
     args = parser.parse_args()
```
tcibinan, line 356: I would recommend adding additional `--storage-class-exclude=GLACIER,DEEP_ARCHIVE` argument which will allow skipping certain storages classes. It can be used along with `--show-archived=False` to completely ignore glacier archived/unarchived files.

pipe-cli/mount/pipefuse/archived.py
```diff
@@ -0,0 +1,87 @@
...
+        storage_lifecycle = None
+        result = []
+        for item in items:
+            if not item.is_dir and item.storage_class != 'STANDARD':
+                path = self._normalize_path(path)
+                if storage_lifecycle is None:
+                    storage_lifecycle = self._get_storage_lifecycle(path)
+                file_path = path + item.name
+                if not self._file_restored(file_path, storage_lifecycle):
+                    continue
```
tcibinan: It seems that `storage_lifecycle` dictionary can be used more efficiently considering the fact that it can only contain either exact file path or some parent folder path.

tcibinan (changes requested):

pipe-cli/pipe.py
tcibinan: Not applicable anymore. Decided to keep the option in pipe storage mount after discussion.

pipe-cli/mount/pipefuse/archived.py
```diff
@@ -0,0 +1,137 @@
...
+            response = self._pipe.get_storage_lifecycle(self._bucket, path)
+            items = []
+            if not response:
+                return set()
+            for item in response:
+                if item.is_restored():
+                    items.append(item.path)
+            return set(items)
+        except _ALL_ERRORS as e:
+            logging.info(e)
```
tcibinan: Let's use the following command instead. It automatically adds trace to the logs. Otherwise the error trace will be lost.

```python
logging.exception('Storage lifecycle retrieving has failed')
```

Same goes for two more similar cases in this class.

pipe-cli/mount/pipefuse/pipefs.py
```diff
@@ -364,7 +366,11 @@ def _wrapped_attr(*args, **kwargs):
                 raise FuseOSError(errno.ENODATA)
             except InvalidOperationException:
                 raise FuseOSError(errno.EINVAL)
-            except Exception:
+            except Exception as e:
+                err_msg = str(e)
+                if isinstance(e, ClientError) \
+                        and err_msg and 'InvalidObjectState' in err_msg and 'storage class' in err_msg:
+                    raise FuseOSError(errno.EACCES)
```
tcibinan: Let's add exception logging here as well.

pipe-cli/mount/pipe-fuse.py
```diff
@@ -342,6 +350,9 @@ def parse_mount_options(options_string):
     parser.add_argument("-th", "--threads", action='store_true', help="Enables multithreading.")
     parser.add_argument("-d", "--monitoring-delay", type=int, required=False, default=600,
                         help="Delay between path lock monitoring cycles.")
+    parser.add_argument("--show-archived", action='store_true', help="Show archived files.")
+    parser.add_argument("--storage-class-exclude", type=str, required=False,
+                        help="Comma-separated values with storage classes shall be excluded.")
```
tcibinan: Let's use the same mechanism as we have for `--xattrs-include-prefix/--xattrs-exclude-prefix` where we multiple arguments rather than comma-separated value.

tcibinan (changes requested):

tcibinan (approved):

mzueva (approved):

Issue #3028, opened 2023-01-25
Storage volume shall be calculated for all tiers


mzueva:
**Background**
With implementation of #2721 S3 objects may reside in a various storage classes. At the moment when storage class is estimated all objects are included into the value, but it makes sense to have volume for each tier separately.

**Approach**
- API method to get storage usage shall return storage volume for the following tiers:
  - hot (currently supported)
  - old versions
  - S3 Glacier tiers
  - total (sum of all above)
- Implementation of `pipe du` command shall be checked as well

NShaforostov:
Docs were added via #2919 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/08_Manage_Data_Storage/8.10._Storage_lifecycle.md#view-archive-size).

PR #3051, opened 2023-02-07
Issue #3028 backward compatibility

SilinPavel:
This PR fix problem with backward compatibility

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/search/SearchResultConverter.java
+++ b/api/src/main/java/com/epam/pipeline/manager/search/SearchResultConverter.java
@@ -91,32 +91,25 @@ public class SearchResultConverter {
         final MultiSearchResponse.Item[] responses = tryExtractAllResponses(searchResponse, searchRequest);
         final Map<String, StorageUsage.StorageUsageStats> statsByTiers = buildStorageUsageStatsByStorageTier(responses);
 
+        final Optional<StorageUsage.StorageUsageStats> standardUsageOp =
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/search/SearchResultConverter.java
```diff
@@ -94,6 +96,26 @@ public StorageUsage buildStorageUsageResponse(final MultiSearchRequest searchReq
                 .name(dataStorage.getName())
                 .type(dataStorage.getType())
                 .path(path)
+                .size(
+                    Optional.ofNullable(statsByTiers.get(STANDARD_TIER))
```
mzueva: Maybe we can call `Optional.ofNullable` once?

core/src/main/java/com/epam/pipeline/entity/datastorage/StorageUsage.java
```diff
@@ -31,6 +31,11 @@ public class StorageUsage {
     private String name;
     private DataStorageType type;
     private String path;
+    private Long size;
+    private Long effectiveSize;
+    private Long oldVersionsSize;
+    private Long oldVersionsEffectiveSize;
+    private Long count;
```
mzueva, line 38: I think we should either add `effectiveCount` field, or remove `oldVersionsSize`, `oldVersionsEffectiveSize`. Other way it is a little bit inconsistent

mzueva (changes requested):

mzueva (approved):

Issue #3056, opened 2023-02-08
Review archive operations permissions


mzueva:
**Background**
Related to #2721 
At the moment the following operations are implemented:

| N  | Action | ROLE_ADMIN | OWNER | ROLE_STORAGE_ARCHIVE_MANAGER with read-write access | ROLE_STORAGE_ARCHIVE_READER with read access | ROLE_USER |
| -- | -------|---------------|---------|----------------------------------------|-------------------------------------|--------------|
| 1   | Lifecycle rules management(CRUD) | &check;    | &check;    | &check;    |  |  |
| 2   | Lifecycle rules reading | &check;    | &check;    | &check;    | &check;    |  |
| 3   | Files restoring | &check;    | &check;    | &check;    |  |  |
| 4   | Browsing archived files | &check;    | &check;    | &check;    | &check;    |  |

**Approach**
Review the permissions for:
- API server methods
- GUI
- pipe storage commands (`show-archive` flag & `restore`)

mzueva:
@ekazachkova 
Regarding `pipe du` permissions:
- Column `Archive versions` and archive tier columns shall be shown only for users with `ROLE_STORAGE_ARCHIVE_READER`, both for `brief` and `full` output modes
- Flag `generation` with value `old` shall be allowed only for users with ADMIN or OWNER access to the bucket. If user specifies `all` value without appropriate permissions - command shall still show `current` versions only

PR #3090, opened 2023-02-21
Issue #3056: Review archive operations permissions

ekazachkova:
The current PR provides implementation for issue #3056 

- `pipe storage mount/ls`: if `--show-archive` option specified an error shall be occurred in case if user has not appropriate permissions
- archives for `pipe storage du`: if user has not read archive permissions archive sizes shall not be shown
- versions for `pipe storage du`: if user is not admin or owner `--generation=old` shall not be available, `--generation=all` shall works as `current`.
- archive permissions management added to datastorage lifecycle methods and `GET /datastorage/path/usage` method

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
@@ -820,21 +820,6 @@ public class GrantPermissionManager {
         return cloudProfileCredentialsManagerProvider.hasAssignedUserOrRole(profileId, currentUser.getId(),
                 currentUser.getRoles());
     }
-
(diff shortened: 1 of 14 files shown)


api/src/main/java/com/epam/pipeline/manager/security/GrantPermissionManager.java
```diff
@@ -821,6 +821,20 @@ public boolean hasCloudProfilePermissions(final Long profileId) {
                 currentUser.getRoles());
     }
 
+    public boolean storageArchiveReadPermissions(final AbstractDataStorage storage) {
```
mzueva: Could you move these methods to `StoragePermissionManager` class?

mzueva (changes requested):

mzueva (approved):

PR #3093, opened 2023-02-21
Issue #3028 Set region code to elasticsearch doc for nfs files

SilinPavel:
Related to #3028 
This PR fixes problem when we didn't set `storage_region` field for NFS files while indexing it. 

Such logic leads to wrong price calculation in `billing-agent`, because in such cases we use `defaultCost` as a cost for storing such files, and this cost is calculated as max cost from price list.

### DIFF
--- a/elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/NFSSynchronizer.java
+++ b/elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/NFSSynchronizer.java
@@ -25,6 +25,7 @@ import com.epam.pipeline.entity.datastorage.DataStorageFile;
 import com.epam.pipeline.entity.datastorage.DataStorageType;
 import com.epam.pipeline.entity.datastorage.DataStorageWithShareMount;
 import com.epam.pipeline.entity.datastorage.NFSDataStorage;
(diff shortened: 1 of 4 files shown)


elasticsearch-agent/src/main/java/com/epam/pipeline/elasticsearchagent/service/impl/NFSSynchronizer.java
```diff
@@ -107,15 +109,19 @@ public void synchronize(final LocalDateTime lastSyncTime, final LocalDateTime sy
...
                 .forEach(this::createIndexAndDocuments);
     }
 
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
-    void createIndexAndDocuments(final AbstractDataStorage dataStorage) {
-        EntityPermissionVO entityPermission = cloudPipelineAPIClient
+    void createIndexAndDocuments(final DataStorageWithShareMount storageWithShareMount) {
+        final AbstractDataStorage dataStorage = storageWithShareMount.getStorage();
+        log.debug("Starting to  process storage: {}, id: {}.", dataStorage.getName(), dataStorage.getId());
+        final String regionCode = Optional.ofNullable(storageWithShareMount.getShareMount())
```
mzueva: I think we should preload all regions rather than call region api for each storage

mzueva (approved):

mzueva (approved):

Issue #3075, opened 2023-02-15
Audit pipe mount operations


mzueva:
**Background**
Related to #3059 
Pipe mount read, write delete operations shall be monitored and logged to the API.

tcibinan:
As long as it is inefficient to collect singular read/write operations pipe mount can rather collect only first read/write operations for each file descriptor. Further only unique file read/write operations within certain time period f.e. one minute can be registered with SystemLog API in a bulk asynchronous manner.

maryvictol:
Test cases were created by #3146 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/System_Management/3059-3075_audit_pipe_mount_operation).

PR #3119, opened 2023-03-01
Support data access audit in pipe cli

tcibinan:
Relates to #3075 and depends on #3105.

The pull request brings support for data access audit to pipe cli. From now on all storage cp/mv/rm/restore operations are logged via Cloud Pipeline System Log API.

### DIFF
--- a/pipe-cli/src/utilities/storage/azure.py
+++ b/pipe-cli/src/utilities/storage/azure.py
@@ -23,6 +23,7 @@ import time
 from datetime import timedelta, datetime
 
 from src.model.datastorage_usage_model import StorageUsage
+from src.utilities.audit import StorageDataAccessEntry, DataAccessType
 from src.utilities.encoding_utilities import to_string
 from src.utilities.storage.storage_usage import StorageUsageAccumulator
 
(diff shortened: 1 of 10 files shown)


pipe-cli/src/utilities/storage/azure.py
```diff
@@ -281,6 +287,7 @@ def transfer(self, source_wrapper, destination_wrapper, path=None,
 
         self.create_local_folder(destination_key, lock)
         progress_callback = AzureProgressPercentage.callback(source_key, size, quiet, lock)
+        self.audit.put(StorageDataAccessEntry(source_wrapper.bucket, source_key, DataAccessType.READ))
         self.service.get_blob_to_path(source_wrapper.bucket.path, source_key, to_string(destination_key),
                                       progress_callback=progress_callback)
         if clean:
```
ekazachkova, line 293: Do we need to log DELETE action here?

tcibinan: Added.

ekazachkova (approved):

PR #3120, opened 2023-03-01
Reuse data access audit module between pipe cli and pipe mount

tcibinan:
Relates to #3075 and depends on #3119.

The pull request reuses pipe cli's audit module in pipe mount. Additionally it ensures system log event ids uniqueness.


pipe-cli/src/utilities/storage/mount.py
```diff
@@ -134,6 +137,10 @@ def _get_mount_cmd(self, config, mountpoint, options, additional_arguments, mode
             mount_cmd.extend(['-o', options])
         return mount_cmd
 
+    def get_python_path(self):
+        script_folder = dirname(Config.get_base_source_dir())
```
ekazachkova: It looks like we can simplify to one line here: 
```
return dirname(Config.get_base_source_dir())
```

tcibinan: Simplified.

ekazachkova (approved):

Issue #3131, opened 2023-03-09
Running System Jobs


SilinPavel:
**Background**
Sometimes it's desirable to get some statistics or system information about current system state (collect information about all storages that have size > X gb, list all unattached ebs volumes, set some s3 bucket policy to all storages, etc).
It is also common for such purpose to write some sort of automation script (with help of pipe CLI, Cloud Pipeline API, or AWS CLI), after a while the number of such scripts can grow very quickly and it seems right to have some entry point for such scripts in a system.

So, it would be great to have solution to run these scripts inside Cloud-Pipeline infrastructure, to reduce number of preparation steps to be done to get desire output.   

**Approach**
Let's try to reuse existing Cloud-Pipeline ecosystem features to run such jobs, observe results and store some outputs (logs, etc).
It seems that Pipeline functionality should cover all necessary aspects, with minimal changes required.

Here is a list of requirements for the environment to run such scripts/jobs:
- Access to Cloud Pipeline API
- AWS Access and Credentials
- Ability to use Kube API/kubectl
- Internet Access
- Access to Cloud-Pipeline FS (lustre)
- PipeCLI

And here is a functional mapping of this requirements to existing Cloud-Pipeline features:

|Functional requirement        | Cloud Pipeline Capability | Comment        |
|------------------------------|---------------------------|----------------|
| Access to Cloud Pipeline API |              -            | Out of the box |
| Internet Access	       |              -            | Out of the box |
|Have an output as file on specific bucket |	Pipeline output parameter |	Out of the box |
| Pipe CLI                     | Pipe CLI installed into pre-built  tool image| - |
|Cloud Access and Credentials    | Generation Temp Credentials / node profile | We can either generate credentials as always we do on run startup process, or relay on creds that will be assigned on a node |
| Launching on particular node   | Needs to be implemented | We need ability to specify some of `system` node as a target node for a pipeline pod |
| Kube API / kubectl	       | Needs to be implemented   | Kube Service Account should be created for this purpose and assigned to the pod during "system" job creation phase |
| Pipeline FS (lustre)         | Pipeline Parameters + customization of pipeline code  | Use Pipeline parameter to store information about lustre fs, on pipeline startup, before launching a script, mount this FS |

So, seems that Pipeline should be a good match for this solution.

As result we can pick a couple of things to develop:

 -  Ability to run such jobs on `specific` node (f.e. for prompt response and saving compute resources)
[...]

PR #3140, opened 2023-03-16
Issue 3131 System Jobs capability

SilinPavel:
Related to #3131 

This PR provides possibility to specify `kube service account` to be used to launch a run.

It is also introduce new object `RunAssignPolicy` with this object user can more accurate define how a run should be scheduled (will be useful for `System Jobs` since we want to schedule such jobs to non-worker nodes).

NOTE: This functionality will be available only for `Admins` (security checks are made in `PipelineLaunchAspect`)


SilinPavel:
I'm closing this one since #3148 was merged, and it includes all changes from this PR.

PR #3148, opened 2023-03-21
Issue 3131 System Jobs infra and deployment

SilinPavel:
This PR is related to #3131 depends on #3140 
Provides additional infrastructure configuration to register and setup System Jobs pipeline during `pipectl` deployment process.


api/src/main/java/com/epam/pipeline/manager/execution/PipelineLauncher.java
```diff
@@ -103,40 +105,42 @@ public class PipelineLauncher {
...
+                         final boolean useLaunch, final String pipelineId, final String clusterId) {
+        return launch(run, configuration, endpoints, useLaunch, pipelineId,
+                clusterId, getImagePullPolicy(configuration));
     }
 
-    public String launch(PipelineRun run, PipelineConfiguration configuration,
-                         List<String> endpoints, String nodeIdLabel, boolean useLaunch,
-                         String pipelineId, String clusterId, ImagePullPolicy imagePullPolicy) {
-        GitCredentials gitCredentials = configuration.getGitCredentials();
+    @PipelineLaunchCheck
```
mzueva: I'm not sure if we really need an annotation and aspect for validation, as in fact we have one method with full list of arguments and we can just add the check to its code?

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
```diff
@@ -317,6 +320,7 @@ public PipelineRun runPod(PipelineStart runVO) {
      * @return
      */
     @ToolSecurityPolicyCheck
+    @Transactional(propagation = Propagation.REQUIRED)
```
mzueva: I don't think we need to add `@Transactional` here, moreover I'd say then we need to remove it from other methods, but let's discuss it later

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunManager.java
SilinPavel: I believe that all public methods that interact with database and mutate data should be marked as `transactional` or maybe I'm missing something?

api/src/main/java/com/epam/pipeline/manager/execution/PipelineLauncher.java
SilinPavel: Indeed, you are right.

mzueva (approved):

PR #3163, opened 2023-03-29
issue #3131 fix for run worker nodes

SilinPavel:
Now we are copy `runConfiguration` if we need it to be used as configuration for child runs.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineConfigurationManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineConfigurationManager.java
@@ -308,8 +308,8 @@ public class PipelineConfigurationManager {
         configuration.buildEnvVariables();
     }
 
-    public PipelineConfiguration copyAsMasterConfiguration(final PipelineConfiguration configuration,
-                                                           final boolean isNFS) {
+    public PipelineConfiguration generateMasterConfiguration(final PipelineConfiguration configuration,
+                                                             final boolean isNFS) {
         final PipelineConfiguration copiedConfiguration = configuration.clone();
         updateMasterConfiguration(configuration, isNFS);
         return copiedConfiguration;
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineConfigurationManager.java
```diff
@@ -308,11 +308,20 @@ public void updateMasterConfiguration(PipelineConfiguration configuration, boole
         configuration.buildEnvVariables();
     }
 
-    public void updateWorkerConfiguration(String parentId, PipelineStart runVO,
-            PipelineConfiguration configuration, boolean isNFS, boolean clearParams) {
-        configuration.setEraseRunEndpoints(hasBooleanParameter(configuration, ERASE_WORKER_ENDPOINTS));
-        final Map<String, PipeConfValueVO> configParameters = MapUtils.isEmpty(configuration.getParameters()) ?
-                new HashMap<>() : configuration.getParameters();
+    public PipelineConfiguration copyAsMasterConfiguration(final PipelineConfiguration configuration,
```
mzueva: Maybe we should name the methods similar? `generateMasterConfiguration` and `generateWorkerConfiguration`

api/src/main/java/com/epam/pipeline/manager/pipeline/runner/CloudPlatformRunner.java
```diff
@@ -164,30 +164,41 @@ private List<PipelineRun> runConfigurationEntry(RunConfigurationEntry entry,
...
     }
 
+    private PipelineConfiguration buildRunConfiguration(final RunConfigurationEntry entry,
+                                                        final PipelineConfiguration configuration,
+                                                        final String clusterId, final boolean startNFS,
+                                                        final PipelineStart startVO) {
+        final PipelineConfiguration runConfiguration;
+        if (!StringUtils.hasText(clusterId)) {
+            log.debug("Launching master entry {}", entry.getName());
+            runConfiguration = pipelineConfigurationManager.copyAsMasterConfiguration(configuration, startNFS);
```
mzueva: We can return results immediately instead of assigning the variable

mzueva (changes requested):

Issue #3159, opened 2023-03-28
Report SGE worker node status


mzueva:
**Background**
It would be helpful to report SGE workers status using run label feature.

**Approach**
SGE Autoscaler shall assign label `SGE_IN_USE` to the child nodes which are currently processing some of the SGE jobs. Label shall be removed for nodes which are idle at the moment.

tcibinan:
Cherry-picked to `stage/0.16/uat` via 742fd7f5ef0b2c1d261b735eaa47ac699512fdef.

PR #3179, opened 2023-04-04
Report SGE worker node status

ekazachkova:
The current PR provides implementation for issue #3159 

 - a new tag `SGE_IN_USE` is added for all active runs
 - `sge_setup_master` launches `autoscale_sge.py` for all cases (removed `CP_CAP_AUTOSCALE` flag restriction)

### DIFF
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -2207,8 +2207,7 @@ def load_default_hosts(default_hostsfile):
         return []
 
 
-def init_static_hosts(default_hostsfile, static_host_storage, clock, tagging_active_timeout, static_hosts_enabled,
-                      common_utils, parent_run_id):
+def init_static_hosts(default_hostsfile, static_host_storage, clock, tagging_active_timeout, static_hosts_enabled):
     try:
         if static_host_storage.load_hosts():
             Logger.info('Static hosts already initialized.')
@@ -2219,7 +2218,7 @@ def init_static_hosts(default_hostsfile, static_host_storage, clock, tagging_act
             for host in hosts:
                 static_host_storage.add_host(host)
         else:
-            master_host = common_utils.retrieve_pod_name(parent_run_id)
+            master_host = os.getenv('HOSTNAME')
             static_host_storage.add_host(master_host)
             hosts = [master_host]
         # to prevent false positive run tagging let's add outdated date to hosts:
@@ -2380,9 +2379,8 @@ def main():
                                                                           '.static.%s.storage' % queue_name),
                                                 clock=clock)
     init_static_hosts(default_hostsfile=default_hostfile, static_host_storage=static_host_storage, clock=clock,
-                      tagging_active_timeout=tagging_active_timeout, common_utils=common_utils,
-                      static_hosts_enabled=queue_static and static_hosts_number,
-                      parent_run_id=master_run_id)
+                      tagging_active_timeout=tagging_active_timeout,
+                      static_hosts_enabled=queue_static and static_hosts_number)
     if not autoscale_enabled:
         Logger.info('Using non autoscaling mode...')
         autoscaling_hosts_number = 0
--- a/workflows/pipe-common/scripts/autoscale_sge.py
+++ b/workflows/pipe-common/scripts/autoscale_sge.py
@@ -790,8 +790,8 @@ class ScaleCommonUtils:
 class GridEngineScaleUpOrchestrator:
     _POLL_DELAY = 10
 
-    def __init__(self, scale_up_handler, grid_engine, host_storage, instance_selector, worker_recorder,
-                 batch_size, polling_delay=_POLL_DELAY, clock=Clock()):
+    def __init__(self, scale_up_handler, grid_engine, host_storage, static_host_storage, worker_tags_handler,
+                 instance_selector, worker_recorder, batch_size, polling_delay=_POLL_DELAY, clock=Clock()):
         """
         Grid engine scale up orchestrator.
 
@@ -802,6 +802,8 @@ class GridEngineScaleUpOrchestrator:
         :param scale_up_handler: Scaling up handler.
         :param grid_engine: Grid engine client.
         :param host_storage: Additional hosts storage.
+        :param static_host_storage: Static workers host storage
+        :param worker_tags_handler: Additional workers tags handler.
         :param instance_selector: Additional instances selector.
         :param worker_recorder: Additional hosts recorder.
         :param batch_size: Scaling up batch size.
@@ -810,6 +812,8 @@ class GridEngineScaleUpOrchestrator:
         self.scale_up_handler = scale_up_handler
         self.grid_engine = grid_engine
         self.host_storage = host_storage
+        self.static_host_storage = static_host_storage
+        self.worker_tags_handler = worker_tags_handler
         self.instance_selector = instance_selector
         self.worker_recorder = worker_recorder
         self.batch_size = batch_size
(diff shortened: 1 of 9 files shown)


workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -2061,6 +2195,25 @@ def fetch_instance_launch_params(api, master_run_id, queue, hostlist):
...
+    for nested_run in nested_runs:
+        run_id = nested_run.get('runId')
+        host = common_utils.retrieve_pod_name(run_id)
+        static_host_storage.add_host(host)
+        hosts.append(host)
+    # to prevent false positive run tagging let's add outdated date to hosts:
+    timeout = run_activation_timeout * 2
+    timestamp = clock.now() - timedelta(seconds=timeout)
+    static_host_storage.update_hosts_activity(hosts, timestamp)
+    Logger.info('Static hosts have been initialized.')
```
tcibinan: Grid Engine autoscaling script restarts multiple times throughout run's lifetime.

In case run restarts and _run/parents/runid_ method is called then it will return all running and stopped both static and autoscaling child runs. In order to manage only static child runs the runs shall be filtered by the following parameter _cluster_role_type != additional_.

Moreover we should be very careful with calling _common_utils.retrieve_pod_name_ here because it makes an API request every time. For big clusters of hundreds of child runs this may be dangerous.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -2201,6 +2355,18 @@ def main():
     scale_down_timeout = int(api.retrieve_preference('ge.autoscaling.scale.down.timeout', default_value=30))
     scale_up_polling_timeout = int(api.retrieve_preference('ge.autoscaling.scale.up.polling.timeout',
                                                            default_value=900))
+    common_utils = ScaleCommonUtils(api=api)
+    static_host_storage = FileSystemHostStorage(cmd_executor=cmd_executor,
+                                                storage_file=os.path.join(work_dir,
+                                                                          '.static.%s.storage' % queue_name),
+                                                clock=clock)
+    run_activation_timeout = 30
+    if static_hosts_number:
```
tcibinan: Grid engine autoscaler manages static hosts only if the current queue is static. If queue is static then static hosts are part of this queue. Therefore, we should you the following if here:

```suggestion
    if queue_static and static_hosts_number:
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -2201,6 +2355,18 @@ def main():
     scale_down_timeout = int(api.retrieve_preference('ge.autoscaling.scale.down.timeout', default_value=30))
     scale_up_polling_timeout = int(api.retrieve_preference('ge.autoscaling.scale.up.polling.timeout',
                                                            default_value=900))
+    common_utils = ScaleCommonUtils(api=api)
+    static_host_storage = FileSystemHostStorage(cmd_executor=cmd_executor,
+                                                storage_file=os.path.join(work_dir,
+                                                                          '.static.%s.storage' % queue_name),
+                                                clock=clock)
+    run_activation_timeout = 30
```
tcibinan: Let's call it _tagging_active_timeout_ and use environment variable _CP_CAP_AUTOSCALE_ACTIVE_TIMEOUT_ to get the corresponding value with _30_ as a default value.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -2201,6 +2355,18 @@ def main():
...
+    if static_hosts_number:
+        init_static_hosts(api=api, parent_run_id=master_run_id, common_utils=common_utils,
+                          static_host_storage=static_host_storage, clock=clock,
+                          run_activation_timeout=run_activation_timeout)
+    if not autoscale_enabled:
+        autoscaling_hosts_number = 0
```
tcibinan: Let's add a small log message just to be clear that this is a special autoscaler launch.

```suggestion
    if not autoscale_enabled:
        Logger.info('Using non autoscaling mode...')
        autoscaling_hosts_number = 0
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1300,10 +1311,118 @@ def _validate_existence(self, host, hosts_dict):
...
+    def _add_worker_tag(self, run_id):
+        run = self.api.load_run(run_id)
+        tags = run.get('tags') or {}
+        tags.update({self._WORKER_TAG: 'true'})
+        self.api.update_pipeline_run_tags(run_id, tags)
+
+    def _remove_worker_tag(self, run_id):
+        run = self.api.load_run(run_id)
+        tags = run.get('tags') or {}
+        if tags.__contains__(self._WORKER_TAG):
```
tcibinan: Let's use more pythonish syntax:

```suggestion
        if self._WORKER_TAG in tags:
```

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1300,10 +1311,118 @@ def _validate_existence(self, host, hosts_dict):
...
+            if last_monitored_timestamps and last_monitored_timestamps.last_tag_timestamp:
+                Logger.info("Removing tag from run.")
+                self._untag_run(current_host, timestamp, last_monitored_timestamps)
+                Logger.info("Removed tag from run for host %s." % current_host)
+
+    def _process_outdated_hosts(self, monitored_hosts, current_hosts):
+        for monitored_host in monitored_hosts:
+            if monitored_host not in current_hosts:
+                Logger.info("Host %s scaled down. Removing tags from run." % monitored_host)
+                self._untag_run(monitored_host)
```
tcibinan: Though I do appreciate having lots of logs, at the moment I would suggest to reduce the amount of logs to only two records (add/remove tag). In the current logging setup we don't have a proper usage of logging levels (debug in specific). This log is extremely important and we shall keep it brief and human readable even for clusters even with hundreds of workers.

In the future I want to use _pipeline.log.logger.CloudPipelineLogger_ which does better job in logging level management. Once it is used, we can add more debug logs. But for now let's keep it brief.

```
...
Logger.info('Marking active host %s...' % current_host)
...
Logger.info('Marking inactive host %s...' % current_host)
```

tcibinan (changes requested):

workflows/pipe-common/scripts/autoscale_sge.py
tcibinan: Seems to be resolved with the usage of filesystem host storages and the comment below.

workflows/pipe-common/scripts/autoscale_sge.py
```diff
@@ -1300,10 +1317,115 @@ def _validate_existence(self, host, hosts_dict):
...
+            current_hosts = self.host_storage.load_hosts()
+            hosts_activity = self.host_storage.get_hosts_activity(current_hosts)
+            hosts_activity.update(self.static_host_storage.get_hosts_activity(self.static_hosts))
+            monitored_hosts = list(self.last_monitored_hosts.keys())
+            current_hosts += self.static_hosts
+            self._process_current_hosts(current_hosts, hosts_activity)
+            self._process_outdated_hosts(monitored_hosts, current_hosts)
+            Logger.info('Finish tags processing step at %s.' % self.clock.now())
+        except Exception as e:
+            Logger.warn('Tags processing has failed: %s' % str(e))
```
tcibinan, line 1361: Let's log error stacktraces as well.

```suggestion
        except Exception as e:
            Logger.warn('Tags processing has failed: %s' % str(e))
            Logger.warn(traceback.format_exc())
```

tcibinan (changes requested):

tcibinan (approved):

mzueva (approved):

PR #3180, opened 2023-04-04
GUI disable OK button for roles and groups

AlfiyaRF:
This PR disables 'OK' button for groups and roles dialog if it isn't modified


rodichenko:
Merged with b5a6f36957c021eb6271add361389a18c1a0373e

PR #3191, opened 2023-04-06
Issue #3188 Lustre billing implementation

mzueva:
Implemented cost estimation for lustre storage based on capacity & throughput.


SilinPavel (approved):
Seems good to me

Issue #3187, opened 2023-04-05
Add storage id to storage audit logs


mzueva:
**Background**

Relates #3059.

To allow report generation it would be helpful to add storage id a new field to all audit logs.

**Approach**
Add `storage_id` field to storage audit logs.

PR #3194, opened 2023-04-07
Use storage id in data access audit in api

tcibinan:
Relates #3187.

The pull request adds storage id field to data access audit records in api.


api/src/main/java/com/epam/pipeline/manager/datastorage/providers/aws/s3/S3Helper.java
```diff
@@ -434,7 +435,7 @@ public DataStorageDownloadFileUrl generateDataStorageItemUploadUrl(String bucket
         return generatePresignedUrl(client, expires, ownerTag, DEFAULT_CANNED_ACL.toString(), request);
     }
 
-    public DataStorageFile createFile(String bucket, String path, byte[] contents, String owner)
+    public DataStorageFile createFile(S3bucketDataStorage bucket, String path, byte[] contents, String owner)
```
SilinPavel: let's set all params final in methods that are changed by this PR

SilinPavel (changes requested):

PR #3232, opened 2023-05-11
pipe storage mount: timeout support 

ekazachkova:
## Background
It would be useful to wait for `pipe-fuse` initialisation more accurate.

## Approach
- `pipe-fuse` script sets specific name to the file system
- `pipe storage mount` command waits for `pipe-fuse` script with the following manner:
  - checks if `pipe-fuse` process still alive; if not exists immediately
  - checks mount point readiness; wait if not ready
  - if mount point ready checks file system name  
- If timeout expired terminate existing process
- A new envvar `CP_PIPE_FUSE_MOUNT_DELAY` to manipulate pooling delay (Default 0.5 sec)
- Increase default value for `--timeout` option up to 10 sec
- rename `CP_PIPE_FUSE_MOUNT_TIMEOUT` instead of `CP_PIPE_FUSE_TIMEOUT` to specify `pipe-fuse` waiting timeout

### DIFF
--- a/deploy/contents/install/preferences/launch.system.parameters.json
+++ b/deploy/contents/install/preferences/launch.system.parameters.json
@@ -355,10 +355,10 @@
         "passToWorkers": true
     },
     {
-        "name": "CP_PIPE_FUSE_TIMEOUT",
+        "name": "CP_PIPE_FUSE_MOUNT_DELAY",
         "type": "int",
         "description": "Allows to specify the pipe fuse timeout",
-        "defaultValue": "500",
+        "defaultValue": "10000",
         "passToWorkers": false
     },
     {
--- a/deploy/contents/install/preferences/launch.system.parameters.json
+++ b/deploy/contents/install/preferences/launch.system.parameters.json
@@ -355,7 +355,7 @@
         "passToWorkers": true
     },
     {
-        "name": "CP_PIPE_FUSE_MOUNT_DELAY",
+        "name": "CP_PIPE_FUSE_MOUNT_TIMEOUT",
         "type": "int",
         "description": "Allows to specify the pipe fuse timeout",
         "defaultValue": "10000",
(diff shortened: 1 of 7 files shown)


pipe-cli/src/utilities/storage/mount.py
```diff
@@ -194,8 +201,57 @@ def run(self, config, mount_cmd, mount_timeout=5*MS_IN_SEC, python_path=None, lo
                                               stdout=output,
                                               stderr=subprocess.STDOUT,
                                               env=mount_environment)
+            self._wait_mount_point(mount_timeout, mount_aps_proc, mountpoint)
+
+    def _wait_mount_point(self, mount_timeout, mount_aps_proc, mountpoint):
+        max_init_try = os.environ.get('CP_MOUNT_MAX_INIT_TRY', 100)
```
tcibinan: What do you think if we take existing `mount_timeout` as a timeout and just check every second if it is ready. In that case we also need to change default timeout value in `pipe storage mount` command from 1 to f.e. 60 seconds.

tcibinan (changes requested):
As long as we use such low level operations, I would suggest testing these changes on the most popular distributions such as centos, ubuntu and debian using non root user.

Additionally, I think that we should test the changes on Windows for non admin user as well.

deploy/contents/install/preferences/launch.system.parameters.json
```diff
@@ -355,10 +355,10 @@
         "passToWorkers": true
     },
     {
-        "name": "CP_PIPE_FUSE_TIMEOUT",
+        "name": "CP_PIPE_FUSE_MOUNT_DELAY",
```
tcibinan: It should be `CP_PIPE_FUSE_MOUNT_TIMEOUT` here.

workflows/pipe-common/scripts/mount_storage.py
tcibinan, line 1: Could you please perform the same in `syncmounts.sh`. The script also uses `CP_PIPE_FUSE_TIMEOUT`.

tcibinan (changes requested):

tcibinan (approved):

PR #3267, opened 2023-06-05
Introduce grid engine autoscaler package

### LINKED ISSUE #2498: Extract grid engine autoscaler script to a separate module
**Background**

Once #2492 is merged `workflows/pipe-common/scripts/autoscale_sge.py` script will become almost 2k lines. The script consists of a lot of classes with different responsibilities such as grid engine management, autoscaling instances selection, scaling up/down handling, autoscaling cycling, faults recovering and testing.

**Approach**

It seems that the autoscaling script became pretty complex and should be managed as a separate application in cloud pipeline similarly to pipe cli, pipe fuse, fs browser and etc.

Once grid engine autoscaler is a separate application its classes can be split into multiple files and built and distributed as a single python library.

tcibinan:
Resolves #2498.

The pull request extracts grid engine autoscaler `pipeline.hpc` package from original `autoscale_sge.py` script.


workflows/pipe-common/pipeline/hpc/ssh.py
```diff
@@ -0,0 +1,43 @@
+#  Copyright 2017-2023 EPAM Systems, Inc. (https://www.epam.com/)
```
SilinPavel: why this file has a name ssh? maybe cmd will be more accurate?

workflows/pipe-common/pipeline/hpc/avail.py
```diff
@@ -0,0 +1,58 @@
+#  Copyright 2017-2023 EPAM Systems, Inc. (https://www.epam.com/)
```
SilinPavel: I think it is better to emphasize that it is about instance, so how about to name this file as `instance_avail`

SilinPavel (changes requested):

workflows/pipe-common/pipeline/hpc/ssh.py
tcibinan: Done

workflows/pipe-common/pipeline/hpc/avail.py
tcibinan: I've extracted `instance` package with `avail`, `provider` and `selector` subpackages.

SilinPavel: LGTM

SilinPavel (approved):

PR #3276, opened 2023-06-08
Introduce RSMAP-like GPU devices mgmt in SGE

### LINKED ISSUE #3274: Support RSMAP-like GPU devices mgmt in SGE
**Background**

Currently, we use comsumable `gpus` resource in SGE to distribute GPU jobs across cluster. The consumable resource is of `INT` type and represents a number of GPUs on a single instance.

Certain applications expect grid engine to automatically assign available, not yet occupied GPU devices to individual jobs. Usually, such mechanism is implemented via a consumable resource of `RSMAP` type. Unfortunately, SGE version which is used in Cloud Pipeline doesn't support `RSMAP` type.

**Approach**

Support for `RSMAP`-like GPU devices assignment shall be implemented.

tcibinan:
Resolves #3274.

The pull request brings support for RSMAP-like GPU devices assignment in SGE. If enabled, then all jobs which require GPUs via `-l gpus=2` will have `CUDA_VISIBLE_DEVICES` environment variable set to unoccupied GPU device ids like `0`, `2,3` and etc. 

The following run parameters are introduced:
- `CP_CAP_SGE_GPU_RSMAP_DEVICES` enables RSMAP-like GPU devices assignment for all existing and newly created SGE queues.

### DIFF
--- a/workflows/pipe-common/shell/sge_epilog
+++ b/workflows/pipe-common/shell/sge_epilog
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# Copyright (c) 2015 Kota Yamaguchi
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+# 3. The name of the author may not be used to endorse or promote products
+#    derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+# OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+# IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+# NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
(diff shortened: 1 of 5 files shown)


workflows/pipe-common/shell/sge_epilog
```diff
@@ -0,0 +1,58 @@
...
+for GPU_DEVICE in $JOB_GPU_DEVICES; do
+    echo -n "Releasing GPU device $GPU_DEVICE... "
+    if ! rmdir "$CP_CAP_GE_CONSUMABLE_RESOURCE_LOCKS_GPU/inuse/$GPU_DEVICE" 2>/dev/null; then
+        echo "NOT OK"
+        continue
+    fi
+    echo "OK"
+done
+
+exit 0
```
SilinPavel, line 58: we exiting here with 0 code without any concerns if all GPUs released above or not. Was it intention?

workflows/pipe-common/shell/sge_epilog
```diff
@@ -0,0 +1,58 @@
...
+JOB_GPUS_NUMBER="${JOB_GPUS_NUMBER:-0}"
+
+if [ "$JOB_GPUS_NUMBER" -le 0 ]; then
+    exit 0
+fi
+
+JOB_GPU_DEVICES=$(grep -oP "CUDA_VISIBLE_DEVICES=\K.*" "$JOB_ENV_FILE" | sed -e "s/,/ /g")
+for GPU_DEVICE in $JOB_GPU_DEVICES; do
+    echo -n "Releasing GPU device $GPU_DEVICE... "
+    if ! rmdir "$CP_CAP_GE_CONSUMABLE_RESOURCE_LOCKS_GPU/inuse/$GPU_DEVICE" 2>/dev/null; then
```
SilinPavel, line 51: in which cases we will not be able to delete it? Can we forcefully delete directory here? 
I worry about the case when for some reason we will not remove it and next jobs will fail because of `synthetic` insufficient GPUs.

workflows/pipe-common/shell/sge_epilog
tcibinan: In case any of the previously executing commands fails f.e. _grep_ call, the epilog script may exit with non zero exit code. This is very dangerous because if an epilog script ends with non zero or non hundred exit code then the corresponding queue will be automatically disabled until it is manually enabled again.

workflows/pipe-common/shell/sge_epilog
tcibinan: Generally, it is a very case related to prolog and epilog scripts misconfiguration and I think it is quite ok to have such a warning. As I see it, most likely, it can be a permission level issue. In that case forceful deletion won't help but we will still have the warning for troubleshooting. So I would propose to keep it as it is.

SilinPavel: Got it. I agree

workflows/pipe-common/shell/sge_epilog
SilinPavel: Ok, sounds good

SilinPavel (approved):

PR #3278, opened 2023-06-08
Issue 3271 versioned storage manager role

SilinPavel:
This PR changes ACL checks logic and introduces new role `ROLE_VERSIONED_STORAGE_MANAGER`.
This role is used in the same manner as `ROLE_PIPELINE_MANAGER` but for pipelines with type `VERSIONED_STORAGE`.


api/src/main/java/com/epam/pipeline/entity/user/DefaultRoles.java
```diff
@@ -37,7 +37,10 @@ public enum DefaultRoles {
     ROLE_STORAGE_ARCHIVE_MANAGER(new Role(null, "ROLE_STORAGE_ARCHIVE_MANAGER", true, false, null, null, null, null)),
     ROLE_STORAGE_ARCHIVE_READER(new Role(null, "ROLE_STORAGE_ARCHIVE_READER", true, false, null, null, null, null)),
     ROLE_STORAGE_MANAGER(new Role(null, "ROLE_STORAGE_MANAGER", true, false, null, null, null, null)),
-    ROLE_STORAGE_TAG_MANAGER(new Role(null, "ROLE_STORAGE_TAG_MANAGER", true, false, null, null, null, null));
+    ROLE_STORAGE_TAG_MANAGER(new Role(null, "ROLE_STORAGE_TAG_MANAGER", true, false, null, null, null, null)),
+    ROLE_PIPELINE_MANAGER(new Role(null, "ROLE_PIPELINE_MANAGER", true, false, null, null, null, null)),
+    ROLE_VERSIONED_STORAGE_MANAGER(new Role(null, "ROLE_VERSIONED_STORAGE_MANAGER", true, false,
```
mzueva, line 42: We also need sql migration to create new role `ROLE_VERSIONED_STORAGE_MANAGER `

api/src/main/java/com/epam/pipeline/manager/security/pipeline/PipelinePermissionManager.java
```diff
@@ -0,0 +1,74 @@
...
+public class PipelinePermissionManager {
+
+    private static final String WRITE = "WRITE";
+    private static final String READ = "READ";
+
+    private final CheckPermissionHelper permissionHelper;
+    private final EntityManager entityManager;
+
+    public boolean hasCreatePermission(final PipelineType type, final Long folderId) {
+        Folder folder = (Folder) entityManager.load(AclClass.FOLDER, folderId);
```
mzueva: I think it is better check `foldeId` for null before loading the enity
Missing `final`

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/manager/security/pipeline/PipelinePermissionManager.java
```diff
@@ -0,0 +1,74 @@
...
+        return hasCopyPermission(pipeline, folderId);
+    }
+
+    public boolean hasCopyPermission(final Pipeline pipeline, final Long folderId) {
+        return permissionHelper.isAllowed(READ, pipeline)
+                && folderId != null && hasCreatePermission(pipeline.getPipelineType(), folderId)
+                && checkManagerRole(pipeline.getPipelineType());
+    }
+
+    private boolean checkManagerRole(final PipelineType type) {
```
mzueva, line 64: Please add check for `type` value for null (let's consider it `PIPELINE` by default)

SilinPavel: I believe that it is enough to just to check that `type == PipelineType.VERSIONED_STORAGE`, because if type is null we will go to else branch as a default behavior

api/src/main/java/com/epam/pipeline/entity/user/DefaultRoles.java
SilinPavel: Great catch!

PR #3286, opened 2023-06-14
Issue 3285 change approach to scan tool

SilinPavel:
Now we scan tool independently in Docker Comp and Clair services.
If any of these services will fail, results of another one will be persisted any way.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/docker/scan/AggregatingToolScanManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/docker/scan/AggregatingToolScanManager.java
@@ -58,7 +58,6 @@ import okhttp3.ResponseBody;
 import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


mzueva (approved):

api/src/main/java/com/epam/pipeline/manager/docker/scan/AggregatingToolScanManager.java
```diff
@@ -433,49 +444,62 @@ private DockerClient getDockerClient(String tool, DockerRegistry registry) {
...
-                                                 final boolean hasCudnnVersion) {
+    private ToolVersionScanResult gatherResults(final Tool tool, final String tag,
+                                                final String lastLayerRef, final String digest,
+                                                final String defaultCmd, final int layersCount,
+                                                final boolean hasCudnnVersion) throws IOException {
+
+        ToolScanStatus toolScanStatus = ToolScanStatus.COMPLETED;
+
+        final Pair<Boolean, ClairScanResult> clairScanResult =
+                getScanResult(clairService != null, () -> clairService.getScanResult(lastLayerRef));
```
tcibinan: It looks like the pair can be in two forms only: `(false, null)` and `(true, value)`. What do you think if we just use optional here? And same for `dockerComponentService`.

```suggestion
        final Optional<ClairScanResult> clairScanResult = Optional.ofNullable(clairService)
            .map(service -> service.getScanResult(lastLayerRef));
```

SilinPavel: good approach, thanks!

tcibinan (approved):

PR #3316, opened 2023-07-07
GUI Nat gateway table improvements: table search, separate recent cha…

AleksandrGorodetskii:
…nges table rows


rodichenko:
Merged with f3bad9f401e35c17c5af9102db4627b157832410

PR #3320, opened 2023-07-31
Grid Engine Autoscaling: SLURM support

SilinPavel:
Relates #3330.

This PR adds new grid engine provider `SLURM` for autoscaling.


workflows/pipe-common/pipeline/hpc/gridengine.py
```diff
@@ -484,3 +546,226 @@ def validate(self, jobs):
...
+        _perform_command(
+            action=lambda: self.disable_host(host),
+            msg='Disabling GE host execution daemon.',
+            error_msg='Disabling GE host execution daemon has failed.',
+            skip_on_failure=skip_on_failure
+        )
+        _perform_command(
+            action=lambda: self.cmd_executor.execute(SlurmGridEngine._SCONTROL_DELETE_NODE % host),
+            msg='Disabling GE host execution daemon.',
+            error_msg='Disabling GE host execution daemon has failed.',
```
tcibinan: Rather should be

```suggestion
            msg='Remove host from GE.',
            error_msg='Removing host from GE has failed.',
```

workflows/pipe-common/pipeline/hpc/gridengine.py
```diff
@@ -484,3 +546,226 @@ def validate(self, jobs):
...
+    def get_pe_allocation_rule(self, pe):
+        raise AllocationRuleParsingError("Slurm doesn't have PE preference.")
+
+    def delete_host(self, host, skip_on_failure=False):
+        _perform_command(
+            action=lambda: self.disable_host(host),
+            msg='Disabling GE host execution daemon.',
+            error_msg='Disabling GE host execution daemon has failed.',
+            skip_on_failure=skip_on_failure
+        )
```
tcibinan: No need to disable host here because `self.disable_host` should be called beforehand by `GridEngineScaleDownHandler`.

workflows/pipe-common/pipeline/hpc/gridengine.py
tcibinan, line 1: I would rather move all grid engine specific implementations to separate `sge.py` and `slurm.py` files which can import all the common stuff from `gridengine.py`.

workflows/pipe-common/pipeline/hpc/gridengine.py
```diff
@@ -484,3 +546,226 @@ def validate(self, jobs):
...
+        return ResourceSupply()
+
+    def get_engine_type(self):
+        return GridEngineType.SLURM
+
+    def is_valid(self, host):
+        node_state = self._get_host_state(host)
+        for bad_state in SlurmGridEngine._NODE_BAD_STATES:
+            if bad_state in node_state:
+                return False
```
tcibinan: Let's log bad_state here.

```suggestion
                Logger.warn('Execution host %s GE state is %s which makes host invalid.' % (host, bad_state))
                return False
```

workflows/pipe-common/pipeline/hpc/gridengine.py
```diff
@@ -484,3 +546,226 @@ def validate(self, jobs):
...
+        return True
+
+    def kill_jobs(self, jobs, force=False):
+        job_ids = [str(job.root_id) for job in jobs]
```
tcibinan: Need only unique root ids here.
```suggestion
        job_ids = set(str(job.root_id) for job in jobs)
```

workflows/pipe-common/pipeline/hpc/gridengine.py
```diff
@@ -484,3 +546,226 @@ def validate(self, jobs):
...
+            [
+                entry.split(value_sep, 1) for entry in text.split(line_sep)
+            ]
+        }
+
+    def _find_memory_value(self, job_dict, resource_dict):
+        if "MinMemoryNode" in job_dict:
+            return job_dict.get("MinMemoryNode")
+        elif "mem" in resource_dict:
+            resource_dict.get("mem")
```
tcibinan: Missing return?

```suggestion
            return resource_dict.get("mem")
```

As long as we have such complex parsing, let's add tests to cover all corner cases which we use in `self._parse_jobs`.

workflows/pipe-common/pipeline/hpc/param.py
```diff
@@ -249,6 +249,12 @@ def __init__(self):
         self.event_ttl = GridEngineParameter(
             name='CP_CAP_AUTOSCALE_EVENT_TTL', type=PARAM_INT, default=3 * 60 * 60,
             help='Specifies event ttl in seconds after which an event is removed.')
+        self.sge_selected = GridEngineParameter(
+            name='CP_CAP_SGE', type=PARAM_BOOL, default=True,
+            help='Defines if SGE selected as grid engine.')
+        self.slurm_selected = GridEngineParameter(
+            name='CP_CAP_SLURM', type=PARAM_BOOL, default=False,
+            help='Defines if Slurm selected as grid engine.')
```
tcibinan: Let's move these parameters to `GridEngineQueueParameters` class instead because they describe a queue that should be autoscaled.

tcibinan (changes requested):

workflows/pipe-common/pipeline/hpc/param.py
SilinPavel: From my point of view its describing general settings of autoscaling rather then queue, don't you think?

tcibinan: In practice these two settings specify a type of a grid engine queue. And I believe that it is the characteristic of a queue the same way as queue name is.

workflows/pipe-common/pipeline/hpc/gridengine.py
SilinPavel: yep, good idea

workflows/pipe-common/scripts/autoscale_grid_engine.py
```diff
@@ -29,7 +29,7 @@
 from pipeline.hpc.cloud import CloudProvider
 from pipeline.hpc.cmd import CmdExecutor
 from pipeline.hpc.event import GridEngineEventManager
-from pipeline.hpc.gridengine import SunGridEngine, SunGridEngineDemandSelector, SunGridEngineJobValidator, SlurmGridEngine, SlurmDemandSelector, SlurmJobValidator, GridEngineType
+from pipeline.hpc.engine.gridengine import SunGridEngine, SunGridEngineDemandSelector, SunGridEngineJobValidator, SlurmGridEngine, SlurmDemandSelector, SlurmJobValidator, GridEngineType
```
tcibinan: Can we import slurm and sge classes from the corresponding packages instead?

Issue #3326, opened 2023-08-07
Integrate Slurm and SGE with Cloud-Pipeline to notify about Job status by email.


SilinPavel:
**Background**
In some cases it would be nice to be able to configure Slurm and SGE to send emails about `Job` statuses to users.
For this purpose it seems logical to send these messages through our existing notification system.

PR #3327, opened 2023-08-07
Issue 3326 Slurm and SGE integration with Cloud-Pipeline notifications

SilinPavel:
Relates to #3326

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
@@ -745,17 +745,20 @@ public class NotificationManager implements NotificationService { // TODO: rewri
         final List<Long> copyUserIds = ListUtils.emptyIfNull(messageVO.getCopyUsers())
                 .stream()
                 .filter(StringUtils::isNotBlank)
-                .map(this::getUserByName)
+                .map(this::getUserByNameOrEmail)
                 .map(PipelineUser::getId)
                 .collect(Collectors.toList());
         message.setCopyUserIds(copyUserIds);
-        message.setToUserId(getUserByName(messageVO.getToUser()).getId());
+        message.setToUserId(getUserByNameOrEmail(messageVO.getToUser()).getId());
         return message;
     }
 
-    private PipelineUser getUserByName(final String username) {
-        final PipelineUser user = userManager.loadUserByName(username);
-        Assert.notNull(user, messageHelper.getMessage(MessageConstants.ERROR_USER_NAME_NOT_FOUND, username));
+    private PipelineUser getUserByNameOrEmail(final String usernameOrEmail) {
+        PipelineUser user = userManager.loadUserByName(usernameOrEmail);
+        if (user == null) {
+            user = userManager.loadUserByEmail(usernameOrEmail);
+        }
+        Assert.notNull(user, messageHelper.getMessage(MessageConstants.ERROR_USER_NAME_NOT_FOUND, usernameOrEmail));
         return user;
     }
 
--- a/api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
@@ -755,7 +755,7 @@ public class NotificationManager implements NotificationService { // TODO: rewri
 
     private PipelineUser getUserByNameOrEmail(final String usernameOrEmail) {
         final PipelineUser user = Optional.ofNullable(userManager.loadUserByName(usernameOrEmail))
-                .orElse(userManager.loadUserByEmail(usernameOrEmail));
+                .orElseGet(() -> userManager.loadUserByEmail(usernameOrEmail));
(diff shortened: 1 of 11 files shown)


api/src/main/java/com/epam/pipeline/manager/notification/NotificationManager.java
```diff
@@ -745,17 +745,18 @@ private NotificationMessage toMessage(final NotificationMessageVO messageVO) {
...
+        message.setToUserId(getUserByNameOrEmail(messageVO.getToUser()).getId());
         return message;
     }
 
-    private PipelineUser getUserByName(final String username) {
-        final PipelineUser user = userManager.loadUserByName(username);
-        Assert.notNull(user, messageHelper.getMessage(MessageConstants.ERROR_USER_NAME_NOT_FOUND, username));
+    private PipelineUser getUserByNameOrEmail(final String usernameOrEmail) {
+        final PipelineUser user = Optional.ofNullable(userManager.loadUserByName(usernameOrEmail))
+                .orElse(userManager.loadUserByEmail(usernameOrEmail));
```
tcibinan: It's better to only call the second method if the first one didn't succeed.

```suggestion
                .orElseGet(() -> userManager.loadUserByEmail(usernameOrEmail));
```

scripts/pipeline-launch/launch.sh
```diff
@@ -729,6 +729,7 @@ function configure_owner_account() {
     if [ "$OWNER" ]; then
         # Crop OWNER account by @ if present
         IFS='@' read -r -a owner_info <<< "$OWNER"
+        export REMOTE_OWNER="$OWNER"
```
tcibinan: Why is it called remote owner? Maybe something like `OWNER_ACCOUNT`, `OWNER_USER_NAME` or `OWNER_NAME` works better. What do you think?

workflows/pipe-common/shell/pipe_mail
```diff
@@ -0,0 +1,85 @@
...
+
+function construct_email_subject {
+    local _grid_engine_msg=$1
+    echo "Run #$RUN_ID. Job info: $_grid_engine_msg"
+}
+
+function get_cp_user_id {
+    local _provider_user_id=$1
+    local _effective_user_id="$_provider_user_id"
+    echo "Sending to $_provider_user_id ..." >> ${_PIPE_MAIL_LOG_FILE}
```
tcibinan: Probably, we can add the following snippet to the start of the script.

```
exec >> file
exec 2>&1
```

Then we can get rid of specific redirections altogether.

```suggestion
    echo "Sending to $_provider_user_id ..."
```

workflows/pipe-common/shell/pipe_mail
```diff
@@ -0,0 +1,85 @@
...
+    ;;
+esac
+done
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
+echo "Sending emails to: $@" >> ${_PIPE_MAIL_LOG_FILE}
+while [[ $# -gt 0 ]]
+do
+    _to_user=$(get_cp_user_id "$1")
+    pipe_notify "$_NOTIFICATION_SUBJECT" "$_NOTIFICATION_SUBJECT" "$_to_user" &>> ${_PIPE_MAIL_LOG_FILE}
```
tcibinan: Can we also add the same notifications to pipeline run logs?

```
pipe_log_info "Receiver $_to_user. $_NOTIFICATION_SUBJECT" "Notifications"
```

workflows/pipe-common/shell/sge_setup_master
```diff
@@ -60,6 +60,20 @@ configure_global() {
...
+        qconf -sconf $HOSTNAME > ./$HOSTNAME
+        if [ "$?" -eq "0" ]; then
+            sed -i "/mailer/c\mailer $COMMON_REPO_DIR/shell/pipe_mail" ./$HOSTNAME
+            qconf -Mconf ./$HOSTNAME
+        else
+            echo "mailer $COMMON_REPO_DIR/shell/pipe_mail" > ./$HOSTNAME
+            qconf -Aconf ./$HOSTNAME
+        fi
+        rm -f ./$HOSTNAME
+    fi
```
tcibinan: Let's configure only global host here and use some universal paths like `LOCAL_COMMON_REPO_DIR`.

```bash
export LOCAL_RUN_DIR="/runs-local/pipeline"
export LOCAL_COMMON_REPO_DIR="$LOCAL_RUN_DIR/CommonRepo"
ln -s "$RUN_DIR" "$LOCAL_RUN_DIR"
```

scripts/pipeline-launch/launch.sh
SilinPavel: Changed to OWNER_CP_ACCOUNT, thanks!

workflows/pipe-common/shell/sge_setup_master
SilinPavel: After a discussion:
 - linking only COMMON_REPO_DIR since it can be overridden in a first place, and linking of RUN_DIR to some common location won't solve a problem
 - using `COMMON_REPO_DIR_MUTUAL_LOC` env to store path to mutual location of CommonRepo dir (default value is `/usr/local/CommonRepo`)

tcibinan (approved):

PR #3332, opened 2023-08-11
Issue #3326. Fix SGE installation: link pipe_mail to /usr/bin/mail

SilinPavel:
This PR fixes problem with SGE notification integration.
Due to strange behavior of sge_execd with updating global config values we changed logic as follow:
 - instead of changing `mailer` value if sge_conf we just link `pipe_mail` to `/usr/bin/mail` 
 - introducing new `CP_CAP_MAIL` system launch param, through this parameter custom `pipe_mail` client can be enabled

### DIFF
--- a/scripts/pipeline-launch/launch.sh
+++ b/scripts/pipeline-launch/launch.sh
@@ -221,6 +221,19 @@ function cp_cap_publish {
             
             sed -i "/$_SGE_WORKER_INIT/d" $_WORKER_CAP_INIT_PATH
             echo "$_SGE_WORKER_INIT" >> $_WORKER_CAP_INIT_PATH
+
+            # Due to strange behavior of sge_execd with updating global config values,
+            # we need to change default mail client if favor to our custom
+            if check_cp_cap "CP_CAP_GRID_ENGINE_NOTIFICATIONS"
+            then
+                _PIPE_MAIL_ENABLER_SCRIPT="pipe_mail_enabler"
+
+                sed -i "/$_PIPE_MAIL_ENABLER_SCRIPT/d" $_MASTER_CAP_INIT_PATH
+                echo "$_PIPE_MAIL_ENABLER_SCRIPT" >> $_MASTER_CAP_INIT_PATH
+
+                sed -i "/$_PIPE_MAIL_ENABLER_SCRIPT/d" $_WORKER_CAP_INIT_PATH
+                echo "$_PIPE_MAIL_ENABLER_SCRIPT" >> $_WORKER_CAP_INIT_PATH
+            fi
       fi
 
       if check_cp_cap "CP_CAP_SLURM"
@@ -1806,6 +1819,16 @@ echo "------"
 echo
 ######################################################
 
+
+######################################################
+echo "Configure custom mail client if requested"
+echo "-"
+######################################################
+
+if check_cp_cap CP_CAP_MAIL; then
+    pipe_mail_enabler
+fi
+######################################################
 
 
 ######################################################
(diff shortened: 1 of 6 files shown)


scripts/pipeline-launch/launch.sh
```diff
@@ -221,6 +221,19 @@ function cp_cap_publish {
...
+            if check_cp_cap "CP_CAP_GRID_ENGINE_NOTIFICATIONS"
+            then
+                _PIPE_MAIL_ENABLER_SCRIPT="pipe_mail_enabler"
+
+                sed -i "/$_PIPE_MAIL_ENABLER_SCRIPT/d" $_MASTER_CAP_INIT_PATH
+                echo "$_PIPE_MAIL_ENABLER_SCRIPT" >> $_MASTER_CAP_INIT_PATH
+
+                sed -i "/$_PIPE_MAIL_ENABLER_SCRIPT/d" $_WORKER_CAP_INIT_PATH
+                echo "$_PIPE_MAIL_ENABLER_SCRIPT" >> $_WORKER_CAP_INIT_PATH
+            fi
```
tcibinan, line 236: What do you think if we simplify it by moving this to the beginning of launch.sh?

```bash
if check_cp_cap CP_CAP_GRID_ENGINE_NOTIFICATIONS && check_cp_cap CP_CAP_SGE; then
    export CP_CAP_MAIL="true"
fi
```

workflows/pipe-common/shell/pipe_mail_enabler
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+echo "Linking $COMMON_REPO_DIR_MUTUAL_LOC/shell/pipe_mail as default linux mail client"
+rm -rf "/usr/bin/mail" && ln -s "$COMMON_REPO_DIR_MUTUAL_LOC/shell/pipe_mail" "/usr/bin/mail" \
+    && rm -rf "/bin/mail" && ln -s "$COMMON_REPO_DIR_MUTUAL_LOC/shell/pipe_mail" "/bin/mail"
```
tcibinan, line 18: This code won't link `/bin/mail` if `/usr/bin/mail` is missing. I would suggest the following implementation which links both files independently.
```suggestion
function symlink_binary() {
    local source="$1"
    local target="$1"

    if [ ! -f "$target" ]; then
        echo "Failed to link $target->$source"
        return 1
    fi

    mv "$target" "$target.backup"
    ln -s "$source" "$target"
    echo "Linked $target->$source"
}

symlink_binary "$COMMON_REPO_DIR_MUTUAL_LOC/shell/pipe_mail" "/bin/mail"
symlink_binary "$COMMON_REPO_DIR_MUTUAL_LOC/shell/pipe_mail" "/usr/bin/mail"
```

scripts/pipeline-launch/launch.sh
SilinPavel: that was my idea too, but the thing is that in this case we won't be able to propagate it to workers, since CP_CAP_SGE will not be set to `true` for them

workflows/pipe-common/shell/pipe_mail_enabler
SilinPavel: Do you mean that `rm -rf ` will return bad error code? I thing that `rm -f` will return `0` even if file doesn't exist

Issue #3336, opened 2023-08-17
Remove unavailable node from kube cluster if it already removed from a cloud


SilinPavel:
**Background**
Cloud-Pipeline has a feature to detect unavailable nodes. A node can be unavailable in two cases:
 - Some temperately problem, such as OOM, when kubelet would be killed. After restart of kubelet service such node will be back to normal state
 - Terminal events. Such as termination of a node by some 3rd party or spot instance that was taken from us.

In the second case we know for sure that such node will never come back. There is no need to wait some time (`cluster.node.unavailable.grace.period.minutes`), and we can remove such node from cluster immediately.

PR #3338, opened 2023-08-18
Issue #3336. Remove node from cluster immediately if it isn't present on a cloud

SilinPavel:
Related to issue #3336  
Check if node isn't present on a cloud while tagging it as unavailable and if so - remove in immediately.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSInstanceService.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSInstanceService.java
@@ -267,20 +267,24 @@ public class AWSInstanceService implements CloudInstanceService<AwsRegion> {
 
     @Override
     public CloudInstanceState getInstanceState(final AwsRegion region, final String nodeLabel) {
-        final Instance aliveInstance = ec2Helper.getAliveInstance(nodeLabel, region);
-        if (Objects.isNull(aliveInstance)) {
-            return CloudInstanceState.TERMINATED;
-        }
-        final String instanceStateName = aliveInstance.getState().getName();
-        if (InstanceStateName.Pending.toString().equals(instanceStateName)
-                || InstanceStateName.Running.toString().equals(instanceStateName)) {
-            return CloudInstanceState.RUNNING;
-        }
-        if (InstanceStateName.Stopping.toString().equals(instanceStateName)) {
-            return CloudInstanceState.STOPPING;
-        }
-        if (InstanceStateName.Stopped.toString().equals(instanceStateName)) {
-            return CloudInstanceState.STOPPED;
+        try {
+            final Instance aliveInstance = ec2Helper.getAliveInstance(nodeLabel, region);
+            if (Objects.isNull(aliveInstance)) {
+                return CloudInstanceState.TERMINATED;
+            }
+            final String instanceStateName = aliveInstance.getState().getName();
+            if (InstanceStateName.Pending.toString().equals(instanceStateName)
+                    || InstanceStateName.Running.toString().equals(instanceStateName)) {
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/ScaleDownHandler.java
```diff
@@ -198,6 +212,32 @@ private void log(final PipelineRun run, final String text) {
...
+            final Long cloudRegionId = pipelineRunManager.findRun(runId)
+                    .map(PipelineRun::getInstance)
+                    .map(RunInstance::getCloudRegionId)
+                    .orElseThrow(
+                            () -> new IllegalArgumentException(
+                                    String.format("Can't find region by runId: %d", runId))
+                    );
+            state = cloudFacade.getInstanceTerminationState(cloudRegionId, instanceId);
+            existsOnCloud = cloudFacade.instanceExists(cloudRegionId, instanceId);
+        }
```
tcibinan: Do we really need to do two requests here? Because both instance existence check and termination state retrieval is e.g. `describe instances` request on AWS.

It seems like we can just use a single call to `CloudFacade::getInstanceState` here.

SilinPavel: Yep, you are right. With `CloudFacade::getInstanceState` that will be more efficient!

mzueva (approved):

api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/ScaleDownHandler.java
```diff
@@ -215,27 +214,16 @@ private void log(final PipelineRun run, final String text) {
...
-                    );
-            state = cloudFacade.getInstanceTerminationState(cloudRegionId, instanceId);
-            existsOnCloud = cloudFacade.instanceExists(cloudRegionId, instanceId);
+            state = cloudFacade.getInstanceState(runId);
         }
-        log.debug("Check if unavailable node {} is alive. Exists: {}. Termination state: {}", instanceId, existsOnCloud,
-                state.map(InstanceTerminationState::getStateMessage).orElse(EMPTY_TERMINATION_STATE));
-        return existsOnCloud && !state.isPresent();
+        log.debug("Check if unavailable node {} is alive. Termination state: {}", instanceId, state);
+        return state != null && state != CloudInstanceState.TERMINATED;
```
tcibinan: When an instance gets preempted on GCP it is getting stopped rather than terminated. So it seems like this won't work for GCP at least. I think that we should check if instance is running.

```suggestion
        return state == CloudInstanceState.RUNNING;
```

api/src/main/java/com/epam/pipeline/manager/cloud/aws/AWSInstanceService.java
```diff
@@ -267,20 +267,24 @@ public List<InstanceDisk> loadDisks(final AwsRegion region, final Long runId) {
...
+            if (InstanceStateName.Stopping.toString().equals(instanceStateName)) {
+                return CloudInstanceState.STOPPING;
+            }
+            if (InstanceStateName.Stopped.toString().equals(instanceStateName)) {
+                return CloudInstanceState.STOPPED;
+            }
+        } catch (AwsEc2Exception e) {
+            log.error("Fail to get instance state by instance label {} for regionId {}", nodeLabel, region.getId());
         }
         return null;
```
tcibinan: All clouds but AWS return `CloudInstanceState.TERMINATED` if nothing else works. Can we make it common for all clouds?

tcibinan (approved):

Issue #1019, opened 2020-03-27
Introduce an API method, that can be used to refresh the prices for a specific Cloud Region


sidoruka:
At the moment, it's hard to refresh the existing price list in the Cloud Pipeline. 

It would be nice to have a separate API to:
* Reload the prices for a specific region
* Reload the prices for all the regions

PR #3361, opened 2023-09-07
Introduce configurable price list filtering and batching

tcibinan:
Relates #1019.

The pull request introduces configurable filtering and batching to price list refresh operations.

The following system preferences are introduced:

- `instance.offer.fetch.gpu` enables instance offer gpu details fetching. Defaults to *true*.
- `instance.offer.filter.unique` enables unique instance offers filtering. Defaults to *true*.
- `instance.offer.filter.term.types` specifies a comma-separated list of allowed term types of instance offers. Defaults to *OnDemand,Spot,LowPriority,Preemptible*.
- `instance.offer.insert.batch.size` specifies inserting batch size for instance offers. Defaults to *10000*.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/cluster/InstanceOfferDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/cluster/InstanceOfferDao.java
@@ -23,22 +23,26 @@ import com.epam.pipeline.entity.cluster.InstanceOffer;
 import com.epam.pipeline.entity.cluster.InstanceType;
 import com.epam.pipeline.entity.region.CloudProvider;
 import com.epam.pipeline.utils.StreamUtils;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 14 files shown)


api/src/main/java/com/epam/pipeline/dao/cluster/InstanceOfferDao.java
```diff
@@ -75,9 +75,9 @@ public void removeInstanceOffersForRegion(Long regionId) {
     }
 
     @Transactional
-    public void replaceInstanceOffersForRegion(final Long id, final List<InstanceOffer> offers) {
+    public void replaceInstanceOffersForRegion(final Long id, final List<InstanceOffer> offers, int batchSize) {
```
ekazachkova: Could you add final to arg `batchSize `?

ekazachkova (commented):
Also, could you add license to all new files?

api/src/main/java/com/epam/pipeline/dao/cluster/InstanceOfferDao.java
tcibinan: Done

ekazachkova (approved):

PR #3426, opened 2024-01-12
Mount storages improvements

ekazachkova:
This PR provides a couple of features related to mount storages process.

1. `CP_CAP_LIMIT_MOUNTS` , `CP_CAP_SKIP_MOUNTS` and `CP_CAP_FORCE_MOUNTS` parameters shall respect storage paths. 
#### Background
Datastorages may be unregistered and registered again with a new ID. 

#### Usage example:
```
CP_CAP_LIMIT_MOUNTS=1,2,bucket,3,bucket2/path2
```

Note: This PR provides changes for mount during the launch run case. `pipe-fuse` process may fail in running launches due to inability to refresh temporary credentials.

2. Implement an ability to rewrite mount options for separate launch for specific storages.
- add `CP_CAP_MOUNT_OPTIONS_` prefix to  run parameters
#### Usage example:
```
CP_CAP_MOUNT_OPTIONS_2="mount_path=/mount_point:mount_params=-o nolock"
CP_CAP_MOUNT_OPTIONS_3="mount_path=/another_mount_point"
```
- this prefix shall be passed to workers
  - list of such prefixes may be updated via `CP_CAP_AUTOSCALE_INHERITABLE_PARAMETER_PREFIXES`var

### DIFF
--- a/workflows/pipe-common/scripts/mount_storage.py
+++ b/workflows/pipe-common/scripts/mount_storage.py
@@ -332,8 +332,10 @@ class MountStorageTask:
 
     def _load_mount_options_from_environ(self):
         for env_name, env_value in os.environ.items():
-            if env_name.startswith(MOUNT_OPTIONS_ENV_PREFIX):
-                yield int(env_name[len(MOUNT_OPTIONS_ENV_PREFIX):]), MountOptions().parse(env_value)
+            if env_name.startswith(MOUNT_OPTIONS_ENV_PREFIX) and not env_name.endswith('_PARAM_TYPE'):
+                storage_id = env_name[len(MOUNT_OPTIONS_ENV_PREFIX):]
+                if storage_id.isdigit():
+                    yield int(storage_id), MountOptions().parse(env_value)
 
     def _collect_storages_metadata(self, available_storages_with_mounts):
         storages_metadata_raw = self._load_storages_metadata_raw(available_storages_with_mounts)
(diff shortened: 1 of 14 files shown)


workflows/pipe-common/scripts/mount_storage.py
```diff
@@ -301,6 +330,11 @@ def run(self, mount_root, tmp_dir):
             Logger.fail('Unhandled error during mount task: {}.'.format(str(e)), task_name=self.task_name)
             traceback.print_exc()
 
+    def _load_mount_options_from_environ(self):
+        for env_name, env_value in os.environ.items():
+            if env_name.startswith(MOUNT_OPTIONS_ENV_PREFIX):
+                yield int(env_name[len(MOUNT_OPTIONS_ENV_PREFIX):]), MountOptions().parse(env_value)
```
tcibinan: Currently, only storage ids are supported as keys in `CP_CAP_MOUNT_OPTIONS_*`. Don't we want to support storage paths here as well? Because otherwise storage recreation will break storage mounting similarly to `CP_CAP_LIMIT_MOUNTS`.

Breaks storage recreation:
```
export CP_CAP_MOUNT_OPTIONS_1="..."
```

Supports storage recreation:
```
export CP_CAP_MOUNT_OPTIONS_MY_STORAGE_PATH="..."
```

mzueva (approved):

Issue #3436, opened 2024-02-02
Deployment: Introduce new environment variable CP_SERVICES_ENABLED


SilinPavel:
**Background**
For some cases it would be convenient to have possibility to know which services are enabled for deployment.

**Approach**
Let's introduce during deployment a new env `CP_SERVICES_ENABLED`. 
This env should contain:
 - List of names for the services that was configured for deployment, e.g. `cp-api-srv,cp-edge,cp-git` 
 - Or special value `all` if no services were explicitly configured

This env var should be propagated to `install-config` file in order to be presented in `cp-global-config` config map and as result be accessible from service pods such as `edge`.

SilinPavel:
done by https://github.com/epam/cloud-pipeline/pull/3441

PR #3441, opened 2024-02-05
Added new variable for list of installed services

oleksii-mart:
Related to issue #3436, added new variable `CP_SERVICES_ENABLED`, that contain list of names for the services that was configured for deployment.
If no services were explicitly configured in deploy command, the value will be `all`.

### DIFF
--- a/deploy/contents/install/app/install-utils.sh
+++ b/deploy/contents/install/app/install-utils.sh
@@ -610,6 +610,11 @@ function parse_options {
         -s|--service)
         parse_env_option "$2"
         services_count=$((services_count+1))
+        if [ -n "${CP_SERVICES_ENABLED}" ]; then 
+           export CP_SERVICES_ENABLED="${CP_SERVICES_ENABLED},${2}"
+        else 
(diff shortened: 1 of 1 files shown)


deploy/contents/install/app/install-utils.sh
```diff
@@ -627,6 +632,10 @@ function parse_options {
     esac
     done
     set -- "${POSITIONAL[@]}" # restore positional parameters
+
+    if [ -z "${CP_SERVICES_ENABLED}" ]; then
```
SilinPavel: we should make `CP_SERVICES_ENABLED=all` only in case when based on our logic `CP_INSTALL_SERVICES_ALL == 1`. Based on code this place is here:  https://github.com/epam/cloud-pipeline/pull/3441/files#diff-5c368da199a8aa904a21120fa49deab5adb5dfe6f1f66389a61ec978407c7855L745-L748

So, basically we should define `CP_SERVICES_ENABLED="all"` in the same block of code.

oleksii-mart: Fixed

PR #3443, opened 2024-02-06
Changed script in init to check services from CP_SERVICES_ENABLED

oleksii-mart:
Added checking what services was deployed and changed behaviour of init to copy *.conf files to nginx only for those services.


deploy/docker/cp-edge/init
```diff
@@ -120,15 +120,20 @@ export CP_EDGE_INGRESS_API_CONNECT_TIMEOUT="${CP_EDGE_INGRESS_API_CONNECT_TIMEOU
...
-if [ "$CP_KUBE_SERVICES_TYPE" == "ingress" ]; then
-    export CP_DOLLAR="$"
-    for ingress_config in $(ls /etc/nginx/ingress/cp-*.conf); do
-        config_content="$(envsubst < $ingress_config)"
-        echo "$config_content" > "$ingress_config"
-    done
-    unset CP_DOLLAR
-else
-    rm -f /etc/nginx/ingress/*
+services=$CP_SERVICES_ENABLED
```
SilinPavel: let's add default value in case `CP_SERVICES_ENABLED` isn't present. By default let's think that all services included (to work in the same manner as previously)

PR #3462, opened 2024-02-23
Deploy: Added AWS LB Controller addon installation

oleksii-mart:
AWS LB Controller addon installation in EKS cluster by terraform module and helm.


deploy/infra/aws/aws-native-infra/terraform/infra/service-account.tf
```diff
@@ -0,0 +1,16 @@
+# resource "kubernetes_service_account" "service-account" {
```
SilinPavel: should be delete it?

PR #3557, opened 2024-06-10
Collect сore components

ekazachkova:
This PR brings an ability to collect core components - core pods with description if required.

- a new `cloud-pipeline/core-component=<kind>` label shall be added to each core pod. This label indicates that concrete pod shall be considered as core pod. The value shall be equal to pods highest owner kind (`Deployment`/`StatefulSet`/...).
- this label shall be added to `cluster.kube.core.component.label` system preference
- a new API method shall be implemented:

 1.  Load core components
`GET /cluster/pods/core`

Response:
<pod description as for cluster/node/<…>/load>
...
- `parentType` - owner's type (`Deployment`/`StatefulSet`/`CronJob`/...)
- `parentName` - owner's name (`cp-api-srv`/`cp-vm-monitor`/...)
- `status`
	- `reason`
	- `message`
	- `timestamp`  # pod's start timestamp
- `containers`
	<container description as for cluster/node/<…>/load>
	...
	- `restartCount`
	- `lastRestartStatus` (interface as for `status` field)
 2. Load pod description
 `GET cluster/pods/info?podId=<podID>&detailed=<true/false>`
- `podId` - the pod's name [required]
- `detailed` - flag that indicates full pod info loading [default: false]
Response:
- `events`: # list of objects
	- `type` 
	- `reason` 
	- `message` 
- `description` #  string that contains pod info description; returns if detailed=true only
 3. Load container logs
 `GET cluster/containers/logs?podId=<>&containerId=<>&limit=<>`
- `podId` - the pod's name [required]
- `containerId` - the container name [required]
- `limit` - the number of lines to return [default: value from `system.log.line.limit` system preference]

All methods shall be available for `ADMIN` role only.


api/src/main/java/com/epam/pipeline/controller/cluster/ClusterController.java
```diff
@@ -286,4 +288,41 @@ public Result<InstanceDNSRecord> requestDnsRecord(@RequestParam final Long regio
                                                       @RequestBody final InstanceDNSRecord dnsRecord) {
         return Result.success(infrastructureApiService.createInstanceDNSRecord(regionId, dnsRecord));
     }
+
+    @GetMapping("/cluster/core/pods")
```
SilinPavel: I think is it better to swap `core` and `pods` to get the same order as in `/cluster/pods/info`

api/src/main/java/com/epam/pipeline/entity/cluster/CorePodInstance.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode(callSuper = true)
+@Data
+public class CorePodInstance extends PodInstance {
+
+    private String parentType;
+    private String parentName;
+    private PodInstanceStatus status;
```
SilinPavel: What do you think if we will add pod status and containers to the regular `PodInstance` class? do you see any problems with that approach? I think it is reasonable to have these fields not only for `CorePod`

api/src/main/java/com/epam/pipeline/mapper/cluster/KubernetesMapper.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Mapper(componentModel = "spring")
+public interface KubernetesMapper {
+
+    EventEntity mapEvent(Event event);
+
+    default CorePodInstance mapCorePod(final Pod kubePod) {
```
SilinPavel: if we will move status and containers in https://github.com/epam/cloud-pipeline/pull/3557/files#diff-4dfff71cd8f8757c75a2e2887c9632ff9a5b6ea9c4fe336a9b5b0fea351e9f5fR29 to PodInstance, would this method helpful in other places where we map pod object? or maybe we can reuse already existing code to map it and remove this one

SilinPavel (changes requested):

SilinPavel (approved):
good to me

Issue #3297, opened 2023-06-21
Grace period for unavailable nodes


mzueva:
**Background**
At the moment API autoscaler immediately scales down unavailable nodes. Since some nodes may recover from such conditions it would be useful to have a grace period after which node is terminated.

**Approach**
Let's introduce a new preference `cluster.node.unavailable.grace.period`, if it is set to a positive value unavailable nodes shall be terminated only after specified period.

PR #3560, opened 2024-06-18
Allow to specify a grace period for unavailable nodes directly for run

ekazachkova:
Extends #3297

This PR brings ability to specify a grace period after which unavailable nodes are terminated for run using  NODE_UNAVAILABLE_GRACE_PERIOD_MINUTES system parameter.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/ScaleDownHandler.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/ScaleDownHandler.java
@@ -460,7 +460,7 @@ public class ScaleDownHandler {
                 .map(run -> MapUtils.emptyIfNull(run.getEnvVars()).get(NODE_UNAVAILABLE_GRACE_PERIOD_MINUTES))
                 .filter(StringUtils::isNotBlank)
                 .filter(StringUtils::isNumeric)
-                .map(Integer::valueOf)
+                .map(NumberUtils::toInt)
                 .map(Duration::ofMinutes)
                 .orElse(defaultDuration);
     }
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/ScaleDownHandler.java
```diff
@@ -448,4 +453,15 @@ private Optional<String> getTimestampTag(final String name) {
                 .filter(StringUtils::isNotEmpty)
                 .map(suffix -> name + suffix);
     }
+
+    private Duration findNodeUnavailableGraceDuration(final Optional<PipelineRun> pipelineRun,
+                                                      final Duration defaultDuration) {
+        return pipelineRun
```
SilinPavel, line 459: What will happen if `NODE_UNAVAILABLE_GRACE_PERIOD_MINUTES` is not correct? (f.e. empty or string instead of number)
let's double check that we will not fail here, and if we have something inappropriate (like `str` or int < 0), let's return default value here

api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/ScaleDownHandler.java
```diff
@@ -448,4 +453,15 @@ private Optional<String> getTimestampTag(final String name) {
...
+                .map(run -> MapUtils.emptyIfNull(run.getEnvVars()).get(NODE_UNAVAILABLE_GRACE_PERIOD_MINUTES))
+                .filter(StringUtils::isNotBlank)
+                .filter(StringUtils::isNumeric)
```
SilinPavel, line 462: let's also filter out values `< 0`

Issue #3568, opened 2024-06-21
Compose a Dockerfile based on tool version image history


SilinPavel:
User should be able to easily compose dockerfile based on tool image history.

To be able to implement this, lets do the following:

### Server
new method GET `/tool/dockerfile` with parameters:
- `id` - toolId
- `version` - toolVersion
- `from` - string representing the image name to be used in `FROM` Dockerfile directive 

This method should return Dockerfile to download with the following rules:
 - First command with pattern: "ADD file:<> in /" should be replaced as "FROM <parameter from>" 
 - If history has directive ARG <arg-name>=<arg-value>, for all commands that have prepended values <arg-name>=<arg-value> we should clean up such commands. f.i.:
    ```
        ARG TARGETARCH=amd64
        RUN |1 TARGETARCH=amd64 /bin/sh -c yum install -y cuda-libraries-11-3
    ``` 
    Should be transformed to:
    ```
        ARG TARGETARCH=amd64
        RUN  /bin/sh -c yum install -y cuda-libraries-11-3
    ```
 - If the command doesn't start with correct dockerile directive (https://docs.docker.com/reference/dockerfile/) add `RUN` directive to the line
 -  Save only latest `ENTRYPOINT` and `CMD` command and move it to the end of the file
 -  if command `COPY` or `ADD` has patter like `file/folder/multi:<some hash value> in <destination>` as we can't know what is `file/folder/multi:<some hash value>`, let's replace in with `<source-location>` placeholder
 
### UI 
Add to the tool image version history page a button which will generate and download Dockerfile for selected tool version.
 - User should be asked to provide `from` parameter for the Dockerfile

NShaforostov:
Docs added via https://github.com/epam/cloud-pipeline/pull/3669 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/10_Manage_Tools/10.7._Tool_version_menu.md#generate-dockerfile).

PR #3572, opened 2024-06-25
Issue 3568 Compose a Dockerfile based on tool version image history

okolesn:
[Compose a Dockerfile based on tool version image history #3568](https://github.com/epam/cloud-pipeline/issues/3568)


api/src/test/java/com/epam/pipeline/manager/docker/DockerParsingUtilsTest.java
```diff
@@ -48,4 +55,68 @@ public void shouldCalculateLatestAndEarliestDateTimeProperly() {
         assertThat(DockerParsingUtils.getLatestDate(description).toInstant().atZone(ZoneOffset.UTC).getMinute())
                 .isEqualTo(LATEST_MINUTES);
     }
+
+    @Test
+    public void testProcessCommands() {
```
SilinPavel, line 80: I also want to ask to check in this test that not only number of commands is right, but also that its structure was parsed correctly.
f.i.:
`|4 CELL_PROFILER_VERSION=4.2.1 CYTHON_VERSION=0.29.30 NUMPY_VERSION=1.23.1 PYTHON_3_DISTRIBUTION_URL=https://www.python.org/ftp/python/3.8.13/Python-3.8.13.tgz /bin/sh -c "..."` -> `RUN /bin/sh -c "..."`

and

`"set -o pipefail; command -v wget >/dev/null 2>&1 && { LAUNCH_CMD=\"wget --no-check-certificate -q -O - 'var1'\"; }; command -v curl >/dev/null 2>&1 && { LAUNCH_CMD=\"curl -s -k 'var2'\"; }; eval var3 | bash /dev/stdin \"var4\" 'var5' 'var6'"` was filtered out

etc.

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -93,6 +106,62 @@ public static Map<String, String> getLabels(final RawImageDescription rawImage)
...
+                entrypointIndexes.get(entrypointIndexes.size() - 1);
+
+        final int startIndex = commands.get(0).startsWith(ADD_FILE) ? 1 : 0;
+
+        for (int i = startIndex; i < commands.size(); i++) {
+            String command = commands.get(i);
+            if (command.startsWith(ARG)) {
+                result.add(command);
+            } else if (command.startsWith(ADD_FILE) || command.startsWith("ADD multi:")) {
+                result.add("ADD <source-file>");
```
SilinPavel: we should replace only `file:...` and `multi:...` but the destination should remain in the command:
`ADD file:12312l1l23k12l3lk /usr/bin/app` -> `ADD <source-file> /usr/bin/app`

SilinPavel: The same for COPY

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -93,6 +106,62 @@ public static Map<String, String> getLabels(final RawImageDescription rawImage)
...
+    public static List<String> processCommands(final String from, final List<String> commands,
+                                               final List<OSSpecificLaunchCommandTemplate> podLaunchTemplatesLinux,
+                                               final String podLaunchTemplatesWin) {
+        final List<String> result = new ArrayList<>();
+        result.add(String.format(FROM_TEMPLATE, from));
+        if (CollectionUtils.isEmpty(commands)) {
+            return result;
+        }
+        final List<String> podLaunchPatterns = getPodLaunchPatterns(podLaunchTemplatesLinux, podLaunchTemplatesWin);
+        final List<Integer> cmdIndexes = new ArrayList<>();
```
SilinPavel: you can store here not the whole list of indices but only the last one, and then just use it

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -93,6 +106,62 @@ public static Map<String, String> getLabels(final RawImageDescription rawImage)
...
+        final List<Integer> entrypointIndexes = new ArrayList<>();
+        final List<String> args = new ArrayList<>();
+        for (int i = 0; i < commands.size(); i++) {
+            String command = commands.get(i);
+            if (command.startsWith(ARG)) {
+                args.add(command.replace(ARG, ""));
+            } else if (command.startsWith(CMD)) {
+                cmdIndexes.add(i);
```
SilinPavel: instead of adding to list additional value, just save it as the last one and the use it

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -141,6 +210,30 @@ private static <T> T getMinElement(Stream<T> stream, Comparator<T> comparator) {
...
+        return input;
+    }
+
+    private static List<String> getPodLaunchPatterns(final List<OSSpecificLaunchCommandTemplate>
+                                                             podLaunchTemplatesLinux,
+                                                     final String podLaunchTemplatesWin) {
+        final List<String> result = podLaunchTemplatesLinux.stream()
+                .map(r -> getLaunchPodPattern(r.getCommand()))
+                .collect(Collectors.toList());
+        result.add(getLaunchPodPattern(podLaunchTemplatesWin));
```
SilinPavel: we have nice util method that can be reused here `StreamUtils.appended()` like:
```
StreamUtils.appended(
    podLaunchTemplatesLinux.stream().map(r -> getLaunchPodPattern(r.getCommand())),
    getLaunchPodPattern(podLaunchTemplatesWin)
)
```

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -93,6 +106,62 @@ public static Map<String, String> getLabels(final RawImageDescription rawImage)
...
+
+        for (int i = startIndex; i < commands.size(); i++) {
+            String command = commands.get(i);
+            if (command.startsWith(ARG)) {
+                result.add(command);
+            } else if (command.startsWith(ADD_FILE) || command.startsWith("ADD multi:")) {
+                result.add("ADD <source-file>");
+            } else if (command.startsWith(COPY_FILE)) {
+                result.add("COPY <source-file>");
+            } else if (command.startsWith(CMD)) {
```
SilinPavel, line 133: for the CMD and ENTRYPOINT I propose to insert it to the end of the file if they are present

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -93,6 +106,62 @@ public static Map<String, String> getLabels(final RawImageDescription rawImage)
...
+                cmdIndexes.add(i);
+            } else if (command.startsWith(ENTRYPOINT)) {
+                entrypointIndexes.add(i);
+            }
+        }
+        final int lastCmdIndex = CollectionUtils.isEmpty(cmdIndexes) ? -1 : cmdIndexes.get(cmdIndexes.size() - 1);
+        final int lastEntrypointIndex = CollectionUtils.isEmpty(entrypointIndexes) ? -1 :
+                entrypointIndexes.get(entrypointIndexes.size() - 1);
+
+        final int startIndex = commands.get(0).startsWith(ADD_FILE) ? 1 : 0;
```
SilinPavel: I would move this line closer to [FROM](https://github.com/epam/cloud-pipeline/pull/3572/files#diff-92af65798e97542bc9df5de528071ff150788a2b937522d289cdad576cf881a7R113) and leave a comment why we are doing that

api/src/main/java/com/epam/pipeline/manager/docker/DockerParsingUtils.java
```diff
@@ -93,6 +106,62 @@ public static Map<String, String> getLabels(final RawImageDescription rawImage)
...
+                result.add("COPY <source-file>");
+            } else if (command.startsWith(CMD)) {
+                if (i == lastCmdIndex) {
+                    result.add(command);
+                }
+            } else if (command.startsWith(ENTRYPOINT)) {
+                if (i == lastEntrypointIndex) {
+                    result.add(command);
+                }
+            } else if (COMMANDS.stream().noneMatch(command::startsWith)
```
SilinPavel: It seems that we can lost here some commands which are a valid docker commands and should be added as is, no?

SilinPavel (changes requested):

Issue #3573, opened 2024-06-26
Introduce hard/soft limits for commiting run and run tool/pipeline


SilinPavel:
**Background**
Docker images can be extremely big, so it is nice to have a mechanism to warn/reject user from creation or running such big images.

**Approach**
1. Let's introduce 2 new `System preference` - `commit.container.size.limits`, `launch.tool.size.limits` (Object) with structure:
    ```
    {
      "soft": <size value in bytes, long>,
      "hard": <size value in bytes, long>
    }
    ```
    > Provide validation for the values of the properties`

2. Modify a method in `PipelineRunController`
    GET `/run/{runId}/commit/check` to provide more extended result like:
    ```
    {
        "enoughSpace": {
           "status": <ENUM (FAIL, WARN, OK)>,
           "message": <can be null, message in case of warning or fail>,
           "value": <raw value of the metric, for enoughSpace true/false>
        },
        "containerSize":  {
            "status": <ENUM (FAIL, WARN, OK)>,
            "message": <can be null, message in case of warning or fail>,
            "value": <raw value of the metric, for containerSize size in bytes>
        }
    }
    ```
    To do so, let's introduce new server method in `DockerContainerOperationManager` to be able to ask node which is host the run, current size of the run container.
     - This method should work similar as commit procedure (ssh to the node) and utilize command `docker inspect ${CONTAINER_ID} -s --format "{{.SizeRw}}"`
     - Timeout should be implemented for the method to not to hang much time
     - If something goes wrong (timeout or error) return `-1` as `containerSize`

3. On UI:
    - During commit change logic to verify not only  `enoughSpace` but also check that `containerSize` check status
      - If WARN: 
        - Warn user that container image will be too big, but let to commit if needed
      - If FAIL 
        - Restrict to commit
   - During Tool/Pipeline run, check if tool version size < system preference `launch.tool.size.limits` `soft/hard`
      - If bigger: 
        - For soft : Warn user that container image is too big, but let to run if needed
        - For hard : Restrict to run

NShaforostov:
Docs added via https://github.com/epam/cloud-pipeline/pull/3669 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/10_Manage_Tools/10.4._Edit_a_Tool.md#container-size-limits) and [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/10_Manage_Tools/10.5._Launch_a_Tool.md#restrictions).

PR #3587, opened 2024-07-01
Issue 3573 - Introduce hard/soft limits for commiting run and run tool/pipeline

okolesn:
[Introduce hard/soft limits for commiting run and run tool/pipeline #3573](https://github.com/epam/cloud-pipeline/issues/3573)


api/src/main/java/com/epam/pipeline/manager/docker/DockerContainerOperationManager.java
```diff
@@ -267,6 +270,38 @@ public long getContainerLayers(final PipelineRun run) {
...
     }
 
+    public long getContainerSize(final PipelineRun run) {
+        final String containerId = kubernetesManager.getContainerIdFromKubernetesPod(run.getPodId(),
+                run.getActualDockerImage());
+        final long size;
+        try {
+            Assert.notNull(containerId,
+                    messageHelper.getMessage(MessageConstants.ERROR_CONTAINER_ID_FOR_RUN_NOT_FOUND, run.getId()));
+            final String getSizeCommand = DockerContainerLayersCommand.builder()
```
SilinPavel: Now, when we use this class in several situations, not only to get layers, I think it would be right to rename it. How about `SimpleDockerContainerCommand` since it basically can be any command that doesn't have other parameters except `containerId` and `scriptUrl`?

api/src/main/java/com/epam/pipeline/manager/docker/DockerContainerOperationManager.java
```diff
@@ -267,6 +270,38 @@ public long getContainerLayers(final PipelineRun run) {
...
+                    .runScriptUrl(getSizeCommandUrl)
+                    .containerId(containerId)
+                    .build()
+                    .getCommand();
+            Process sshConnection = submitCommandViaSSH(run.getInstance().getNodeIP(), getSizeCommand,
+                    getSshPort(run));
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(sshConnection.getInputStream()))) {
```
SilinPavel, line 291: I think we need to wait with `sshConnection.waitFor` then check for success before collecting an output, because this collecting is blocking operation and we can have a risk to block here forever

api/src/main/java/com/epam/pipeline/manager/docker/DockerContainerOperationManager.java
```diff
@@ -267,6 +270,38 @@ public long getContainerLayers(final PipelineRun run) {
...
+                final String result = reader.lines().collect(Collectors.joining());
+                size = Long.parseLong(result);
+            }
+            boolean isFinished = sshConnection.waitFor(
```
SilinPavel: can be final

api/src/main/java/com/epam/pipeline/manager/docker/DockerContainerOperationManager.java
```diff
@@ -267,6 +270,38 @@ public long getContainerLayers(final PipelineRun run) {
...
+        final long size;
+        try {
+            Assert.notNull(containerId,
+                    messageHelper.getMessage(MessageConstants.ERROR_CONTAINER_ID_FOR_RUN_NOT_FOUND, run.getId()));
+            final String getSizeCommand = DockerContainerLayersCommand.builder()
+                    .runScriptUrl(getSizeCommandUrl)
+                    .containerId(containerId)
+                    .build()
+                    .getCommand();
+            Process sshConnection = submitCommandViaSSH(run.getInstance().getNodeIP(), getSizeCommand,
```
SilinPavel: can be final

SilinPavel (changes requested):

Issue #3588, opened 2024-07-02
Implement way to persist current setup of the system and deploy a new environmet out of saved configuration


SilinPavel:
**Background**
Sometimes its can be very beneficial to be able to persist setup and all internal configuration (Application version, cp-global-config map, API system preferences, docker images) from existing environment.
Later on this saved configuration can be used to deploy (or re-deploy) identical environment. 

**Approach**

### New pipectl command scrap

We can expand `pipectl` and add new command `scrap` which will collect the next information about the environment:
 - Application version (filename: **app-version.json**). Basically, response of the GET `/app/info`:
   ```
    {
      "version": "0.17.0.16801.5baeeb6f6923291180ab27ac2c38c1e00b7066d8",
      "components": {
        "pipe-cli": "f90c9a9eb1fffeea1457f4c2a164db68cf6942c1",
        "data-transfer-service": "f9207173b27a5734666baf888a0b6fff5d589d17",
        "cloud-pipeline-webdav-client": "22737b274b2b7f9fc6004a109509ea117adfdadf"
      }
    }
   ```
 - System Preferences (filename: **system-preferences.json**). Response of the GET `/preferences`
   
   ```
    [
        {
          "name": "<>",
          "value": "<>",
          "preferenceGroup": "<>",
          "visible": true/false,
          "type": "STRING/.../..."
        },
        ...
    ]
   ```
 - Kubernetes `cp-config-global` configMap content in the format of env file (filename: **cp-config-global.properties**):
   ```
      KEY=value
      KEY1=value1
      ....
   ```
 - List of Cloud-pipeline services deployed (filename: **cp-services.json**)
   ```
   ["cp-api-srv", "cp-edge", ... ]
   ```
 - List of docker containers deployed in format of (directory: **dockers-manifest**):
   - manifest.txt 
     ```
     <registry-url>/library/centos:7,library/centos:7
     <registry-url>/my-group/ubuntu:20.04,my-group/ubuntu:20.04
     .....
     ```
   - library/
     - centos:7/
       - icon.png
       - README.md
       - spec.json
   - my-group/
     - ubuntu:20.04/
       - icon.png
       - README.md
       - spec.json  
 
 - List of users (filename: **users.json**):
   ```
   [
     {
        "id": 1,
        "userName": "<>",
        "roles": [
          {
            "id": 1,
            "name": "ROLE_ADMIN",
            "predefined": true,
            "userDefault": false,
            "defaultProfileId": 1
          },
          ...
[...]

PR #3589, opened 2024-07-03
Added scrap scripts

oleksii-mart:
#3588


deploy/contents/sync/scrap/get_configmap.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+get_configmap() {
+  local ssh_key="$1" #SSH Key for connection
+  local user="$2" #User whom connection  
+  local node_ip="$3"  #Server address name 
+  local namespace="$4" #Kubernetes namespace 
+  local configmap_name="$5" #Name of the dowloaded configmap
+  local output_dir="$6" #Output directory name where config-map will be saved
```
SilinPavel: let's combine last two parameters into one

deploy/contents/sync/scrap/get_configmap.sh
```diff
@@ -0,0 +1,18 @@
...
+  local output_file="cp-config-global.yaml" #Output file name where config-map will be saved
+    
+#SSH connection is done within this command
+    ssh -i $ssh_key -oStrictHostKeyChecking=no $user@$node_ip sudo kubectl get configmap $configmap_name -n $namespace -o yaml > $output_dir/$output_file
```
SilinPavel: we need to check that this command was successful and only if so - wite output to the file

SilinPavel: and if no - exit with some bad code to handle it in the main script

deploy/contents/sync/scrap/get_configmap.sh
SilinPavel: also check all other scripts for the same

deploy/contents/sync/scrap/get_configmap.sh
SilinPavel: also check all other scripts for the same

deploy/contents/sync/scrap/get_services.sh
```diff
@@ -0,0 +1,19 @@
...
+
+get_services() {
+    local ssh_key="$1" # SSH key
+    local user="$2" # USER
+    local server_address="$3" # Server address
+    local output_dir="$4" #Output directory name where preferences will be saved
+    local output_file="cp-services" #Output file name where services will be saved
+
+    #SSH connection is done within this command
+    ssh -i $ssh_key -oStrictHostKeyChecking=no $user@$server_address sudo kubectl get nodes -o json | jq -r '.items[] | .metadata.labels | to_entries[] | select(.key | startswith("cloud-pipeline/")) | .key | sub("cloud-pipeline/";"") | select(startswith("cp-"))' > $output_dir/$output_file
```
SilinPavel: let's split this operation:
 - First, get the output from ssh
 - Second, check that command was successful
 - Finally, do the post processing of the data

deploy/contents/sync/scrap/get_version.sh
```diff
@@ -0,0 +1,25 @@
...
+
+    if [ -z "$server_address" ]; then
+        echo "Please provide the server address"
+        exit 1
+    fi
+
+    
+
+    # Send GET request to the server
+    curl -H "Authorization: Bearer ${token}" "https://${server_address}/${api_endpoint}" -H "Accept: application/json" | jq -r '.payload' > $output_dir/$output_file
```
SilinPavel: let's split this operation:

  - First, get the output from ssh
  - Second, check that command was successful
  - Finally, do the post processing of the data

deploy/contents/sync/scrap/scrap.sh
```diff
@@ -0,0 +1,130 @@
...
+        ;;
+esac        
+done
+
+if [ -z "$ssh_key" ]; then
+    echo "Please provide the ssh key"
+    exit 1
+fi
+
+if [ -z "$server_address" ]; then
```
SilinPavel: in many scripts we have our well defined `API_URL` and `API_TOKEN` env vars, lets use it here as a second thing to check:
 - if it present - use it as a value
 - if not, fail as we do now

deploy/contents/sync/scrap/scrap.sh
```diff
@@ -0,0 +1,130 @@
...
+if [ -z "$configmap_name" ]; then
+   configmap_name="cp-config-global"
+fi
+
+if [ -z "$namespace" ]; then
+   namespace="default"
+fi
+
+if [ ! -d "$output_dir" ]; then
+    echo "Checking if output directory exist. Directory does not exist. Creating directory"
```
SilinPavel: it is good idea also check if directory is empty:
 - if empty - all good
 - if not - lets introduce new flag `-f/--force` if it was provided - script can write to the directory with some content, if not - fail with a massage like `Output directory isn't empty, pipectl will not continue. To override this behaviour specify flag --force`

deploy/contents/sync/scrap/scrap.sh
```diff
@@ -0,0 +1,130 @@
...
+    echo "Directory $output_dir created."
+else
+    echo "Checking if output directory exist. Directory exists"
+fi
+
+echo "Getting configmap from server..."
+get_configmap $ssh_key $user $node_ip $namespace $configmap_name $output_dir
+
+
+if [ $? -ne 0 ]; then
```
SilinPavel: let's introduce another file `revision_metadata.csv` which will contains several lines in csv format. Each line represent state of the collected information like `application version` or `preferences`, etc

```
application_version,0
preferences,1
```
where first column - name of the piece of the information, second column result code `0' - all good, `!= 0` - data can be corrupted

deploy/contents/sync/scrap/scrap.sh
```diff
@@ -0,0 +1,130 @@
...
+    echo "Error occured while downloading configmap file" 
+     exit 1
```
SilinPavel: let's not fail here, instead populate `revision_metadata.csv` with appropriate data and proceed

SilinPavel (changes requested):

Issue #3596, opened 2024-07-08
Improve pipectl install command to be able to deploy Cloud-Pipeline out of point-in-time configuration


SilinPavel:
**Approach**
Based on #3588 results let's implement a way to specify such directory as an input for `pipectl install` command.
 - If such option (-pc/--point-in-time-configuration) was provided, `install` command should be launched with:
   - List of service from `cp-serices.json` enabled
   - `cp-config-global.properties` should be used as `install-config`
   - `dockers-manifest` from point-in-time configuration should be used instead of inner  `dockers-manifest` from pipectl binary
     - Bear in mind that `dockers-manifest` from point-in-time configuration has slightly different structure (additional level in directory structure due to docker-registry folder) So, code should be adjusted to be able to work with both versions of  `dockers-manifest` folder
   - `system-preferences.json` should be used on the final stage of the installation. `install` process should iterate over this list and update corresponding items with these values.

SilinPavel:
Let's introduce additional option (should take effect only when -pc/--point-in-time-configuration enabled)
`-pcm/--point-in-time-configuration-modules` (example of  possible value: `services,tools`)
depending on this setting we will define which modules we should grab and apply from `pitc`:
 - if `-pcm` isn't present - apply all all modules
 - else - check that module is enabled in this setting

SilinPavel:
Let's also change approach to save registered users:
Instead of use `GET /users` let's use `POST /user/export` with all flags enabled

Then let's use this `csv` file in pipectl install from pitc and utilize `POST /users/import` method to register this list of users, if configured

PR #3603, opened 2024-07-11
Issue 3596: Pipectl install cp-services from point-in-time configuration

oleksii-mart:
Install cp-services for #3596

### DIFF
--- a/deploy/contents/install/app/install-utils.sh
+++ b/deploy/contents/install/app/install-utils.sh
@@ -591,7 +591,7 @@ function if_data_is_available_in_point_in_time_configuration {
   return 1
 }
 
-function install_services_from_recovery_file {   
+function enable_services_from_point_in_time_configuration {   
     local service_point_in_time_configuration=$(if_data_is_available_in_point_in_time_configuration $CP_POINT_IN_TIME_CONFIGURATION_DIR services) 
     if [ $? -eq 0 ]; then
        while read -r key; do 
--- a/deploy/contents/install/app/install-utils.sh
+++ b/deploy/contents/install/app/install-utils.sh
@@ -562,7 +562,7 @@ function array_contains_or_empty () {
     return 1
 }
 
-function count_services {
+function enable_services {
     parse_env_option "$1"
     services_count=$((services_count+1))
     if [ -n "${CP_SERVICES_ENABLED}" ]; then 
@@ -570,32 +570,36 @@ function count_services {
     else 
         export CP_SERVICES_ENABLED="${1}"
     fi
-    }
+}
 
-function export_service_all {
-        export CP_INSTALL_SERVICES_ALL=1
-        export CP_SERVICES_ENABLED="all"
-    }
+function enable_all_services {
+    export CP_INSTALL_SERVICES_ALL=1
+    export CP_SERVICES_ENABLED="all"
+}
 
-function services_from_pitc {
-    local services_list="$CP_POINT_IN_TIME_CONFIGURATION_DIR/cp-services.json"
-    if [ ! -s "$services_list" ]; then
-       export_service_all
-    else
-       while IFS=',' read -r app switch; do
-        if [ $app = "app_services" ]; then
-            if [ $switch -eq 0 ]; then
-                while read -r key; do 
-                    count_services "$key"
-                    echo "$key=${!key}"
-                done < <(cat "$services_list" | jq -r '.[]')
-            else 
-                export_service_all       
-            fi
-        fi
-        done < "$CP_POINT_IN_TIME_CONFIGURATION_DIR/_pitc.csv"
+function check_data_to_scrap_file {
+  local csv_file="${1}/_pitc.csv"
+  local data_to_scrap="$2"
+
+  while IFS="," read -r data data_file exit_code; do
+  if [ "$data_to_scrap" == "$data" ] && [ -s "$data_file" ] && [ $exit_code -eq 0 ]; then
+    export json_file=$data_file
+    return 0
+  fi   
+    done < "$csv_file"
+    return 1
+}
+
+function install_services_from_recovery_file {   
+    if check_data_to_scrap_file $CP_POINT_IN_TIME_CONFIGURATION_DIR services; then
+       while read -r key; do 
+       enable_services "$key"
+       echo "$key=${!key}"
+       done < <(cat "$json_file" | jq -r '.[]')  
+    else  
+       enable_all_services
     fi
-}    
+}  
 
 function parse_options {
     local services_count=0
@@ -643,7 +647,7 @@ function parse_options {
         shift # past argument
         ;;
         -s|--service)
-        count_services "$2"
+        enable_services "$2"
         shift # past argument
         shift # past value
         ;;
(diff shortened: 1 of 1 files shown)


deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
...
 
+function count_services {
+    parse_env_option "$1"
+    services_count=$((services_count+1))
+    if [ -n "${CP_SERVICES_ENABLED}" ]; then 
+        export CP_SERVICES_ENABLED="${CP_SERVICES_ENABLED},${1}"
+    else 
+        export CP_SERVICES_ENABLED="${1}"
+    fi
+    }
```
SilinPavel: formatting

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
     return 1
 }
 
+function count_services {
```
SilinPavel: why you decide to name it this way? what is the main purpose of this function? For me it more about `enable` service than `count`

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
...
+    parse_env_option "$1"
+    services_count=$((services_count+1))
+    if [ -n "${CP_SERVICES_ENABLED}" ]; then 
+        export CP_SERVICES_ENABLED="${CP_SERVICES_ENABLED},${1}"
+    else 
+        export CP_SERVICES_ENABLED="${1}"
+    fi
+    }
+
+function export_service_all {
```
SilinPavel: again, what does it mean `export` service, why not `enable_all_services`

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
...
+        export CP_INSTALL_SERVICES_ALL=1
+        export CP_SERVICES_ENABLED="all"
+    }
+
+function services_from_pitc {
```
SilinPavel: it is better if name of the function is an action (because function is actually a some action)
f.i. this one will `enable` services for the installation from point-in-time configuration, right? so let's name it appropriate

SilinPavel: Also it is better to not to use abbreviations (`pitc`) in the name if the function if such abbreviation is not `well-known`

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
...
+    local services_list="$CP_POINT_IN_TIME_CONFIGURATION_DIR/cp-services.json"
+    if [ ! -s "$services_list" ]; then
+       export_service_all
```
SilinPavel: do we really need this here? 
I believe that we will enable all services if needed https://github.com/epam/cloud-pipeline/pull/3603/files#diff-5c368da199a8aa904a21120fa49deab5adb5dfe6f1f66389a61ec978407c7855R788-R790 here anyway

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
...
+    else
+       while IFS=',' read -r app switch; do
+        if [ $app = "app_services" ]; then
+            if [ $switch -eq 0 ]; then
+                while read -r key; do 
+                    count_services "$key"
+                    echo "$key=${!key}"
+                done < <(cat "$services_list" | jq -r '.[]')
+            else 
+                export_service_all
```
SilinPavel: do we really need this here? 
I believe that we will enable all services if needed https://github.com/epam/cloud-pipeline/pull/3603/files#diff-5c368da199a8aa904a21120fa49deab5adb5dfe6f1f66389a61ec978407c7855R788-R790 here anyway

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,41 @@ function array_contains_or_empty () {
...
+        export CP_INSTALL_SERVICES_ALL=1
+        export CP_SERVICES_ENABLED="all"
+    }
+
+function services_from_pitc {
+    local services_list="$CP_POINT_IN_TIME_CONFIGURATION_DIR/cp-services.json"
+    if [ ! -s "$services_list" ]; then
+       export_service_all
+    else
+       while IFS=',' read -r app switch; do
```
SilinPavel: I propose to move code that check that we have all necessary files for this step to a different function. Then here we can use this function.
It would be even better and natural to make such function common to be able to check presents of any configuration in `pitc` folder, based on passed parameter

SilinPavel (changes requested):

deploy/contents/install/app/install-utils.sh
oleksii-mart: So if file not exist or empty be better to export services_count=0 to enable all services? if allready we have checking of the services_count value

deploy/contents/install/app/install-utils.sh
```diff
@@ -747,10 +785,13 @@ function parse_options {
                             "$CP_EDGE_EXTERNAL_PORT"
     fi
 
+    if [ $CP_POINT_IN_TIME_CONFIGURATION_DIR ]; then
+        install_services_from_recovery_file
```
SilinPavel: install_services_from_recovery_file -> enable_services_from_point_in_time_configuration

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,45 @@ function array_contains_or_empty () {
     return 1
 }
 
+function enable_services {
```
SilinPavel: enable_services -> enable_service

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,45 @@ function array_contains_or_empty () {
...
+    done < "$csv_file"
+    return 1
+}
+
+function install_services_from_recovery_file {   
+    if check_data_to_scrap_file $CP_POINT_IN_TIME_CONFIGURATION_DIR services; then
+       while read -r key; do 
+       enable_services "$key"
+       echo "$key=${!key}"
+       done < <(cat "$json_file" | jq -r '.[]')
```
SilinPavel: json_file should be a parameter of this function

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,45 @@ function array_contains_or_empty () {
...
+        export CP_SERVICES_ENABLED="${1}"
+    fi
+}
+
+function enable_all_services {
+    export CP_INSTALL_SERVICES_ALL=1
+    export CP_SERVICES_ENABLED="all"
+}
+
+function check_data_to_scrap_file {
```
SilinPavel: check_data_to_scrap_file -> if_data_is_available_in_point_in_time_configuration

deploy/contents/install/app/install-utils.sh
```diff
@@ -562,6 +562,45 @@ function array_contains_or_empty () {
...
+  local csv_file="${1}/_pitc.csv"
```
SilinPavel: let's split this into 2 vars
```
local point_in_time_configuration_dir="${1}"
local pitc_metadata_file=$point_in_time_configuration_dir/_pitc.csv
```

SilinPavel (changes requested):

PR #3618, opened 2024-07-26
Issue 3596: Install system-preferences from poin-in-time-configuration

oleksii-mart:
Related to https://github.com/epam/cloud-pipeline/issues/3596 added function to set up system-preferences from point-in-time-configuration file.

### DIFF
--- a/deploy/contents/install/app/install-utils.sh
+++ b/deploy/contents/install/app/install-utils.sh
@@ -605,6 +605,22 @@ function enable_services_from_point_in_time_configuration {
     fi
 }
 
+function set_preferences_from_point_in_time_configuration {
+  local point_in_time_configuration_preference_file=$(is_data_is_available_in_point_in_time_configuration system_preference)
+  
+  if [ $point_in_time_configuration_preference_file ]; then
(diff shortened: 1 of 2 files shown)


deploy/contents/install/app/install-utils.sh
```diff
@@ -605,6 +605,22 @@ function enable_services_from_point_in_time_configuration {
...
+  if [ $point_in_time_configuration_preference_file ]; then
+
+     local preferences_file_with_values=$(cat "$point_in_time_configuration_preference_file" | jq -c '[.[] | select(has("value"))]')
+     
+     payload=$(echo "$preferences_file_with_values" | jq -r '.[] | "\(.name) \(.value) \(.visible)"')
+     while IFS= read -r pref_name pref_value pref_visible; do
+       api_set_preference $pref_name $pref_value $pref_visible
+     done <<< "$payload"
+  else
+     exit 1
```
SilinPavel: should not exit with 1 but just finish function call, it is not a bad situation for us, rather just a case when we won't apply preferences

Issue #3622, opened 2024-07-30
Allow to limit 'run_as' feature


mzueva:
**Background**
At the moment when a user is allowed to use `run as` feature with some service account, the user can use it for any available pipeline or tool. This may lead to some security problems, especially if the service account has wide permissions. It would be helpful to allow to limit to which pipelines and tools `run as` feature can be used.

**Approach**
When a user is granted `run as` permission to another account, we can specify only access type (SSH, Endpoint). The following options shall be added:
- `pipelinesAllowed` (boolean) - whether user can launch any pipeline with `run as` feature. If this value is not set (null), user can launch pipelines to preserve existing behaviour.
- `pipelines` (list of long ids) - if specified limits pipelines to be launched by the user with `run as` feature
- `toolsAllowed` (boolean) - whether user can launch tools with `run as` feature. If this value is not set (null), user can launch tools to preserve existing behaviour.
- `tools` (list of long ids) - if specified limits tools to be launched by the user with `run as` feature

**_Note:_** For pipelines `run as` availability shall be checked actually for both: for pipeline itself and the tool

PR #3624, opened 2024-07-31
Allow to limit `run as` for specific tools and pipelines

mzueva:
Provides implementation for #3622 
- New fields added to `RunnerSid` entity, db scheme update
- Added tools/pipelines check when tool or pipeline is launched using `run as` feature

**_Note:_** Lists of allowed pipelines and tools are stored in DB as a comma-delimited list, as converting `@Embeddable` object to an `@Entity` would require more changes and potentially cause backwards compatibility issues.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunAsManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunAsManager.java
@@ -162,19 +162,27 @@ public class PipelineRunAsManager {
                 MessageConstants.ERROR_RUN_ALLOWED_SID_NOT_FOUND, runAsUser));
         final Long pipelineId = runVO.getPipelineId();
         if (Objects.nonNull(pipelineId)) {
-            Assert.isTrue(isNullOrTrue(allowedRunnerSid.getPipelinesAllowed()),
-                    messageHelper.getMessage(
-                            MessageConstants.ERROR_RUN_AS_PIPELINES_NOT_ALLOWED, currentUser, runAsUser));
-            if (CollectionUtils.isNotEmpty(allowedRunnerSid.getPipelines())) {
-                Assert.isTrue(allowedRunnerSid.getPipelines().contains(pipelineId),
-                        messageHelper.getMessage(MessageConstants.ERROR_RUN_AS_PIPELINE_NOT_ALLOWED,
-                                currentUser, pipelineId, runAsUser));
-            }
-            final PipelineConfiguration currentUserConfiguration = configurationManager.getPipelineConfiguration(runVO);
-            validateTool(currentUserConfiguration.getDockerImage(), runAsUser, allowedRunnerSid, currentUser);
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/epam/pipeline/entity/user/RunnerSid.java
```diff
@@ -17,27 +17,75 @@
...
     private boolean principal = true;
 
     @Enumerated(EnumType.STRING)
     private RunAccessType accessType = RunAccessType.ENDPOINT;
+    private Boolean pipelinesAllowed;
+    private Boolean toolsAllowed;
+    @JsonIgnore
+    private String pipelinesList;
+    @JsonIgnore
+    private String toolsList;
```
tcibinan, line 54: As long as we return instances of this class in `UserController::getRunners`, we shall remove `@JsonIgnore` annotations here. Otherwise, a client won't be able to read runners details once they are saved.

Other option would be to return `RunnerSidVO` instances in `UserController::getRunners`. In this case both structures would be the same.

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunAsManager.java
```diff
@@ -146,9 +150,45 @@ private PipelineRun runWithUserContext(final String runAsUserName, final Callabl
...
+        final Long pipelineId = runVO.getPipelineId();
+        if (Objects.nonNull(pipelineId)) {
+            Assert.isTrue(isNullOrTrue(allowedRunnerSid.getPipelinesAllowed()),
+                    messageHelper.getMessage(
+                            MessageConstants.ERROR_RUN_AS_PIPELINES_NOT_ALLOWED, currentUser, runAsUser));
+            if (CollectionUtils.isNotEmpty(allowedRunnerSid.getPipelines())) {
+                Assert.isTrue(allowedRunnerSid.getPipelines().contains(pipelineId),
+                        messageHelper.getMessage(MessageConstants.ERROR_RUN_AS_PIPELINE_NOT_ALLOWED,
+                                currentUser, pipelineId, runAsUser));
+            }
```
tcibinan: I would also extract `validatePipeline` method here just to be consistent with `validateTool`.

core/src/main/java/com/epam/pipeline/entity/user/RunnerSid.java
mzueva: Client will receive parsed ids information via getters `getPipelines` and `getTools()` defined in this class. In order to switch completely from entity `RunnerSid` to `RunnerSidVO` completely we also need to add changes to `PipelineUser` class, which I prefer to avoid.

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunAsManager.java
mzueva: Sure, extracted method

tcibinan (approved):

Issue #3619, opened 2024-07-29
Migrate runs to archive table according to user/group configuration


ekazachkova:
**Background**
Cloud Pipeline does not allow remove runs and it would be nice to support it because a huge runs count can affect queries performance.

`billing service` may require runs data for whole period. Therefore, API shall be able to return archived runs with statuses.  

**Approach**

### 1. API

- add new metadata field `run_archive_days` for users or groups. The name of this field shall be managed via `system.archive.run.metadata.key` preference. This value indicates the number of days to archive runs - all runs that finished more that `N` days ago shall be archived for user or group entity that provides this metadata value. 
- create a new DB table `run_archive` with the same structure as `pipeline_run`
- create a new DB table `archive_run_status_change` with the same structure as `run_status_change`
- billing service shall be able to load archived data: support `archive` flag for `GET run/activity` API method. If enabled runs data from `run_archive` and `archive_run_status_change` tables shall be loaded as well for requested period. Default: off.
- add a new API method to support runs migration `POST /runs/archive`. Monitoring service shall call this method.

This operation available for admin users only. 

Steps:
 - collect metadata with `run_archive_days` key
 - extract owners and days from collected metadata
 - collect days for owners properly: 
   - if user has no archive run metadata and belongs to a multiple groups that have archive runs metadata the minimal value of days count shall be used. 
   - If user has an archive run metadata - this value  shall be used regardless of whether the value in groups metadata
 - load all master runs in terminal state where owner and end_date columns match the restriction from metadata.
 - load child runs
 - create archive runs
 - remove all dependents (see list below)
 - remove runs by ids

DB table dependents:
`run_user`
`pipeline_run_log`
`restart_run`
`pipeline_run_service_url`
 `run_status_change`
`stop_serverless_run`

- add a new API method `POST /runs/archive/owners` that migrates runs for specified user or group. Request parameters:
  - `ownerSid` - user name/id or group name/id
  - `principal` - true for users
  - `days` - optional, if not specified the number of days shall be loaded from the metadata. If no corresponding metadata is specified for the requested user or group, an error shall occur. If days parameter specified this value shall be used to process runs regardless of whether the value is present in the metadata.

This operation shall be available for admin users only. 

NOTES: 
1. archive runs operations shall be asynchronous. Introduce a new executor `BackgroundJobs`. The executor's pool size shall be changed via `background.api.jobs.pool.size=${CP_API_BACKGROUND_JOBS_POOL_SIZE:10}` application property.
[...]

PR #3647, opened 2024-08-22
Migrate runs to archive table according to user/group configuration - dry-run

ekazachkova:
relates #3619

Provides ability to enable `dry-run` regime using `system.archive.run.dry-run.regime` system property (default: off). 

- all load queries shall be logged
- all update queries shall be logged (excluding batch queries) and omitted

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
@@ -22,7 +22,7 @@ import com.epam.pipeline.controller.vo.PipelineRunFilterVO;
 import com.epam.pipeline.controller.vo.run.RunChartFilterVO;
 import com.epam.pipeline.dao.DaoHelper;
 import com.epam.pipeline.dao.DaoUtils;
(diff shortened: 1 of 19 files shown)


api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
```diff
@@ -93,6 +96,9 @@ public class PipelineRunDao extends NamedParameterJdbcDaoSupport {
     @Autowired
     private RunServiceUrlDao serviceUrlDao;
 
+    @Autowired
```
SilinPavel: would it be possible to inherit from this class instead of have it as dep? And then, you can move `1686-1689` lines to `JdbcTemplateReadOnlyWrapper`. it should slightly reduce number of copy-paste lines in a future, when we will reuse this approach for other DAOs, if needed.

P.S. with this approach it would be better to rename `JdbcTemplateReadOnlyWrapper` as `JdbcTemplateWithDryRunWrapper`

SilinPavel (changes requested):

Issue #3659, opened 2024-08-27
Create personal docker registry group for the user on user creation


SilinPavel:
Let's add possibility to configure creation of the private docker registry group on user creation

Two new system prefs:
 - `system.docker.registry.create.user.group` boolean: enables or disables functionality
 - `system.default.docker.registry.id` long: specifies where user docker group should be created

For functionality to work both `sys prefs` should be fulfilled.
If on user creation docker registry group with the same name already exists, user still be created without any error.

PR #3663, opened 2024-08-29
Issue 3659 Create personal docker registry group for user on creation

SilinPavel:
Is related to #3659

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/user/UserManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/user/UserManager.java
@@ -87,7 +87,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/epam/pipeline/manager/user/UserManager.java
```diff
@@ -146,12 +152,33 @@ public class UserManager implements SecuredEntityManager {
...
-                               List<String> groups, Map<String, String> attributes,
-                               Long defaultStorageId) {
-        PipelineUser user = create(name, roles, groups, attributes);
-        user = configureUserDefaultStorage(user, defaultStorageId);
-        return configureUserDefaultMetadata(user);
+    public PipelineUser create(final String name, final List<Long> roles,
+                               final List<String> groups,
+                               final Map<String, String> attributes,
+                               final Long defaultStorageId) {
+        return Stream.of(create(name, roles, groups, attributes))
```
ekazachkova: It seems that using `Optional` would be more appropriate here

ekazachkova (approved):

Issue #3638, opened 2024-08-16
pipe storage du: Support --depth when --cloud disabled


ekazachkova:
**Background**
At the moment, Cloud Pipeline allows to fetch data  for `pipe storage du` with `--depth` option from cloud provider only and it would be nice to support `--depth` with disabled `--cloud` flag. 

**Approach**
- extract folder names directly from cloud provider given specified depth
- for each folder path load stat data from API

Notes:
- for `--output-mode=brief` let's flush data as soon as it is received (not supported for `--output-mode=all` since all item's storage classes shall loaded to construct table header properly)

Also, it would be nice to support `--depth` for `NFS` as part of this issue.

PR #3668, opened 2024-08-30
pipe storage du: Support --depth when --cloud disabled (NFS support)

ekazachkova:
relates to #3638


pipe-cli/pipe.py
```diff
@@ -294,6 +294,8 @@ def cli():
       CP_TRACE=[True|False]                  Enables verbose errors.
       CP_CLI_STORAGE_BATCH_SIZE              The number of objects per request for pipe storage operations (Default: 1000)
       CP_CLI_STORAGE_ASYNC_BATCH_ENABLE      Enables asynchronous batch transfer
+      CP_CLI_STORAGE_LIST_API_PAGE_SIZE      The number of storage items allowed to be loaded from the API
```
SilinPavel, line 297: is there a case when in one `cli` command we can use `CP_CLI_STORAGE_BATCH_SIZE` and `CP_CLI_STORAGE_LIST_API_PAGE_SIZE` simultaneously? if no, I think we can reuse  `CP_CLI_STORAGE_BATCH_SIZE`

PR #3738, opened 2024-10-09
Fix cluster cost calculation

kamyshova:
This PR contains a fix cluster cost calculation if a run with child runs is not specified as master run.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/costs/ClusterCostsMonitoringServiceCore.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/costs/ClusterCostsMonitoringServiceCore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 EPAM Systems, Inc. (https://www.epam.com/)
+ * Copyright 2021-2024 EPAM Systems, Inc. (https://www.epam.com/)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ import java.math.RoundingMode;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/costs/ClusterCostsMonitoringServiceCore.java
```diff
@@ -28,29 +31,46 @@
...
-                .filter(PipelineRun::isMasterRun)
-                .collect(Collectors.toMap(PipelineRun::getId, Function.identity()));
+        final PagingRunFilterVO filter = new PagingRunFilterVO();
+        filter.setPageSize(DEFAULT_PAGE_SIZE);
+        filter.setStatuses(Collections.singletonList(TaskStatus.RUNNING));
+        filter.setEagerGrouping(false);
+        filter.setUserModified(false);
+        int page = DEFAULT_PAGE;
+        final Map<Long, PipelineRun> masters = new HashMap<>();
+        Integer totalCount = null;
```
SilinPavel: it seems that there is no need for such var, just use `listPagedResult.getTotalCount()` in line 71

api/src/main/java/com/epam/pipeline/manager/cluster/costs/ClusterCostsMonitoringServiceCore.java
```diff
@@ -28,29 +31,46 @@
...
+        boolean isMorePagesAvailable = true;
+        while (isMorePagesAvailable) {
```
SilinPavel: since we aiming to load all runs there is no need in paging I guess, we simply can specify page size = MAX_INT

SilinPavel (changes requested):

api/src/main/java/com/epam/pipeline/manager/cluster/costs/ClusterCostsMonitoringServiceCore.java
kamyshova: Paging was removed according to searchPipelineRuns update with checkMaxPageSize

api/src/main/java/com/epam/pipeline/manager/cluster/costs/ClusterCostsMonitoringServiceCore.java
kamyshova: Update searchPipelineRuns() with checkMaxPageSize to allow to specify page size = Integer.MAX_VALUE

Issue #3751, opened 2024-10-23
Allow to grant permissions to group


okolesn:
**Background**

At the moment only admin users and users with ROLE_READER assigned can view list of groups and group details. In addition to this platform shall allow to grant access to user using ACL approach, as it is done for other types of entities.

**Approach**

Access for admins and ROLE_READER users remain the same
Admins shall be allowed to grant access to group to other users/groups (standard RWE permissions)
Users shall be able to view list of groups they have READ access to
Users with READ permission are allowed to view group details
Users with WRITE permission are allowed to modify group

maryvictol:
Test cases were created by #3811 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/Role_Based_Access_Control/3751_allow_to_grant_permissions_to_group).

PR #3752, opened 2024-10-23
Issue 3751 Allow to grant permissions to group

okolesn:
[Allow to grant permissions to group #3751](https://github.com/epam/cloud-pipeline/issues/3751)


api/src/main/java/com/epam/pipeline/acl/user/RoleApiService.java
```diff
@@ -19,58 +19,72 @@
...
     @Autowired
     private RoleManager roleManager;
 
-    @PreAuthorize(ADMIN_ONLY + OR_USER_READER)
-    public Collection<Role> loadRolesWithUsers() {
-        return roleManager.loadAllRoles(true);
+    @PostFilter(USER_READ_FILTER)
+    @AclMaskList
+    public List<Role> loadRolesWithUsers() {
+        return new ArrayList<>(roleManager.loadAllRoles(true));
```
SilinPavel: instead of recreating list once again (in dao class we already returning a list) let's just change signature of the corresponding methods

api/src/main/java/com/epam/pipeline/acl/user/RoleApiService.java
```diff
@@ -19,58 +19,72 @@
...
     }
 
-    public Collection<Role> loadRoles() {
-        return roleManager.loadAllRoles(false);
+    @PostFilter(USER_READ_FILTER)
+    @AclMaskList
+    public List<Role> loadRoles() {
+        return new ArrayList<>(roleManager.loadAllRoles(false));
```
SilinPavel: same as previous comment

api/src/main/java/com/epam/pipeline/manager/metadata/MetadataManager.java
```diff
@@ -407,7 +403,7 @@ private void checkEntityExistsAndCanBeModified(final Long entityId, final AclCla
     private Object loadEntity(final Long entityId, final AclClass entityClass) {
         Object entity;
         if (entityClass.equals(AclClass.ROLE)) {
```
SilinPavel: since now Role is a `SecuredEntity` we can support same way to load it as for all other types of `SecuredEntity`

api/src/main/java/com/epam/pipeline/manager/user/RoleManager.java
```diff
@@ -176,8 +187,8 @@ public Optional<Role> findRoleByName(final String name) {
         return roleDao.loadRoleByName(name);
     }
 
-    public ExtendedRole loadRoleWithUsers(Long roleId) {
-        loadRole(roleId);
+    public ExtendedRole loadRoleWithUsers(final Long roleId) {
+        load(roleId);
```
SilinPavel: seems to me as an necessary load call, what you think?

SilinPavel (changes requested):

api/src/main/resources/dao/user-dao.xml
```diff
@@ -115,6 +116,7 @@
                       r.user_default as role_user_default,
                       r.default_storage_id as role_default_storage_id,
                       r.default_profile_id as role_default_profile_id,
+                      r.owner as role_owner
```
SilinPavel: there should be a `,`

Issue #3676, opened 2024-09-04
Provisioning FSx Lustre FS with user provisioned IOPS


SilinPavel:
Now Lustre FS supports user provisioned IOPS: https://docs.aws.amazon.com/fsx/latest/LustreGuide/performance.html#dne-metadata-performance
So, lets provide opportunity for user to customize IOPS when launch cluster with shared lustre FS:

 - New launch.system.preference `CP_CAP_SHARE_FS_IOPS` - Integer, to specify IOPS level according to AWS documentation

maryvictol:
**Test case** was created by #3755 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/Test_Environment/3676.md).

maryvictol:
**e2e-test** was implemented via https://github.com/epam/cloud-pipeline/pull/3755.

PR #3755, opened 2024-10-30
Add testcase and e2e test for issue 3676

maryvictol:
This PR add testcase and e2e test for issue Provisioning FSx Lustre FS with user provisioned IOPS #3676

### DIFF
--- a/e2e/gui/src/test/java/com/epam/pipeline/autotests/PlatformPreferencesTest.java
+++ b/e2e/gui/src/test/java/com/epam/pipeline/autotests/PlatformPreferencesTest.java
@@ -238,6 +238,52 @@ public class PlatformPreferencesTest extends AbstractSeveralPipelineRunningTest 
     }
 
     @Test
+    @TestCase(value = {"3676"})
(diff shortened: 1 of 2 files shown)


e2e/gui/src/test/java/com/epam/pipeline/autotests/PlatformPreferencesTest.java
```diff
@@ -237,6 +237,52 @@ public void allowToSpecifyLustreFSTypeAndThoughput() {
         }
     }
 
+    @Test
```
kamyshova, line 245: I suppose we should add @CloudProviderOnly(values = {Cloud.AWS}) annotation to the test

kamyshova (changes requested):

NShaforostov (approved):

e2e/gui/src/test/java/com/epam/pipeline/autotests/PlatformPreferencesTest.java
maryvictol: Done

kamyshova (approved):

Issue #3866, opened 2025-01-20
[NGS Portal] Add "project id" optional property for runs / run filters


rodichenko:
It would be useful to have a run's associated **project** identifier, if any; also, *filter runs* method should accept this property (an array of project identifiers) to filter runs based on the project(s).

PR #3886, opened 2025-01-31
Issue 3866 [NGS Portal] Add "project id" optional property for runs /run filters - fixes

okolesn:
[[NGS Portal] Add "project id" optional property for runs / run filters #3866](https://github.com/epam/cloud-pipeline/issues/3866)

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineDao.java
@@ -252,8 +252,6 @@ public class PipelineDao extends NamedParameterJdbcDaoSupport {
             final Long folderId = rs.getLong(FOLDER_ID.name());
             if (!rs.wasNull()) {
                 pipeline.setParentFolderId(folderId);
-            }
-            if (!rs.wasNull()) {
                 pipeline.setParent(new Folder(folderId));
             }
         }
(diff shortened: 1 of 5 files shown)


api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineDao.java
```diff
@@ -253,6 +253,9 @@ private static void initPipeline(final ResultSet rs, final Pipeline pipeline) th
             if (!rs.wasNull()) {
                 pipeline.setParentFolderId(folderId);
             }
+            if (!rs.wasNull()) {
```
SilinPavel: It seems that this if is the same as previous, can we join code under the ifs?

api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineDao.java
```diff
@@ -253,6 +253,9 @@ private static void initPipeline(final ResultSet rs, final Pipeline pipeline) th
...
+                pipeline.setParent(new Folder(folderId));
```
SilinPavel, line 255: As far as I understand we have the same problem (that parent is not set during load, as result f.e. FolderApiService,loadProject method can't work correctly ) not only for pipeline but also at least for DataStorage and Folder, RunConfiguration. So, let's fix it as well

SilinPavel (changes requested):

PR #3894, opened 2025-02-05
Issue 3833 sync_to_storage: batch mode

SilinPavel:
Batch mode for sync_to_storage:

Can be enabled with `CP_SYNC_TO_STORAGE_BATCH_MODE=1`

Limitations:
Would be possible to use only in the next cases:
 - All entries in sync config are files
 - All files configured for the same bucket to sync: either to default sync bucket (with out destination path in spec) of with the same bucket in destination path configured explicitly

### DIFF
--- a/pipe-cli/src/utilities/datastorage_operations.py
+++ b/pipe-cli/src/utilities/datastorage_operations.py
@@ -184,6 +184,16 @@ class DataStorageOperations(object):
                                       permission_to_check, include, exclude, force, skip_existing, sync_newer,
                                       verify_destination, on_unsafe_chars, on_unsafe_chars_replacement, on_empty_files)
             sys.exit(0)
+
+        # TODO: rewrite tuple to object
+        # items - collection of items (tuples), each item represents file + additional information about this file
+        # item structure:
+        #  object_type = item[0]
+        #  full_path = item[1]
+        #  relative_path = item[2]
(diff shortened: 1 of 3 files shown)


pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -802,11 +818,22 @@ def _transfer_item(cls, item, manager, source_wrapper, destination_wrapper, tran
         full_path = item[1]
         relative_path = item[2]
         size = item[3]
+
+        # Unused here but added for visibility of the field
+        source_modification_date = None
+        if len(item) >= 5:
+            source_modification_date = item[4]
+
+        if len(item) >= 6:
```
ekazachkova: it looks like we can extract method here and reuse it at [303](https://github.com/epam/cloud-pipeline/pull/3894/files#diff-e4e3d5936cbd7e5d07fbdc6e346208d3677b3910de8a7e1bed2d65a5984e2b37R303) line:
```
def <find_relative_path>(item, relative_path):
    return item[5] if len(item) >= 6 else relative_path
```
and also in this case we do not need to use if/else in [351](https://github.com/epam/cloud-pipeline/pull/3894/files#diff-e4e3d5936cbd7e5d07fbdc6e346208d3677b3910de8a7e1bed2d65a5984e2b37R351) line

Issue #3919, opened 2025-02-18
Allow to grant user/group permissions on storage paths


ekazachkova:
**Background**
At the moment, Cloud Pipeline does not allow to grant permissions directly on storage path and it would be nice to support it.

**Approach**
### API part
1.  a new flag `pathPermissionsEnabled` shall added for storage object. This flag indicates that path permissions management available for specified storage.  This flag shall be available to apply only on storage creation.
2. a new method  to update all permissions for specific user or group shall be implemented: 
`POST /datastorage/<storageID>/paths/permissions?sidName=<user/group>&principal=<true/false>` 
```
# Request Body
[{
    # to grant permissions on file
    "fileName": "file.txt",
    "mask": 1,
    "folderPath": "path/to/folder"
  },
  {
    # to grant permissions on folder
    "mask": 5,
    "folderPath": "path/to/folder"
  }, ...
]
```
This method shall be available for admins and storage owners. 
If `pathPermissionsEnabled` flag is switched off for specified storage an error shall be occurred.
This method shall be used when owner/admin desired to grant path permissions for specific user or group. All already existing permissions for specified user/group will be overwrited.

3. a new method to fetch all path permissions for current user on specified storage shall be implemented:
`GET /datastorage/<storageID>/paths/permissions`
```
# Response Body
[{
    # if permissions granted on file
    "fileName": "file.txt",
    "mask": 1,
    "folderPath": "path/to/folder"
  },
  {
    # if permissions granted on folder
    "mask": 5,
    "folderPath": "path/to/folder"
  }, ...
]
```
This method shall be available for users with at least READ permissions on storage. 
If `pathPermissionsEnabled` flag is switched off for specified storage an error shall be occurred. 
This method shall be useful for `pipe storage` CLI operations.

4. all currently implemented API methods for storage files/folders management shall respect permissions on paths:
   - if write permissions are not granted to file/folder deletion shall not be available
   - if permissions are only granted for some of the folders or files in the requested directory folders or files without permissions shall not be listed
   - and so on ...

5. when storage deletion requested all path permissions shall be removed as well.
6.  a new API method to load users or groups that have path permissions for specified storage:
`GET /datastorage/<storageID>/paths/permissions/sids`
```
[{
    "name": "USER",
    "principal": true 
},
{
    "name": "GROUP",
    "principal": false 
},
]
```
This method shall be available for admins and storage owners. 

7. a new API method to delete storage path permissions for specified users or groups by storage ID.
[...]

PR #3923, opened 2025-02-24
Issue #3919: [API] Allow to grant user/group permissions on storage paths 

ekazachkova:
relates to #3919 

Implements points 1-7 for API part

### DIFF
--- a/api/src/main/java/com/epam/pipeline/acl/datastorage/DataStorageApiService.java
+++ b/api/src/main/java/com/epam/pipeline/acl/datastorage/DataStorageApiService.java
@@ -22,6 +22,7 @@ import com.epam.pipeline.controller.vo.DataStorageVO;
 import com.epam.pipeline.controller.vo.data.storage.UpdateDataStorageItemVO;
 import com.epam.pipeline.controller.vo.EntityFilterVO;
 import com.epam.pipeline.controller.vo.security.EntityWithPermissionVO;
+import com.epam.pipeline.dto.datastorage.permissions.StoragePathPermissions;
 import com.epam.pipeline.entity.AbstractSecuredEntity;
 import com.epam.pipeline.entity.SecuredEntityWithAction;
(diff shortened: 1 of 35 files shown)


api/src/main/java/com/epam/pipeline/acl/datastorage/DataStorageApiService.java
```diff
@@ -416,4 +418,25 @@ public List<DataStorageTagSearchResult> searchDataStorageItemByTag(
             final DataStorageObjectSearchByTagRequest request) {
         return dataStorageManager.searchDataStorageItemByTag(request);
     }
+
```
mzueva: I'd suggest to have a separate set of controller-service-manager for this permissions as data storage classes are already too big

core/src/main/java/com/epam/pipeline/entity/datastorage/AbstractDataStorage.java
```diff
@@ -86,6 +86,11 @@ public abstract class AbstractDataStorage extends AbstractSecuredEntity {
     private Set<String> linkingMasks;
     private Long sourceStorageId;
 
+    /**
+     * Indicates that permissions can be granted for particular paths in this storage.
+     */
+    private boolean pathPermissionsEnabled;
```
mzueva, line 92: Let's also add this new field to `cloud-pipeline-common` package

mzueva (changes requested):

api/src/main/java/com/epam/pipeline/acl/datastorage/DataStorageApiService.java
ekazachkova: done

core/src/main/java/com/epam/pipeline/entity/datastorage/AbstractDataStorage.java
ekazachkova: added

Issue #3988, opened 2025-06-04
[GCP] Data Catalog improvements


mzueva:
**Background**
At the moment we observe some issues with data catalog including:
- low performance for large indices
- instability of security logs
- duplication of indices

**Approach**
The following solutions shall be implemented:
- instability of security logs is addressed via https://github.com/epam/cloud-pipeline/issues/3947, as audit logs are now stored and managed using GCP APIs
- low performance for large indices: 
  - allow to exclude files from indexing
  - support multi node deployment for Elasticsearch on GKE to improve service performance
 - duplication of indices: introduce additional clean up logic during index management

PR #3998, opened 2025-06-16
Issue #3988 add logic for removing detached/unused indexes

kbashpayev:
Issue #3988
Extended logic with additional deletion of unused/detached indexes from alias which are occurred time to time.


mzueva:
merged into `stage/0.20/gcp`

PR #4005, opened 2025-06-20
Issue #3988 Add logic for squashing hidden files into single document/record.

kbashpayev:
Extend logic with squashing hidden files into single file/document/record.

2 new properties added:
1. sync.hidden.files.use.squashing - property for switching squashing hidden files
2. sync.hidden.files.squashed.name - the name under the which hidden files will be squashed


mzueva:
merged into `stage/0.20/gcp`

Issue #3059, opened 2023-02-08
Data access audit


mzueva:
**Background**
Any access (read/write/delete, excluding listing) to the data stored in the object storages shall be logged.

NShaforostov:
Docs added via #3147 and located [here](https://github.com/epam/cloud-pipeline/blob/develop/docs/md/manual/12_Manage_Settings/12.12._System_logs.md#type-filter).

maryvictol:
Test cases were created by #3146 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/System_Management/3059-3075_audit_pipe_mount_operation).

PR #4024, opened 2025-07-04
Audit enhancements

ekazachkova:
relates to #3059

- Support runID in audit log message with the following format: `[run_id #12345] <message>`. The runID shall be added to the audit logs for all PipeCLI commands that are run from Cloud Pipeline instances. To determine run ID the `RUN_ID` environment variable shall be used.
- Log mount event to audit logs with the following format `MOUNT <mount bucket>`.
- Log run share event to audit logs.


pipe-cli/src/common/audit.py
```diff
@@ -319,6 +322,12 @@ def _get_storage_id(self, storage):
     def flush(self):
         pass
 
+    def _get_run_id(self):
+        run_id = os.getenv('RUN_ID')
+        if not run_id:
+            return ''
+        return '[#%s] ' % run_id
```
SilinPavel: Wouldn't it be better to explicitly mention that #<runid> is a run, like `[run_id #< runid >]`?

SilinPavel (changes requested):

maryvictol:
Test cases were created by #4031 and located [here](https://github.com/epam/cloud-pipeline/tree/develop/docs/testcases/UI/System_Management/3059-3075-4039_audit_pipe_mount_operation/).

PR #4027, opened 2025-07-09
GUI Remove firecloud

AleksandrGorodetskii:


rodichenko (changes requested):
The PR is not GUI-only, it touches `deploy/...` and `docs/testcases/...` with unrelated changes. Can we create a separate PR with GUI-only firecloud removal changes instead?

PR #4054, opened 2025-08-11
Update Run Shift Policy: Allow restarting a worker run

kamyshova:
This PR allows restarting a worker run via the `CP_CAP_RESCHEDULE_WORKER_RUN=true` run parameter or the system preference `launch.run.reschedule.worker.enabled`.


api/src/main/java/com/epam/pipeline/manager/pipeline/RunRegionShiftHandler.java
```diff
@@ -148,7 +149,11 @@ private boolean validate(final PipelineRun parentRun, final PipelineRun currentR
             return false;
         }
 
-        if (parentRun.isWorkerRun()) {
+        final boolean shouldRescheduleWorker = parentRun.getParameterValue(CP_CAP_RESCHEDULE_RUN_WORKER_PARAM)
```
SilinPavel: I would move this code to the method:

```
 if (parentRun.isWorkerRun() && !shouldRescheduleWorker()) {
 ....
}
```

this will allow to call such method only if we currently process worker node

SilinPavel (approved):

Issue #4081, opened 2025-09-11
Support fallback instance types


mzueva:
**Background**
At the moment platform allows to select a specific instance type for a run. For large workloads when we don't have guarantee that specific instance types are available, we need to support list of fallback instance types, that shall be used, if main instance type is not available.

**Approach**
1. POST method `restapi/run` (`PipelineRunController.runPipeline`) shall support additional body field `List<String> fallbackInstanceTypes`
2. Value of `fallbackInstanceTypes` shall be persisted in DB in `pipeline_run` object
3. `AutoscaleManager` class should load `fallbackInstanceTypes` from db and propagate it to `createNodeForRun` (as part of `requiredInstance` parameter), further to `AWSInstanceService.scaleUpNode method` and finally added to `NodeUpCommand` command as `--fallback_ins_types` argument
4. `NodeUpCommand` will return four arguments as output, including actual started instance type, which shall be persisted as `PipelineRun.runInstance.nodeType` field.

PR #4089, opened 2025-09-25
issue_4081_add_fallback_types

kbashpayev:
Issue #4081, add fallback instance types for large workloads when we don't have guarantee that specific instance type is available.


api/src/main/java/com/epam/pipeline/manager/cloud/CommonCloudInstanceService.java
```diff
@@ -131,6 +131,11 @@ private void readInstanceId(final RunInstance instance, final String output) {
             instance.setNodeId(node[0]);
             instance.setNodeIP(node[1]);
             instance.setNodeName(node[2]);
+        } else if (node.length == 4) {
+            instance.setNodeId(node[0]);
+            instance.setNodeIP(node[1]);
+            instance.setNodeName(node[2]);
+            instance.setNodeType(node[3]);
```
SilinPavel, line 138: can be simplified like:\
```
} else if (node.length == 4) {
            instance.setNodeType(node[3]);
}
```

api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
```diff
@@ -1352,6 +1359,18 @@ public static PipelineRun parsePipelineRun(ResultSet rs) throws SQLException {
             instance.setNodeId(rs.getString(NODE_ID.name()));
             instance.setNodeIP(rs.getString(NODE_IP.name()));
             instance.setNodeType(rs.getString(NODE_TYPE.name()));
+
+            try {
+                String fallbackInstanceTypes = rs.getString(INSTANCE_FALLBACK_TYPES.name());
+                if (fallbackInstanceTypes != null && !fallbackInstanceTypes.equals("{}")) {
```
SilinPavel, line 1365: it should be `!fallbackInstanceTypes.equals("[]")` as we have fallbackInstanceTypes as an array

api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
```diff
@@ -1260,6 +1263,10 @@ private static void addInstanceFields(PipelineRun run, MapSqlParameterSource par
                     instance.map(RunInstance::getCloudProvider).map(CloudProvider::name).orElse(null));
             params.addValue(NODE_PLATFORM.name(), instance.map(RunInstance::getNodePlatform).orElse(null));
             params.addValue(NODE_POOL_ID.name(), instance.map(RunInstance::getPoolId).orElse(null));
+            instance.ifPresent(ins ->
```
SilinPavel, line 1266: we should always pass a parameter, even if it isn't present, to compile correct SQL.
should be in the same notation as previous optional params:
```
params.addValue(INSTANCE_FALLBACK_TYPES.name(), instance.map(RunInstance::getFallbackInstanceTypes).map(JsonMapper::convertDataToJsonStringForQuery).orElse(null));
```

api/src/main/java/com/epam/pipeline/dao/pipeline/PipelineRunDao.java
```diff
@@ -1352,6 +1359,18 @@ public static PipelineRun parsePipelineRun(ResultSet rs) throws SQLException {
             instance.setNodeId(rs.getString(NODE_ID.name()));
             instance.setNodeIP(rs.getString(NODE_IP.name()));
             instance.setNodeType(rs.getString(NODE_TYPE.name()));
+
+            try {
```
SilinPavel, line 1363: since we updated all queries, why we need this catch? it seems that we should remove it, because if exception happens in runtime - means we have unexpected situation and should fix it

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineConfigurationManager.java
```diff
@@ -194,6 +194,9 @@ public PipelineConfiguration mergeParameters(PipelineStart runVO, PipelineConfig
         } else {
             configuration.setInstanceType(defaultConfig.getInstanceType());
         }
+        if (runVO.getFallbackInstanceTypes() != null) {
+            configuration.setFallbackInstanceTypes(runVO.getFallbackInstanceTypes());
```
SilinPavel, line 198: the same way as for other fields we should fallback to defaultConfig value, to be able to define this value on pipeline level. Otherwise user will need to specify these fallback types each time or pipeline run

core/src/main/java/com/epam/pipeline/entity/configuration/PipelineConfiguration.java
```diff
@@ -88,6 +88,8 @@ public class PipelineConfiguration implements Cloneable {
     @JsonProperty(value = INSTANCE_SIZE)
     private String instanceType;
 
+    private List<String> fallbackInstanceTypes;
```
SilinPavel, line 91: lets have the same notation for new field as for the all others:
```
@JsonProperty(value = FALLBACK_INSTANCE_TYPE)
```
where FALLBACK_INSTANCE_TYPE = 'fallback_instance_type'

SilinPavel (changes requested):

Issue #4107, opened 2025-10-08
[pipe-cli] Allow login without explicitly specified tokens


ekazachkova:
**Background**
At the moment, Cloud Pipeline requires tokens for `pipe configure` and it would be nice to avoid explicitly specifying the token. 

**Approach**

### Overview 

1. user initiates login process:  `pipe configure --login --api <...> --proxy <...>`
2. `pipe-cli` generates `code_verifier` and `code_challenge` pair
3. `pipe-cli` redirects user to browser to login (if available) or prints the login url. This url contains the following essential parts: `code_challenge` and `code_challenge_method`.
4. user logs in and gains access to the system
5. at this time `pipe-cli` polls API with `GET /access/code?code_challenge=<...>` method to obtain authorization code. Authorization code shall be created as soon as user successfully logs in to the platform. 
6. once authorizationn code obtained `pipe-cli` exchanges this code to access token with `GET /access/token?code=<...>&code_verifier=<...>` where 
   - `code` is the code that has been received at step (5) 
   - `code_verifier` is an random string generated at step (2)
7. API confirms `code_verifier` matches `code_challenge` and if so returns access token. 
8. `pipe-cli` saves this token and uses it to access API methods.

#### API methods
A new methods shall be implemented:
1. `GET /access/auth?code_challenge=<>&code_challenge_method=<>` - that initiates user login. This method shall support auth. Once user has logged in API shall:
   - validate `code_challenge` and `code_challenge_method`. Code challenge method shall be allowed (`system.access.allowed.code.challenge.methods`)
   - generate authorization `code`
       - length shall be managed vis preference `system.access.code.length` (default: 23) 
       - `code` shall be used to obtain access token
       - `code` must not not be used more than once
       - `code` must be available for 10 minutes only; manage timeout via system preference `system.access.code.ttl.minutes` (default: 10)
   - save the following records to DB:
       - `user_name` - the user name
       - `code` - authorization code
       - `code_challenge`
       - `code_challenge_method`
       - `created` - the timestamp
       - `issued` - flag, indicates authorizatoin code has beed already issued
    This data shall be used by API for future unauthorized requests to validate user.
2. `GET /access/code?code_challenge=<>` returns authorization `code` for specified `code_challenge` if user has already logged in. If not an empty response shall be returned.
3. `GET /access/token?code_verifier=<>&code=<>` exchanges authorization `code` for JWT token for `pipe-cli`. This method shall:
   - verify received authorization `code` present
[...]

PR #4113, opened 2025-10-13
Issue #4107: [pipe-cli] Allow login without explicitly specified tokens

ekazachkova:
relates to #4107


pipe-cli/pipe.py
```diff
@@ -312,8 +313,10 @@ def cli():
 
 
 @cli.command()
+@click.option('-tl', '--tokenless',
```
SilinPavel: I would prefer to name it something like `--login`. Probably it would reflect the behaviour more accurate

pipe-cli/src/utilities/tokenless_access_manger.py
```diff
@@ -0,0 +1,87 @@
...
+    def fetch_token(self, no_launch_browser):
+        try:
+            code_verifier, code_challenge = pkce.generate_pkce_pair()
+            self._initiate_login(code_challenge, no_launch_browser)
+            code = self._find_access_code(code_challenge)
+            return self._exchange_code_for_token(code, code_verifier)
+        except Exception as error:
+            error_message = str(error)
+            if 'Access is denied' in error_message:
+                click.echo('Something went wrong. Please try again.', err=True)
```
SilinPavel: I think in case when we know what happened, it is better to provide more clear message. But now while we have info that we couldn't obtain a token from server, saying to user only abstract message.

api/src/main/resources/db/migration/v2025.10.15_13.00__access_code.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS pipeline.access_code (
```
SilinPavel, line 1: should we change name of the migration to `access_code`?

api/src/main/java/com/epam/pipeline/manager/access/AccessCodeCleanerCore.java
```diff
@@ -0,0 +1,51 @@
...
+import lombok.extern.slf4j.Slf4j;
+import net.javacrumbs.shedlock.core.SchedulerLock;
+import org.springframework.stereotype.Service;
+
+import java.util.Objects;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class AccessCodeCleanerCore {
```
SilinPavel: we don't have complicated logic in both `AccessCodeCleaner` and `AccessCodeCleanerCore` classes, so I'm not sure if we should split it on two classes.

pipe-cli/src/utilities/tokenless_access_manger.py
```diff
@@ -0,0 +1,87 @@
...
+    def _find_access_code(self, code_challenge):
+        attempts = 0
+        while attempts < self.attempts:
+            attempts = attempts + 1
+            response = AccessAPI(self.api).get_code(code_challenge) or {}
+            code = response.get('code')
+            if code:
+                return code
+            sleep(self.timeout)
+        click.echo('No token received. Please try again.', err=True)
```
SilinPavel: probably typo? should we say `No code received`

pipe-cli/pipe.py
```diff
@@ -346,17 +349,37 @@ def cli():
 @click.option('-cs', '--config-store',
               help='CLI configuration mode(home-dir/install-dir)',
               default='home-dir')
-def configure(auth_token, api, timezone, proxy, proxy_ntlm, proxy_ntlm_user, proxy_ntlm_domain, proxy_ntlm_pass, codec,
-              config_store):
+@click.option('-nb', '--no-launch-browser',
+              help='Prevents the command from automatically opening a web browser. '
+                   'Works in combination with --tokenless option; without this option will be ignored',
```
SilinPavel: > without this option will be ignored

Not sure what will be ignored?

pipe-cli/src/utilities/tokenless_access_manger.py
```diff
@@ -0,0 +1,87 @@
...
+                sys.exit(1)
+            else:
+                click.echo(error_message, err=True)
+                sys.exit(1)
+
+    def _initiate_login(self, code_challenge, no_launch_browser):
+        authorization_url = self._build_login_url(code_challenge)
+
+        if self._browser_allowed(no_launch_browser):
+            click.echo("Please log in by browser.")
```
SilinPavel: maybe better to say `Please continue authentication in the browser.` or `You should be redirected to authentication URL in your browser.` Since it is a user-faced message

SilinPavel (changes requested):

SilinPavel (approved):

Issue #4108, opened 2025-10-09
Support Azure Managed Identity Authorization


okolesn:
**Background**

Workloads deployed on an Azure Kubernetes Services (AKS) cluster require Microsoft Entra application credentials or managed identities to access Microsoft Entra protected resources. 
We need to support Managed Identities.
[Workload Identity Overview](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet)

**Approach**

1. To support managed identities we need to introduce [Azure Identity client library (azure-identity)](https://learn.microsoft.com/en-us/java/api/overview/azure/identity-readme?view=azure-java-stable).

2. Current azure-mgmt-resources:1.19.0 - old SDK (Track 1)  cannot use DefaultAzureCredential from azure-identity.
New azure-resourcemanager-resources - new SDK (Track 2) supports DefaultAzureCredential.
Switching from azure-mgmt-resources to azure-resourcemanager-resources library.

3. During Azure Region registration in Cloud Pipeline we have the following properties to be specified:

For Storages:
- Storage account 
- Storage account key
If account key value is empty, Managed Identity authentification is to be used(DefaultAzureCredential), otherwise we are authentication using account key(StorageSharedKeyCredential)

For Resources:
- Subscription
- Auth file
Auth approach is the same - if Auth file value is empty, Managed Identity will be used for authentication.

okolesn:
**Azure settings**
[Deploy and configure workload identity on an Azure Kubernetes Service (AKS) cluster](https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster)

1. Make sure Workload Identity and OpenID Connect (OIDC) are enabled for AKS:

Azure Portal: Cluster -> Security configuration - > Enable Workload Identity=true, Enable OIDC=true
If not, update the cluster:
```
az aks update --resource-group <resource-group> --name <cluster-name> --enable-oidc-issuer --enable-workload-identity
```

2. Get/Create Managed Identity

```
az identity create \
    --name <identity-name> \
    --resource-group <resource-group> \
    --location <region-id> \
    --subscription <subscription-id>
```
Get Identity name and client-id
```
az identity list --resource-group <resource-group>
```

3. Federate Managed Identity

Get Issuer URL:
Azure Portal: Cluster -> Security configuration -> enable "OpenID Connect (OIDC)" and "Workload Identity" -> Copy OIDC Issuer URL 
```
[...]

PR #4126, opened 2025-10-25
Support Azure Managed Identity Authorization

okolesn:
[Support Azure Managed Identity Authorization #4108](https://github.com/epam/cloud-pipeline/issues/4108)


api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureBlobStorageProvider.java
```diff
@@ -325,9 +327,13 @@ public DataStorageItemType getItemType(final AzureBlobStorage dataStorage,
...
     }
 
-    private AzureStorageHelper getAzureStorageHelper(final AzureBlobStorage storage) {
+    public AzureStorageHelper getAzureStorageHelper(final AzureBlobStorage storage) {
         final AzureRegion region = cloudRegionManager.getAzureRegion(storage);
+        final String storageAccountName = region.getStorageAccount();
         final AzureRegionCredentials credentials = cloudRegionManager.loadCredentials(region);
-        return new AzureStorageHelper(region, credentials, azEvents, messageHelper);
+        final String storageAccountKey = credentials.getStorageAccountKey();
+        final BlobServiceClient blobServiceClient = AzureStorageHelper.getBlobServiceClient(storageAccountName,
```
SilinPavel: do we support here Azure Managed Identity? it seems that we still can authenticate only with `storageAccountKey`

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureHelper.java
```diff
@@ -1,51 +1,98 @@
...
+                subscriptionId = config.get(SUBSCRIPTION_ID);
+                credential = new ClientSecretCredentialBuilder()
+                        .clientId(clientId)
+                        .clientSecret(clientSecret)
+                        .tenantId(tenantId)
+                        .build();
+                profile = new AzureProfile(tenantId, subscriptionId, AzureEnvironment.AZURE);
+            } else {
+                log.info("Authentication using DefaultAzureCredential");
+                credential = new DefaultAzureCredentialBuilder().build();
```
SilinPavel: I think it  would better to have in AzureRegion a new field `managedIdentity`, so we could authenticate not in a implicit but explicit way here like:
````
  new DefaultAzureCredentialBuilder()
                .managedIdentityClientId(managedIdentity)
                .build();
````

So, then this method will accept a region instead of authFile, then we based on provided values can decide how to authenticate

File: api/src/main/java/com/epam/pipeline/manager/region/AzureRegionHelper.java [hunk 3]
```diff
@@ -144,24 +134,16 @@ public String serializeCredentials(AzureRegion region, AzureRegionCredentials cr
...
-                    String.format(BLOB_URL_FORMAT, storageAccountName)),
-                    StorageURL.createPipeline(credentials, pipelineOptions));
-            serviceURL.getProperties().blockingGet();
+            final BlobServiceClient blobServiceClient = AzureStorageHelper.getBlobServiceClient(storageAccountName,
+                    storageAccountKey);
+            blobServiceClient.getProperties();
         } catch (Exception e) {
-            throw new IllegalArgumentException(
-                    messageHelper.getMessage(MessageConstants.ERROR_AZURE_STORAGE_CREDENTIAL_INVALID), e);
+            throw new IllegalArgumentException("Invalid Azure Storage credentials", e);
```
SilinPavel: why can we use `MessageConstants.ERROR_AZURE_STORAGE_CREDENTIAL_INVALID`?

api/src/main/java/com/epam/pipeline/manager/region/AzureRegionHelper.java
SilinPavel: what if there is no storage account credentials, and we use managed identity? probably we should be able to understand which type of creds are configured for region and validate appropriately

SilinPavel: there can be a case when we don't have auth file anymore

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureBlobStorageProvider.java
SilinPavel: why we generate with static `AzureStorageHelper.getBlobServiceClient` client here and then pass it back to AzureStorageHelper constructor? shouldn't we then encapsulate it inside AzureStorageHelper?

api/src/main/java/com/epam/pipeline/manager/cloud/azure/AzureVMService.java
```diff
@@ -290,23 +290,30 @@ private Optional<VirtualMachine> findVmByName(final String authFile,
     }
 
     private Optional<VirtualMachineScaleSet> findVmScaleSetByName(final String authFile,
```
SilinPavel: since we not always now would have auth file, I think in all such places we should remove direct use of auth file, and instead pass a region + have some util place to provide authentication based on azure region settings

SilinPavel (changes requested):

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureHelper.java
```diff
@@ -1,51 +1,103 @@
...
-            final Azure.Configurable builder = Azure.configure()
-                    .withLogLevel(LogLevel.BASIC);
-            return authenticate(authFile, builder)
+            final AzureResourceManager azure = AzureResourceManager
+                    .authenticate(credentials.getCredential(), credentials.getProfile())
                     .withDefaultSubscription();
-        } catch (IOException e) {
-            log.error(e.getMessage(), e);
-            throw new AzureException(e);
+            log.info("Authenticated to subscription: {}", azure.subscriptionId());
```
SilinPavel: probably should be debug level

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureHelper.java
```diff
@@ -1,51 +1,103 @@
...
+            tenantId = config.get(TENANT_ID);
+            subscriptionId = config.get(SUBSCRIPTION_ID);
+            credential = new ClientSecretCredentialBuilder()
+                    .clientId(clientId)
+                    .clientSecret(clientSecret)
+                    .tenantId(tenantId)
+                    .build();
+            profile = new AzureProfile(tenantId, subscriptionId, AzureEnvironment.AZURE);
+        } else {
+            throw new AuthenticationException("Failed to get Azure credentials.");
```
SilinPavel: Can you expand message and log that no values of auth file or menagedIdentity where provided

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
```diff
@@ -162,39 +138,62 @@ public Stream<DataStorageFile> listDataStorageFiles(final AzureBlobStorage dataS
...
-        final int page = Optional.ofNullable(pageSize).orElse(MAX_PAGE_SIZE);
-        final HierarchyIterator iterator = AbstractListingIterator.hierarchy(getContainerURL(storage), prefix, marker,
-                page);
-        final List<AbstractDataStorageItem> items = list(iterator)
+        int effectivePageSize = Optional.ofNullable(pageSize).orElse(MAX_PAGE_SIZE);
+
+        final ListBlobsOptions options = new ListBlobsOptions()
+                .setPrefix(prefix)
+                .setMaxResultsPerPage(effectivePageSize);
+        final PagedIterable<BlobItem> pagedIterable = containerClient.listBlobsByHierarchy("/", options, null);
```
SilinPavel: please use constants with names that good describe such values as `/`

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
```diff
@@ -162,39 +138,62 @@ public Stream<DataStorageFile> listDataStorageFiles(final AzureBlobStorage dataS
...
+
+        final PagedResponse<BlobItem> pageResponse = (continuationToken == null)
+                ? pagedIterable.iterableByPage().iterator().next()
```
SilinPavel: what if result doesn't have `next`? Probably we should handle such case

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
```diff
@@ -205,23 +204,18 @@ public DataStorageFile createFile(final AzureBlobStorage dataStorage,
...
-    }
-
     public DataStorageFolder createFolder(final AzureBlobStorage dataStorage, final String path) {
-        String folderPath = ProviderUtils.withoutLeadingDelimiter(ProviderUtils.withTrailingDelimiter(path.trim()));
-        if (directoryExists(dataStorage, folderPath)) {
+        validateDirectory(dataStorage, path, false);
+        String fullPath = getFullPath(path);
+        fullPath += ProviderUtils.FOLDER_TOKEN_FILE;
+        final BlockBlobClient blobClient = getBlobClient(dataStorage, fullPath);
+        try (ByteArrayInputStream emptyStream = new ByteArrayInputStream(new byte[0])) {
```
SilinPavel: probably new byte[0] can be a constant

api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
```diff
@@ -573,117 +577,74 @@ private void validatePath(final AzureBlobStorage storage, final String path, fin
...
     }
 
     private void copyBlobs(final AzureBlobStorage storage, final String sourceFolder, final String destinationFolder) {
-        listFilesRecursively(storage, sourceFolder)
-                .forEach(item -> {
-                    final String relativePath = StringUtils.removeStart(item.getPath(), sourceFolder);
-                    final String newPath = String.format("%s/%s", destinationFolder, relativePath);
-                    copyBlob(storage, item.getPath(), newPath);
-                });
+        final BlobContainerClient containerClient = getBlobContainerClient(storage);
```
SilinPavel: we create here a client and then in `copyBlob` will create it again, lets reuse it. (I believe this approach can be applied not only here but in other places also, f.e. moveFile)

api/src/main/java/com/epam/pipeline/manager/region/AzureRegionHelper.java
```diff
@@ -141,27 +131,17 @@ public String serializeCredentials(AzureRegion region, AzureRegionCredentials cr
         }
     }
 
-    private void validateStorageAccount(final String storageAccountName, final String storageAccountKey) {
-        Assert.isTrue(StringUtils.isNotBlank(storageAccountName),
```
SilinPavel, line 145: I think that we should leave such checks, but make it conditional depending on region settings

SilinPavel (approved):

Issue #4161, opened 2025-11-14
Data Storage Temp Credentials for Azure doesn't work


okolesn:
**Describe the bug**
/datastorage/tempCredentials/ doesn't work

PR #4162, opened 2025-11-14
Issue 4161 - Data Storage Temp Credentials for Azure doesn't work

okolesn:
[Data Storage Temp Credentials for Azure doesn't work #4161](https://github.com/epam/cloud-pipeline/issues/4161)

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
+++ b/api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
@@ -96,6 +96,8 @@ public class AzureStorageHelper {
     private static final Duration AZURE_STORAGE_TIMEOUT = Duration.ofSeconds(30);
     private static final byte[] EMPTY_CONTENT = new byte[0];
     private static final String STORAGE_CLASS = "StorageClass";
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/epam/pipeline/manager/datastorage/providers/azure/AzureStorageHelper.java
```diff
@@ -453,6 +454,28 @@ public BlobContainerClient getBlobContainerClient(final AzureBlobStorage storage
...
+        final BlobContainerClient blobContainerClient = getBlobContainerClient(dataStorage);
+        final DataStorageAction dataStorageAction = actions.get(0);
+        Assert.isTrue(actions.size() == 1, "Multiple actions is not supported for AZURE provider");
+
+        final BlobServiceSasSignatureValues values = new BlobServiceSasSignatureValues(expiryTime,
+                buildPermissions(dataStorageAction))
+                .setProtocol(SasProtocol.HTTPS_ONLY)
+                .setContentType("container");
+        addIPRangeToSASValue(values);
+        try {
```
SilinPavel: instead of try - catch, let's explicitly check the way how we should generate it, if we have manageIdentity configured -   `generateUserDelegationSas`, if storageKey - `generateSas` else throw that we misconfigured

Issue #4151, opened 2025-11-10
Metadata entities filter API: add facets


rodichenko:
**Background**
It would be useful if Cloud Pipeline **metadata entities** filter (`metadataEntity/filter`) endpoint returns per-column statistics, i.e. how many records for each column value.

**Request payload**
Let's introduce *optional* field `facets` (array of column/field names) - if request contains this field, the per-field per-value statistics should be calculated:

- For each metadata entity field (column), a list of unique values should be extracted.
- For each unique field value, total count of entities should be calculated.
- Additionally, the total count of empty/not set values should be calculated.

**Response format**

If facets were requested, response body should contain the following object:

```json
{
  "payload": {
     "elements": [...],
     "totalCount": ...,
     "facets": {
        "field1": {"counts": {"Unique value 1": 123, "Unique value 2": 3}, "empty": 0},
        "field2": {"counts": {}, "empty": 126},
        ...
     }
  }
}
```

PR #4167, opened 2025-11-18
Metadata entities filter API facets

kbashpayev:
Issue #4151, Metadata entities filter API - add facets, statistics per column.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/dao/metadata/MetadataEntityDao.java
+++ b/api/src/main/java/com/epam/pipeline/dao/metadata/MetadataEntityDao.java
@@ -18,6 +18,7 @@ package com.epam.pipeline.dao.metadata;
 
 import com.epam.pipeline.config.JsonMapper;
 import com.epam.pipeline.dao.DaoHelper;
+import com.epam.pipeline.entity.metadata.Facet;
 import com.epam.pipeline.entity.metadata.FireCloudClass;
 import com.epam.pipeline.entity.metadata.LogicalSearchOperator;
 import com.epam.pipeline.entity.metadata.MetadataClass;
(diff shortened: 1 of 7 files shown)


api/src/main/java/com/epam/pipeline/entity/metadata/MetadataFilter.java
```diff
@@ -90,4 +92,13 @@ public static class OrderBy {
         private boolean desc = false;
         private boolean predefined = false;
     }
+
+    @Getter
+    @Setter
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class Facet {
```
SilinPavel: I would name this class differently from other Facet object. f.e. `FacetRequest`

api/src/main/java/com/epam/pipeline/entity/metadata/Facet.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Getter
+@Setter
+@NoArgsConstructor
+public class Facet {
```
SilinPavel, line 25: let's make this value class immutable, so only `@Getter` and `@Builder`, then in DAO you will create it like 

```
Facet.builder().counts(...).build()
or
Facet.builder().empty(...).build
```

api/src/main/java/com/epam/pipeline/dao/metadata/MetadataEntityDao.java
```diff
@@ -301,6 +320,62 @@ private String buildFilterQuery(MetadataFilter filter) {
...
+        return clauseJoiner.toString();
+    }
+
+    private Map<String, Facet> facetGroupsMapper(List<Triple<String, String, Integer>> groupingResults) {
+        return groupingResults.stream()
+            .collect(Collectors.groupingBy(Triple::getLeft, Collectors.collectingAndThen(
+                Collectors.groupingBy(Triple::getMiddle, Collectors.summingInt(Triple::getRight)),
+                    innerMap -> {
+                        Facet facet = new Facet();
+                        innerMap.forEach((value, count) -> {
```
SilinPavel: it seems that instead of going through all values you could find `empty` one and then do smthg like that:

```
emptyCount = counts.remove(EMPTY)
return Facet.builder().counts(counts).empty(emptyCount).build()
```

SilinPavel (changes requested):

SilinPavel (approved):

PR #4249, opened 2026-01-29
GUI LaunchForm 'metadata-entity' params: resolve folderId from path o…

AleksandrGorodetskii:
…r number

### DIFF
--- a/client/src/components/pipelines/launch/form/LaunchPipelineForm.js
+++ b/client/src/components/pipelines/launch/form/LaunchPipelineForm.js
@@ -5509,6 +5509,7 @@ class LaunchPipelineForm extends localization.LocalizedReactComponent {
             pipelineObject: pipeline
           }
         );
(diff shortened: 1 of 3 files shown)


client/src/components/pipelines/launch/form/LaunchPipelineForm.js
```diff
@@ -5500,7 +5510,7 @@ class LaunchPipelineForm extends localization.LocalizedReactComponent {
           }
         );
         parametersMetadata = await parameterUtilities
-          .getMetadataForParameters(params, this.state.pipeline);
+          .getMetadataForParameters(params, pipeline, this.pipelinesLibrary);
```
rodichenko, line 5514: we need to be sure `pipelinesLibrary` is loaded here

AleksandrGorodetskii: done

Issue #4289, opened 2026-02-20
API shall respect external linux IDs when creating files/folders in NFS


sidoruka:
**Describe the bug**

API shall set UID/GID for the user as instructed by `launch.external.uid.field.name` preference.

PR #4299, opened 2026-02-25
Respect external linux IDs when creating files/folders in NFS

kamyshova:
This PR includes updates regarding #4289.


api/src/main/java/com/epam/pipeline/manager/datastorage/providers/nfs/NFSStorageProvider.java
```diff
@@ -92,6 +94,7 @@
  */
 @Service
 @Slf4j
+@SuppressWarnings("PMD.AvoidCatchingGenericException")
```
SilinPavel: do we need it here?

PR #4320, opened 2026-03-06
Try improve mount performance for buffered files

sidoruka:
### Summary
Previously, when disk buffering was enabled, files were downloaded sequentially in fixed-size chunks
(default 512 MB each) using a single thread before being served from local disk. This caused a delay
proportional to file size before any read could be served from the buffer.

With this change, files are downloaded in parallel using multiple workers (default 8). Each file is
split into equal chunks that are fetched concurrently via a thread pool, then written to the buffer
file in order. This reduces the time-to-first-read for large files significantly.

### Changes

**1. Parallel disk-buffer download (`pipefuse/diskread.py`)**
- Introduced a `download_workers` parameter (default 8 when invoked from pipe-fuse) for parallel full-file download.
- When `download_workers > 1`, file download is split into chunks and fetched in parallel via `ThreadPoolExecutor`; results are written to the buffer file in offset order.
- **Auto chunk size:**
  - If read-ahead/chunk size is not set and `download_workers > 1`, the file is split into `download_workers` equal chunks.
  - If `download_workers == 1`, a single chunk (whole file) is used.
- Added `_download_one_range()` to download a single byte range; sequential behavior is unchanged when `download_workers <= 1` or there is only one chunk.

**2. CLI and configuration (`pipe-fuse.py`)**
- New option: `--read-disk-buffer-download-workers` (default: 8).
  Env: `CP_PIPE_FUSE_READ_DISK_BUFFER_DOWNLOAD_WORKERS`.
- `--read-disk-buffer-read-ahead-size` is now optional (default: `None`). When unset and `download-workers > 1`, chunk size is derived automatically (equal chunks per file).
  Env: `CP_PIPE_FUSE_READ_DISK_BUFFER_READ_AHEAD_SIZE`.
- Safer env parsing for read-ahead size: only convert to int when the env value is non-empty.
- `DiskBufferingReadAllFileSystemClient` is instantiated with `download_workers=read_disk_buffer_download_workers`.

**3. S3 connection pool (`pipefuse/s3.py`)**
- S3 client `max_pool_connections` increased from default (10) to 50 so parallel disk-buffer downloads are not limited by the connection pool.

### Configuration

| CLI flag | Env variable | Default | Description |
|---|---|---|---|
| `--read-disk-buffer-download-workers` | `CP_PIPE_FUSE_READ_DISK_BUFFER_DOWNLOAD_WORKERS` | `8` | Number of parallel workers for full-file disk-buffer download (1 = sequential). |
| `--read-disk-buffer-read-ahead-size` | `CP_PIPE_FUSE_READ_DISK_BUFFER_READ_AHEAD_SIZE` | auto | Chunk size in bytes. If not set and `download-workers > 1`, chunks are auto-sized equally per file. If not set and `download-workers == 1`, the whole file is one chunk. |

### DIFF
--- a/pipe-cli/mount/pipefuse/diskread.py
+++ b/pipe-cli/mount/pipefuse/diskread.py
@@ -120,12 +120,8 @@ class DiskBufferingReadAllFileSystemClient(FileSystemClientDecorator):
                             # Collect results in offset order for sequential write
                             results_by_offset = {}
                             for future in as_completed(future_to_chunk):
-                                co, cl = future_to_chunk[future]
-                                try:
-                                    range_offset, data = future.result()
-                                    results_by_offset[range_offset] = data
-                                except Exception:
-                                    raise
+                                range_offset, data = future.result()
+                                results_by_offset[range_offset] = data
                             for range_offset, range_length in chunks:
                                 data = results_by_offset[range_offset]
                                 logging.info('Persisting buffer range %d-%d for %d:%s'
--- a/pipe-cli/mount/pipefuse/diskread.py
+++ b/pipe-cli/mount/pipefuse/diskread.py
@@ -15,6 +15,7 @@
 import io
 import logging
 import os
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from threading import Thread
 
 import errno
@@ -37,7 +38,7 @@ def mkdir(path):
 
 class DiskBufferingReadAllFileSystemClient(FileSystemClientDecorator):
 
-    def __init__(self, inner, read_ahead_size, path):
+    def __init__(self, inner, read_ahead_size, path, download_workers=1):
         """
         Disk buffering read all file system client decorator.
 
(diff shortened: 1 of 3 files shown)


pipe-cli/mount/pipefuse/diskread.py
```diff
@@ -61,20 +75,64 @@ def download_range(self, fh, buf, path, offset=0, length=0):
...
+                        workers = min(self._download_workers, len(chunks))
+                        with ThreadPoolExecutor(max_workers=workers) as executor:
+                            future_to_chunk = {
+                                executor.submit(self._download_one_range, fh, path, co, cl): (co, cl)
+                                for co, cl in chunks
+                            }
+                            # Collect results in offset order for sequential write
+                            results_by_offset = {}
+                            for future in as_completed(future_to_chunk):
+                                co, cl = future_to_chunk[future]
```
ekazachkova: It looks like this line is redundant. Does this code  affect further logic?

pipe-cli/mount/pipefuse/diskread.py
```diff
@@ -61,20 +75,64 @@ def download_range(self, fh, buf, path, offset=0, length=0):
...
+                                try:
```
ekazachkova: Do we really need try-except block here?

ekazachkova (commented):
A few minor comments that shall not affect logic

Issue #4368, opened 2026-04-13
Non-admin user can't launch run using pipe CLI command


maryvictol:
Non-admin user can't launch run using pipe CLI command.

**Steps to reproduce:**
1. Login as the non-admin user  
2. Open the **Tools** page 
3. Select any tool (for ex. `centos`)  
4. Launch run with Default settings  
5. Click a just launched run in the ACTIVE RUNS tab  
6. Wait until **SSH** hyperlink appears and click it  
7. In the opened tab enter and perform the command: `pipe run -di library/centos:latest -y` 

**Actual result:**
Error message is shown: `Error: Failed to fetch data from server. Server responded with message: Access Denied`

**Expected result:**
Response `Pipeline run scheduled with RunId: <runID>` is shown.

PR #4371, opened 2026-04-16
issue_4368_launch_reservation

kbashpayev:
Issue #4368, making preference - launch.reservation.parameter visible and not null.


api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
```diff
@@ -824,9 +824,9 @@ public class SystemPreferences {
                     new TypeReference<Map<String, ResourcesParameter>>() {},
                     LAUNCH_GROUP, isNullOrValidJson(new TypeReference<Map<String, ResourcesParameter>>() {}));
     public static final ObjectPreference<Map<String, Object>> LAUNCH_RESERVATION_PARAMS =
-            new ObjectPreference<>("launch.reservation.parameters", null,
+            new ObjectPreference<>("launch.reservation.parameters", new HashMap<>(),
```
SilinPavel: How about Collections.emptyMap()?

Issue #4327, opened 2026-03-13
Track usage/validity of the JWT tokens and allow to revoke them


sidoruka:
* Keep tokens' JTIs in the DB
* Allow list tokens (ID and expiration date)
* Allow revoke
* This shall be available from:
  * API
  * GUI
    * CLI -> Pipe CLI menu:
      * Generate token: optional token name (lets show in a popup with a valid date)
      * Token list: Name, ID, Expiration date, Revoke (button)
    * User details -> API Tokens (available to admins and those who have RW access to the user)
      * Token list: Name, ID, Expiration date, Revoke (button)
  * CLI
    * Listing: Name, ID, Expiration date, Revoke (button)
    * Revoke
    * Token name when generating
    * Support json output

sidoruka:
@AleksandrGorodetskii , please take care of the GUI

maryvictol:
@AleksandrGorodetskii, incorrect request at case of lists tokens for the current user: request should be without userId `/pipeline/restapi/user/token/named/list`

PR #4372, opened 2026-04-16
issue 4327 Named JWT tokens implementation

SilinPavel:
relates to #4327 

### Functionality

  • Plain (non-registered) JWTs are still issued with GET /user/token. The response stays a small object whose secret is in token; these tokens are not written to the named-token registry.
  • Named JWTs use the new named issue API: GET /user/token/named. You can pass expiration, tokenName (registry label for listing/admin), and optionally userId (numeric id) when an admin issues for another user. The
    response is one object that includes token (only at issue time) plus metadata such as jti, userId, tokenName, and expiry timestamps — so you get both the secret and the registry row in one step.
  • Listing named tokens is GET /user/token/named/list, with an optional userId for admin-style listing. Results are metadata only (no secret token values).
  • Revocation is DELETE /user/token/revoke with query jti (required) and optional userId for admin revoke on behalf of another user. Once revoked, the JWT stops working even if it was never a “named” row in the
    registry.

  ### Named tokens vs Revocation

  • Named tokens are an optional registry (a table in the database) that helps you see and manage certain tokens (labels, listing, admin actions).
  • Revocation is a separate blocklist: once a token id is revoked, the system rejects it, whether or not it appeared in the named-token list, this functionality can cover all JWTs.

### System preference (short summary)

  • launch.jwt.named.tokens.limit — caps how many named (registry) tokens a user may have. If the value is ≤ 0, that limit is not enforced. This only affects the named issuance path, not plain tokens.

 ### pipe-cli (short summary)

  • pipe token — issues a named token (registry label, optional duration, optional target user-id for admins).
  • pipe configure / --auth-token / token used for API auth — typically uses a plain token from GET /user/token (not registered as a named row by default).
  • pipe list-tokens — lists named tokens (metadata only, no secrets).
  • pipe revoke-tokens — revokes by jti (optional --user-id for admin-style revoke on behalf of another user).

  ### Safety

  • Token secrets are not stored in the listing data from the database; the `token` field is mainly for the moment you just created a token.


api/src/main/java/com/epam/pipeline/acl/user/UserApiService.java
```diff
@@ -232,6 +245,33 @@ public JwtRawToken issueToken(final String userName, final Long expiration) {
         return userManager.issueToken(userName, expiration);
     }
 
+    /**
+     * Issues a plain JWT for the current user (not registered in the named-token table).
+     */
+    @PreAuthorize(ADMIN_OR_GENERAL_USER)
+    public JwtRawToken issueTokenForCurrentUser(final Long expiration) {
```
ekazachkova, line 252: Do we really need this method?

api/src/main/java/com/epam/pipeline/manager/security/NamedJwtTokenManager.java
```diff
@@ -0,0 +1,170 @@
...
+public class NamedJwtTokenManager {
+
+    private final NamedJwtTokenDao namedJwtTokenDao;
+    private final AuthManager authManager;
+    private final UserManager userManager;
+    private final PreferenceManager preferenceManager;
+
+    @Autowired
+    public NamedJwtTokenManager(final NamedJwtTokenDao namedJwtTokenDao,
+                                @Lazy final AuthManager authManager,
```
ekazachkova: is lazy wiring still actual?

api/src/main/java/com/epam/pipeline/manager/security/NamedJwtTokenManager.java
```diff
@@ -0,0 +1,170 @@
...
+                                                @Nullable final String registryLabel) {
+        final UserContext userContext = userManager.loadUserContext(userName);
+        final JwtRawToken raw = userManager.issueToken(userName, expiration, false);
+        registerIssuedNamedJwtToken(raw, userContext.getUserId(), resolveTokenCreatedByUserId(userContext),
+                registryLabel);
+        return buildNamedJwtToken(raw, userContext.getUserId(), resolveTokenCreatedByUserId(userContext),
+                registryLabel, raw.getToken());
+    }
+
+    @Transactional(readOnly = true)
```
ekazachkova: Didn't get it: what is the difference between loadNamedJwtTokensForCurrentUser and loadNamedJwtTokensForUserAsAdmin?

pipe-cli/pipe.py
```diff
@@ -2034,14 +2034,47 @@ def split_tool_path(tool_path, registry, group, tool, version, strict=False):
...
+    as a named token (see list-tokens / revoke-tokens).
     """
-    UserTokenOperations().print_user_token(user_id, duration)
+    UserTokenOperations().print_user_token(user_id, duration, token_name)
+
+
+@cli.command(name='list-tokens')
+@click.option('--user-id', 'user_id', required=False, type=int,
+              help='List tokens registered for this user (requires access). Omit to list current user tokens.')
+@click.option('-o', '--output-format', type=click.Choice(['json']), default=None,
```
ekazachkova: maybe it would be better to keep short name is `-of` to support one style for all `--output-format` related places?

pipe-cli/src/utilities/user_token_operations.py
```diff
@@ -25,35 +27,97 @@ class UserTokenOperations(object):
...
+                row.get('userId'),
+                row.get('issuedAt'),
+                row.get('expiresAt'),
+            ])
+        click.echo(table.get_string())
+
+    def revoke_tokens(self, jtis, user_id=None, output_format=None):
+        cleaned = [j.strip() for j in jtis if j and str(j).strip()]
+        if not cleaned:
+            click.echo('Error: specify at least one non-empty -jti value.', err=True)
```
ekazachkova: In other places where the `--output-format` option is used, we typically also convert errors to JSON format. Should we probably continue to do this?

api/src/main/java/com/epam/pipeline/manager/security/NamedJwtTokenManager.java
```diff
@@ -0,0 +1,170 @@
...
+    public List<NamedJwtToken> loadNamedJwtTokensForUserAsAdmin(final Long userId) {
+        return namedJwtTokenDao.loadByUserId(userId);
+    }
+
+    private NamedJwtToken buildNamedJwtToken(final JwtRawToken rawToken, final Long userId,
+                                             final Long createdByUserId, final String tokenName,
+                                             @Nullable final String secretToken) {
```
ekazachkova: Probably, `showToken` flag would be clearer?

ekazachkova (approved):

Issue #4360, opened 2026-04-06
Capacity block: disk full issue


kamyshova:
**Describe the bug**

On the capacity block node, the `/ebs/docker/btrfs/` storage has completely filled up. This has triggered a **FreeDiskSpaceFailed** issue, causing **DiskPressure** conditions. As a result, new pods cannot be scheduled, and existing workloads may be affected.

PR #4383, opened 2026-04-21
Issue #4360: Support file system autoscaling for capacity blocks

ekazachkova:
relates to #4360 

### Background
Currently, Cloud Pipeline is not properly handles disk autoscaling for Capacity Blocks. The main reason is that `fsautoscale.sh` script queries Cloud Pipeline API `POST run/$_RUN_ID/disk/attach` method that is not compatible with Capacity Blocks:
- pricing is not applicaple
- compute resources cannot be found by Run ID tag
### API
Implemented a new API method `POST cluster/node/{name}/disk/attach` that creates a volume for specified instance. Currently, supported for AWS provider only. This method developed for capacity blocks disk autoscaling. 
### autoscaling
Implemented a new option `-i | --instance-type` for `fsautoscale` script that required to determine capacity block context. If so, a new API method shall be used to attach disk to node.

### DIFF
--- a/scripts/autoscaling/fsautoscale.sh
+++ b/scripts/autoscaling/fsautoscale.sh
@@ -164,7 +164,6 @@ function is_capacity_block_context() {
   _RESERVATION_PARAMS=$(resolve_system_preference "$_PREFERENCES" "launch.reservation.parameters" "{}")
   if echo "$_RESERVATION_PARAMS" | jq -e --arg t "$_INSTANCE_TYPE" 'type == "object" and has($t)' >/dev/null 2>&1
   then
-    pipe_log_debug "Capacity block context detected for instance type ${_INSTANCE_TYPE}; attaching disk by node name."
     return 0
   fi
   return 1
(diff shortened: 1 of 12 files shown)


scripts/autoscaling/fsautoscale.sh
```diff
@@ -316,6 +344,15 @@ then
 fi
 
 pipe_log_debug "Starting filesystem $MOUNT_POINT autoscaling process for node $NODE..."
+
+if is_capacity_block_context "$(get_system_preferences "$API" "$API_TOKEN")" "$INSTANCE_TYPE"
+then
+  pipe_log_debug "Capacity block context detected for instance type ${_INSTANCE_TYPE}; attaching disk by node name."
```
SilinPavel: should be `"$INSTANCE_TYPE"`?

scripts/autoscaling/init_multicloud.sh
```diff
@@ -351,7 +351,8 @@ RestartSec=10
 Environment="API_ARGS=--api-url $_API_URL --api-token $_API_TOKEN"
 Environment="NODE_ARGS=--node-name $_KUBE_NODE_NAME"
 Environment="MOUNT_POINT_ARGS=--mount-point $_MOUNT_POINT"
-ExecStart=/usr/bin/fsautoscale \$API_ARGS \$NODE_ARGS \$MOUNT_POINT_ARGS
+Environment="INSTANCE_TYPE_ARGS=$( [[ $cloud == *"EC2"* ]] && echo "--instance-type $_CLOUD_INSTANCE_TYPE" )"
```
SilinPavel: can we calculate INSTANCE_TYPE_ARGS out of this service file and then use it the same way as API_ARGS and others?

Issue #4374, opened 2026-04-17
Better support for Github integration


sidoruka:
Let's review available options for the Github repos integration. At the moment, we have to specify token directly to each pipeline. May be we can offload this to any sort of more automated option:
* Github apps?
* Separate functionality to maintain secrets in the Cloud Pipeline?

ekazachkova:
## Approach
To support "tokenless" pipeline creation a GitHub Apps can be used. The GitHub App shall be installed to organization or personal account. 

#### Authentication flow

1. Generate a JWT using the App's private key and App ID
2. Locate the installation for a specific repository or organization
3. Verify the installation is in the allowlist
4. Generate an installation access token for API operations
5. Call GitHub API as usual

## Implementation

#### Application Properties

A new application property `github.app.private.key.pem` (`$CP_GITHUB_APP_PRIVATE_KEY`) shall be supported. This value indicates a local path to private key pem file.

#### System Preferences 

The following preferences shall be configured:
- `github.app.id` - GitHub App identifier
- `github.allowed.installations` - allowlist of installation IDs or organization/account names, if empty access over GitHub App shall be denied.
- `github.api.root.url` - GitHub API base URL (default: `https://api.github.com`)

#### API methods
The following methods shall be accessed for `ROLE_USER`.
*Note*: Currently, these methods supported for GitHub repository type only. 

- `GET /pipeline/git/namespaces?type=GITHUB` - returns list with allowed installations:
```
[
    {
        "name": "<organization or personal account name>",
        "id": "<installation ID, for GitHub Apps the numeric value>",
        "type": "<User or Organization>"
    }, ...
]
```
These installation shall meet `github.allowed.installations` whitelist.
- `GET /pipeline/git/{namespaceId}/repositories?type=GITHUB` - returns list with git repos belongs to specified installation ID (`namespaceId`)
```
[
    {
      "id": "<123456>",
      "name": "<test>",
      "description": "<...>",
      "defaultBranch": "<main>",
      "httpUrl": "<https://github.com/org/repo.git>",
      "sshUrl": "<git@github.com:org/repo.git>",
      "path": "<https://github.com/org/repo>"
    }, ...
]
```
User shall select one of the listed available git repositories to associate with pipeline. To register pipeline server expects `httpUrl` as `repository`.

PR #4399, opened 2026-05-06
[API]: Support global github credentials

ekazachkova:
relates to #4374 

The current PR enables pipeline repository registration without the need to explicitly specify a GitHub token.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/git/github/GitHubUtils.java
+++ b/api/src/main/java/com/epam/pipeline/manager/git/github/GitHubUtils.java
@@ -36,12 +36,12 @@ public final class GitHubUtils {
         // no-op
     }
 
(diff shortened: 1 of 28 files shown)


api/src/main/java/com/epam/pipeline/manager/git/github/GitHubUtils.java
```diff
@@ -0,0 +1,57 @@
...
+    public static final String AUTHORIZATION = "Authorization";
+    public static final Integer LIMIT = 100;
+    private static final String LINK = "link";
+    private static final String REL_NEXT = "rel=\"next\"";
+
+    private GitHubUtils() {
+        // no-op
+    }
+
+    public static <T> List<T> pagination(final Function<Integer, Response<List<T>>> apiCall) {
```
SilinPavel: probably should be named not `pagination` but opposite, like `fetchAll` or `fetchAllPages`

Issue #4455, opened 2026-06-29
[Pipe CLI] NFS storage creation does not work


ekazachkova:
**Describe the bug**
NFS storage creation view `pipe storage create` command does not work. 

**To Reproduce**
1. Create storage with `pipe`:
```
pipe storage create --name "<storage name>" \
	--type NFS \
	--path "<NFS path>/<storage path>" \
	--description "< .. >" \
	--parent_folder "< .. >" \
	-c -sts < .. > -lts < .. > -b < .. > -u < .. >
```
2. See error:
```
Failed to create new datastorage. Server responded with message: Cannot invoke "java.lang.Long.longValue()" because "id" is null
```

This error occurs because the Cloud Pipeline API expects the `fileShareMountId ` to be specified. This value can be resolved from `--path` option value.

PR #4456, opened 2026-06-29
Issue #4455: [Pipe CLI] NFS storage creation does not work

ekazachkova:
relates to #4455

### DIFF
--- a/pipe-cli/src/utilities/datastorage_operations.py
+++ b/pipe-cli/src/utilities/datastorage_operations.py
@@ -19,11 +19,13 @@ import multiprocessing
 import os
 import platform
 import sys
+import re
 from botocore.exceptions import ClientError
 from future.utils import iteritems
 from operator import itemgetter
 import concurrent.futures
 
+from src.api.cloud_region import CloudRegion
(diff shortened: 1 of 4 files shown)


pipe-cli/src/utilities/datastorage_operations.py
```diff
@@ -536,8 +538,31 @@ def save_data_storage(cls, name, description, sts_duration, lts_duration, versio
...
+                click.echo("Error: Could not find a file share mount matching path '{}'.".format(path), err=True)
+                sys.exit(1)
         DataStorage.save(name, path, description, sts_duration, lts_duration, versioning, backup_duration, type,
-                         directory.id if directory else None, on_cloud, region_id)
+                         directory.id if directory else None, on_cloud, region_id, file_share_mount_id)
+
+    @classmethod
+    def _resolve_file_share_mount_id(cls, path):
+        server_port_mask = re.compile(r'^[^:]+(:[\d]+)?$', re.IGNORECASE)
+        regions = CloudRegion.get_cloud_regions()
```
SilinPavel, line 553: we have region_id in this case, probably better to load specific region

PR #4491, opened 2026-07-27
Dts java21 migration

kbashpayev:
migration of dts & commons:models to java 21


data-transfer-service/src/main/resources/application.properties
```diff
@@ -16,20 +16,23 @@
...
-spring.http.encoding.force-response=true
+server.servlet.encoding.charset=UTF-8
+server.servlet.encoding.force=true
+server.servlet.encoding.force-response=true
+server.error.include-message=always
 
 # DB
 spring.jpa.hibernate.ddl-auto=update
-spring.datasource.url=jdbc:h2:file:./db/dts
+spring.datasource.url=jdbc:h2:file:./db/v2.3/dts
```
SilinPavel: lets leave new db at the same folder level and move v2.3 to the filename: `v2_3`

SilinPavel (changes requested):

Issue #4516, opened 2026-08-13
InitializedNode step can handle several spot requests at a time


SilinPavel:
**Describe the bug**
actual: Behavior when InitializedNode task for a run report about several spot requests at a time was observed. It happens when node up face max spot request retries during spot run process.
 
expected: Only one spot requets should be followed at  a time

PR #4524, opened 2026-08-18
Issue #4516: InitializedNode step can handle several spot requests at a time

ekazachkova:
relates to #4516 

#### AutoscalManager
- if `ha.deploy.enabled=true`, check API pod is master before each instance type attempt

#### nodeup script
- fix for log message in check_spot_request_exists method
- fail terminal spot statuses immediately
- terminate node before labelling if another node with the same run-id already provisioned

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
@@ -578,12 +578,7 @@ public class AutoscaleManager extends AbstractSchedulingManager {
 
                 CmdExecutionException catchedCmdException = null;
                 for (String nodeType : allNodeTypesToTry) {
-                    if (StringUtils.equalsIgnoreCase(haDeployEnabled, Boolean.TRUE.toString())
-                            && !kubernetesManager.isMasterHost()) {
-                        log.warn("Cancelling scheduling node for run #{} since host is not master.", runId);
-                        // If this pod becomes a master again, a leftover in-progress task would still count toward
-                        // cluster size and nodeup tasks limit => this task shall be removed.
-                        removeNodeUpTask(longId, false);
+                    if (cancelNodeUpForNonMaster(longId, runId)) {
                         return;
                     }
                     try {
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManager.java
```diff
@@ -575,6 +578,14 @@ private void createNodeForRun(List<CompletableFuture<Void>> tasks, String runId,
 
                 CmdExecutionException catchedCmdException = null;
                 for (String nodeType : allNodeTypesToTry) {
+                    if (StringUtils.equalsIgnoreCase(haDeployEnabled, Boolean.TRUE.toString())
```
SilinPavel: lets move to separate method

api/src/test/java/com/epam/pipeline/manager/cluster/autoscale/AutoscaleManagerTest.java
```diff
@@ -126,7 +126,7 @@ public void setUp() throws Exception {
                 autoscalerService, nodesManager, kubernetesManager,
                 preferenceManager, TEST_KUBE_NAMESPACE, cloudFacade,
                 nodePoolManager, reassignHandler, scaleDownHandler, Collections.emptyList(), poolAutoscaler,
-                runRegionShiftHandler, metadataManager);
+                runRegionShiftHandler, metadataManager, Boolean.FALSE.toString());
```
SilinPavel, line 130: would be nice to add tests to check that autoscaler won't proceed with nodeup if it is not master

scripts/autoscaling/aws/nodeup.py
```diff
@@ -1124,9 +1130,9 @@ def get_spot_prices(ec2, aws_region, instance_type, hours=3):
 
 def exit_if_spot_unavailable(run_id, last_status):
     # will exit with code '5' if a spot request can't be fulfilled
-    if last_status in ['capacity-not-available', 'capacity-oversubscribed',
-                       'constraint-not-fulfillable', 'schedule-expired',
-                       'price-too-low', 'system-error']:
+    if (last_status in ['capacity-not-available', 'capacity-oversubscribed',
```
SilinPavel: lets also move rest of the statuses to separate list

scripts/autoscaling/aws/nodeup.py
```diff
@@ -1412,6 +1420,71 @@ def check_spot_request_exists(ec2, num_rep, run_id, time_rep, aws_region, pool_i
...
+        current = next((i for i in instances if i['InstanceId'] == ins_id), None)
+        pipe_log('Instance {} lost duplicate election for RunID {} (winner {}). Terminating.'.format(
+            ins_id, run_id, winner_id))
+        terminate_instance(ec2, ins_id,
+                           spot_request_id=_get_spot_instance_request_id(current) if current else None,
+                           kube_client=kube_client)
+        raise RuntimeError(
+            'Duplicate instance {} for RunID {} (winner {}). Terminated.'.format(ins_id, run_id, winner_id)
+        )
+    try:
```
SilinPavel: maybe better to move it under else, because now it is not clear that in this case ins_id is winner

Issue #4512, opened 2026-08-10
Evolve mechanism of IDLE runs


SilinPavel:
**Background**   
 `ResourceMonitoringManager` monitors running pipeline runs for resource underutilization. When a run's CPU usage stays below a configured threshold for a configured period, the system marks it as `IDLE` and takes a configured action: notify, pause, or stop. This state is tracked via a `lastIdleNotificationTime` field on `PipelineRun` and an `IDLE` tag visible in the UI.
 
**Problem** 
The current idle detection is CPU-only. Runs that have GPUs can be CPU-idle but actively utilizing their GPU — the system treats them as idle and may incorrectly pause or stop them. There is no way to configure different idle behavior for GPU workloads vs CPU workloads, and no visibility into which resource type is driving the idle state.
 
**What Needs to Be Done**
 1. **Split idle detection into per-resource types**: `IDLE_CPU` and `IDLE_GPU`, each with its own threshold, action timeout, and action configuration. GPU idle detection only applies to runs on GPU-enabled instance types.
 2. **Introduce an aggregate `IDLE` state**: a run is fully idle when both `IDLE_CPU` and `IDLE_GPU` are true (or when it has no GPU and is `IDLE_CPU`). The aggregate has its own independent action configuration.
               **a**. In a situation when one of the configs (CPU/GPU) is disabled and run is idle for second resource type, such run shouldn't be marked as IDLE
 3. **Each idle type acts independently**: `IDLE_CPU`, `IDLE_GPU`, and `IDLE` each have their own action (notify/pause/stop) and their own action timeout.
 4. The notification timestamps (`lastIdleNotificationTime`, `lastNetworkConsumptionNotificationTime`, `lastNotificationTime`) are hardcoded columns on the `PipelineRun` entity, making it difficult to extend the monitoring model without schema changes each time.**Use additional tags like IDLE_CPU_date IDLE_GPU_date IDLE_date** to verify if we need to fire and action, etc instead of lastIdleNotificationTime.

5. **Replace all existing idle system preferences** with a single `system.monitoring.idle.settings` preference that holds a list of per-type configuration objects. All of the following existing preferences must be removed:
   - `system.max.idle.timeout.minutes`
   - `system.idle.cpu.threshold.percent`
   - `system.idle.action.timeout.minutes`
   - `system.idle.action`

   The new preference value is a JSON array where each entry configures one idle type:

   ```json
   [
     {
       "type": "CPU",
       "enabled": true,
[...]

PR #4533, opened 2026-08-19
Issue 4512 Evolve idle run mechanism

SilinPavel:
Implementation of #4512


api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -359,10 +363,32 @@ private Map<ELKUsageMetric, Double> getThresholds() {
...
         }
 
-        private void processIdleRuns(final List<PipelineRun> runs) {
+        private void processIdleRuns(final List<PipelineRun> runs,
+                                     final IdleMonitoringConfig.IdleMonitoringType monitoringType,
+                                     final ELKUsageMetric usageMetric) {
             final Map<String, PipelineRun> running = groupedByNode(runs);
 
-            final int idleTimeout = preferenceManager.getPreference(SystemPreferences.SYSTEM_MAX_IDLE_TIMEOUT_MINUTES);
+            Map<IdleMonitoringConfig.IdleMonitoringType, IdleMonitoringConfig> enabledIdleConfigMap =
```
SilinPavel: we could perform it once before calling 

```
processIdleRuns(runs, IdleMonitoringConfig.IdleMonitoringType.CPU, ELKUsageMetric.CPU);
processIdleRuns(runs, IdleMonitoringConfig.IdleMonitoringType.GPU, ELKUsageMetric.GPU_AGGS);
```

and based on active config run specific process method

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -359,10 +363,32 @@ private Map<ELKUsageMetric, Double> getThresholds() {
...
+
+            if (Objects.isNull(processorUnitConf.gracePeriodMinutes())
+                    || Objects.isNull(processorUnitConf.actionTimeoutMinutes())) {
+                log.warn("{} idle monitoring config misses grace period or action timeout, skipping idle check.",
+                        monitoringType.name());
+                return;
+            }
+            final int idleTimeout = processorUnitConf.gracePeriodMinutes();
 
             final Map<String, PipelineRun> notProlongedRuns = running.entrySet().stream()
```
SilinPavel: `notProlongedRuns` can be passed to this method instead of `runs` since we do this filter the same way two times

api/src/main/java/com/epam/pipeline/dao/monitoring/metricrequester/GPURequester.java
```diff
@@ -0,0 +1,81 @@
+/*
```
SilinPavel: we already have GPUAggregationRequester where we do not have implementation for methods to request loadMetrics in MonitoringESDao let's implement it instead of creation a new class

api/src/main/java/com/epam/pipeline/manager/preference/SystemPreferences.java
```diff
@@ -1209,20 +1211,34 @@ CLUSTER_GROUP, isGreaterThan(0).or(isEquals(-1)), true
     /**
      * Controls maximum timeout (in minutes), which a node can stay idle, before an action will be taken
      */
+    @Deprecated
```
SilinPavel: why not just delete these old preferences?

api/src/main/java/com/epam/pipeline/manager/cluster/performancemonitoring/ResourceMonitoringManager.java
```diff
@@ -169,7 +172,8 @@ public void removeOldIndices() {
         @SchedulerLock(name = "ResourceMonitoringManager_monitorResourceUsage", lockAtMostForString = "PT10M")
         public void monitorResourceUsage() {
             List<PipelineRun> runs = pipelineRunManager.loadRunningPipelineRuns();
-            processIdleRuns(runs);
+            processIdleRuns(runs, IdleMonitoringConfig.IdleMonitoringType.CPU, ELKUsageMetric.CPU);
```
SilinPavel: I would better define 2 separate methods, for each of the CPU/GPU metric type.
Then inside these method I would reuse parts of the code that are common, for example:
 - shouldPerformActionOnIdleRun
 - performActionOnIdleRun
 - etc

but in this case we won't have such constructions like:
```
if (monitoringType == IdleMonitoringConfig.IdleMonitoringType.CPU) {
....              
} else {
....
}
```

SilinPavel:
Note: Since now we store notification timestamps in additional tags like `IDLE_CPU_date` etc, we should reset these tags in PipelineRunManager.prolongIdleRun

Issue #4550, opened 2026-08-31
Add possibility to use user run tag value in the Usage Credits rule statements


maryvictol:
parent issue: https://github.com/epam/cloud-pipeline/issues/4461

Now in the Usage Credits rules we can check only existence of a specific tag. 
It would be useful to add possibility to use in the Usage Credits rules statements a check that a user run tag with the specified name not only exists, but also has a specific value.

Similar approach already used for other run fields such as parameters or env vars, we can reuse that approach

PR #4565, opened 2026-09-04
issue_4550_edit_tag_evaluation

kbashpayev:
Issue #4550, Add possibility to use user run tag value in the Usage credits rule statements.

### DIFF
--- a/cloud-pipeline-common/model/src/main/java/com/epam/pipeline/utils/condition/field/PipelineRunField.java
+++ b/cloud-pipeline-common/model/src/main/java/com/epam/pipeline/utils/condition/field/PipelineRunField.java
@@ -23,6 +23,7 @@ import com.epam.pipeline.entity.pipeline.PipelineRun;
 import com.epam.pipeline.entity.pipeline.RunInstance;
 import com.epam.pipeline.entity.pipeline.run.parameter.PipelineRunParameter;
 import lombok.Getter;
+import org.apache.commons.collections4.MapUtils;
 
 import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


cloud-pipeline-common/model/src/main/java/com/epam/pipeline/utils/condition/field/PipelineRunField.java
```diff
@@ -243,6 +244,11 @@ private static Map<String, String> envVarsToMap(final PipelineRun run) {
         return envVars != null ? envVars : Collections.emptyMap();
     }
 
+    private static Map<String, String> tagsToMap(final PipelineRun run) {
```
SilinPavel: we can inline his function instead of  `PipelineRunField::tagsToMap` like `MapUtils.emptyIfNull(run.getTags())`

core/src/main/java/com/epam/pipeline/utils/condition/field/PipelineRunField.java
```diff
@@ -243,6 +244,11 @@ private static Map<String, String> envVarsToMap(final PipelineRun run) {
         return envVars != null ? envVars : Collections.emptyMap();
     }
 
+    private static Map<String, String> tagsToMap(final PipelineRun run) {
```
SilinPavel: same as in https://github.com/epam/cloud-pipeline/pull/4565/changes#diff-1cccbe7ad34b56509eabe45ddcf965f97407e86802753fdcc49ffb8a4b1c7780R247-R249

Issue #4551, opened 2026-08-31
Expand visibility and logs for runs running on capacity block during scheduling


SilinPavel:
**Background**
Currently there are very limited number of logs during scheduling phase of runs on capacity block. Would be nice to expand server logging to see current status if such pods on each autoscaler cycle for example.

PR #4568, opened 2026-09-09
issue_4551_log_unschedulable_pods

kbashpayev:
Issue #4551, add logging to unschedulable pods.

### DIFF
--- a/api/src/main/java/com/epam/pipeline/manager/cluster/PodMonitor.java
+++ b/api/src/main/java/com/epam/pipeline/manager/cluster/PodMonitor.java
@@ -229,7 +229,8 @@ public class PodMonitor extends AbstractSchedulingManager {
                     .map(entry -> entry.getKey() + "=" + entry.getValue())
                     .collect(Collectors.joining(", "));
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/epam/pipeline/manager/cluster/PodMonitor.java
```diff
@@ -208,6 +213,31 @@ public void updateStatus() {
...
+            if (!reason.isPresent()) {
+                return;
+            }
+
+            final String selector = MapUtils.emptyIfNull(pod.getSpec().getNodeSelector())
+                    .entrySet().stream()
+                    .map(entry -> entry.getKey() + "=" + entry.getValue())
+                    .collect(Collectors.joining(", "));
+
+            final String logText = String.format("Run is waiting for a node matching [%s]: %s", selector, reason.get());
```
SilinPavel: Probably it could be another reason why pod can't be sheduled, so lets change message to `Pod for run with node selector [%s] is in Unschedulable state with reason %s`

SilinPavel (changes requested):

Issue #4127, opened 2025-10-26
Incorrect warning about permissions for the docker images, used in a pipeline


maryvictol:
_Edit pipeline info_ pop-up > _Permissions_ tab : GROUP permissions aren't have priority over ROLE permissions at calculate permissions for the docker images, used in a pipeline.

**To Reproduce**
1. Added non-admin user to the `Group1` and give `Role1` to him
2. Open any Tool (`tool1`) and  
- grant _Allow_  `Read` and `Execute` permissions to the `Role1` 
- grant _Allow_  `Read` and _Deny_ `Execute` permissions to the `Group1`

3. Create pipeline (`pipeline1`)  with `tool1` docker image
4. Grant all permissions on `pipeline1` to the non-admin user from step 1

**Actual behavior**
Warning _Users shall have Read and Execute permissions for the docker images, used in a current pipeline. Please review and fix permissions issues below: <`tool1`>: execute denied for <non-admin user>_  appears in the Permissions tab.

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/17095909-09d1-4f65-9dcb-7a430d3b676e" />

**Expected behavior**
Non-admin user has `Execute` permissions to the `tool1` as permissions that were granted for the GROUP should have bigger priority that permission for ROLE (#4110).

**Environment:**
 - Cloud Provider:  AWS
 - Version 0.17.0.21605

PR #4582, opened 2026-09-21
Issue #4127: Resolve SIDs as the API does in the sub-objects permissions warning

rodichenko:
Implements #4127

The _Permissions_ tab's sub-objects warning ("Users shall have Read and Execute permissions for the docker images…") merged every SID of a user into a single bucket where any **deny** won, so a system role denying Execute raised a warning even though one of the user's groups allows it. It now resolves SIDs the way `PermissionGrantingStrategyImpl` does — the user itself, then its groups, then its system roles, with the first tier that rules on a permission deciding it and a deny beating an allow only within a tier. A group is a non-predefined role plus the user's external groups; the client already receives `predefined` on both `/role/loadAll` and `/users/info`, so this is a GUI-only fix.

The manual documented only "an explicit user rule beats a group's or a role's"; the group-over-role tier and the deny-within-a-tier rule are added in the `Docs:` commit.

Test plan
- [x] `npm test` — 182 tests pass; 8 new unit tests on `checkObjectPermissionsConflict` plus a new `PermissionsForm.test.js` driving the warning through the component. Reverting the two sources fails exactly the three #4127 cases.
- [x] `npm run lint`, `npm run stylelint` — no new problems; baselines untouched.
- [x] Live UI (`verify-client-live`, dev server against an AWS deployment): pipeline → gear → _Edit_ → _Permissions_ renders a `ROLE_USER` grant through the new non-principal path with no spurious warning.
- [ ] The #4127 conflict pattern itself in a browser — of 640 pipelines on that deployment, 15 have grants and none exhibits it, and manufacturing one would mean editing ACLs on a shared deployment. Covered by the tests only.

Not addressed here: the scoped admin roles (`ROLE_TOOL_ADMIN` and friends) are not exempted from this warning. Pre-existing, and worth its own issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


client/src/components/roleModel/PermissionsForm.js
```diff
@@ -450,16 +469,22 @@ export default class PermissionsForm extends React.Component {
       const {mask, sid = {}} = granted[d];
       const maskToCheck = mask & subObjectsPermissionsMaskToCheck;
       const {name, principal} = sid;
+      const sidToCheck = principal
+        ? sid
+        : {...sid, predefined: this.isPredefinedRole(name)};
```
rodichenko, line 474: `this.isPredefinedRole(name)` reads the `predefinedRoleNames` computed getter, which depends on `this.props.roles` — the first code path that does, so `/role/loadAll` is now triggered at mount instead of lazily. That races against the `subObjectsPermissions` fetch this method also depends on.

**Failure scenario:** a pipeline grants a permission directly to a system role (not a group). If `subObjectsPermissions` resolves before `/role/loadAll` does, `isPredefinedRole(name)` returns `false` (`predefinedRoleNames` is still `[]`), so `getSidType` misclassifies that role as a `group`. If the granted role's tool rule is `ALLOW` while `ROLE_USER`'s rule is `DENY`, the misclassified `group` tier resolves alone instead of merging with `ROLE_USER` in the `role` tier (deny-wins) — a real sub-objects conflict warning is transiently missing (or a spurious one transiently appears) until roles finish loading and MobX re-renders.

client/src/utils/roleModel.js
```diff
@@ -262,37 +284,34 @@ const checkObjectPermissionsConflict = (mask, sid, sidRoles, objectOwner, object
...
-        merged[permission] = [false];
+  const conflictsBySidType = [sid, ...(sidRoles || []).map((sidRole) => ({
+    ...sidRole,
+    principal: false
+  }))]
+    .reduce((acc, testSid) => {
+      const type = getSidType(testSid);
+      return {
+        ...acc,
+        [type]: [...(acc[type] || []), findConflicts(testSid)]
```
rodichenko: `conflictsBySidType` is built via `{...acc, [type]: [...(acc[type]||[]), ...]}` inside `reduce`, copying the whole accumulator object on every SID instead of mutating a plain object/Map.

Not a correctness bug (sidRoles lists are always small), but it's O(n²) allocation and reads as more complex than a plain `acc[type] = acc[type] || []; acc[type].push(...)` loop.

client/src/components/roleModel/PermissionsForm.js
rodichenko: Retracting this: I chased this further while trying to fix it, and the premise doesn't hold up.

A completely unrelated, pre-existing method (`findGroupDataSource`, used for the "find group" popup's `<AutoComplete>` data source) already reads `this.props.roles.loaded` unconditionally on *every* render, including the very first one at mount — before `componentDidMount` even runs. So `/role/loadAll` was already being kicked off immediately, regardless of this PR; `isPredefinedRole` isn't "the first code to touch `this.props.roles`" as I claimed. I confirmed this empirically: a test asserting `/role/loadAll` gets called on mount with no permissions granted (so nothing else would plausibly trigger it) passes identically whether or not a fetch is added to `componentDidMount`.

Separately, I also worked out the failure-mode math for `checkObjectPermissionsConflict`'s tier resolution and confirmed a guard wouldn't be safe to add anyway: misclassifying a directly-granted role as a group during the loading window can only ever cause a warning to be **transiently hidden that should show** (self-correcting upward once roles load) — never a warning shown that must later disappear. A blanket "don't render anything until roles have loaded" guard doesn't change the observable outcome in the one case where a discrepancy exists (both show nothing during that window), while it does delay warnings that render correctly today without waiting on `/role/loadAll` at all (e.g. when the granted role has no rule of its own on the sub-object).

No change needed here; leaving `PermissionsForm.js` as is.

PR #4583, opened 2026-09-21
java21_issue_4552_granular_pricing

kbashpayev:
add status info for run_status_change table in order to calculate run prices more granularly.


api/src/main/java/com/epam/pipeline/dao/pipeline/RunStatusDao.java
```diff
@@ -75,18 +80,31 @@ public void deleteRunStatusByRunIdsIn(final List<Long> runIds, final boolean dry
         getNamedParameterJdbcTemplate(dryRun).update(deleteRunStatusByIdsQuery, params);
     }
 
+    @Transactional(propagation = Propagation.MANDATORY)
+    public void updatePriceForCurrentActiveRecord(PipelineRun pipelineRun) {
```
SilinPavel, line 84: let's be more precise, we don't have run record object, but do have RunStatus, so -  updatePriceForCurrentActiveRunStatus

api/src/main/java/com/epam/pipeline/manager/pipeline/PipelineRunDockerOperationManager.java
```diff
@@ -169,9 +170,9 @@ public PipelineRun pauseRun(Long runId, boolean checkSize) {
      * @return resumed {@link PipelineRun}
      */
     @Transactional(propagation = Propagation.REQUIRED)
-    public PipelineRun resumeRun(final Long runId) {
+    public PipelineRun resumeRun(final Long runId, PipelineRun pipelineRun) {
```
SilinPavel, line 173: why do we need to pass run here? we are loading it in `loadRunForPauseResume`, and since it happening in the same transaction we will see new prices here, then when we pass it in ` dockerContainerOperationManager.resumeRun(pipelineRun, ..)` `dockerContainerOperationManager` will use this object to get prices to update a status

SilinPavel (changes requested):

billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/service/impl/converter/RunToBillingRequestConverter.java
```diff
@@ -266,17 +269,66 @@ private List<PipelineRunBillingInfo> createBillingsForPeriod(final PipelineRunWi
...
+        // It is the only costs basis of an old fashioned segment, therefore it shall not be zeroed out.
+        // At the same time it is safe to take from the run level price, since it is read for the segments
+        // which have no compute and disk prices in the snapshot at all.
+        final BigDecimal pricePerHour = Optional.ofNullable(snapshot.getPricePerHour())
+                .map(this::scaledForUser)
+                .orElseGet(runPrice::getOldFashionedPricePerHour);
+        final BigDecimal computePricePerHour = scaledForBilling(snapshot.getComputePricePerHour());
+        // The run level disk price is not used as a fallback here, since it belongs to the latest
+        // instance configuration rather than to the one the segment has been executed on.
+        final BigDecimal diskPricePerHour = scaledForBilling(
```
SilinPavel: I believe we should get disk price from run when in status it is absent, otherwise we will calculate whole disk price as zero

billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/service/impl/converter/RunToBillingRequestConverter.java
```diff
@@ -266,17 +269,66 @@ private List<PipelineRunBillingInfo> createBillingsForPeriod(final PipelineRunWi
...
+
+        final RunPricePerHour snapshot = statusInfo.getPrice();
+        if (snapshot.getComputePricePerHour() == null) {
+            log.warn("[RUN_ID:{}] Invalid price snapshot (computePricePerHour is null), "
+                            + "falling back to run-level price",
+                    status.getRunId());
+            return runPrice;
+        }
+
+        // Snapshots taken before the price per hour has been stored fall back to the run level price.
```
SilinPavel: I'm not sure about relevance of this comment. It is just our strategy, that we are taking run price as a fallback (because it is how system worked before this changes)

billing-report-agent/src/main/java/com/epam/pipeline/billingreportagent/service/impl/converter/RunToBillingRequestConverter.java
```diff
@@ -266,17 +269,66 @@ private List<PipelineRunBillingInfo> createBillingsForPeriod(final PipelineRunWi
...
+        final RunStatusInfo statusInfo = status.getRunStatusInfo();
+        if (statusInfo == null || statusInfo.getPrice() == null) {
+            log.debug("[RUN_ID:{}] No price snapshot in status, using run-level price: compute=${}, disk=${}",
+                    status.getRunId(), runPrice.getComputePricePerHour(), runPrice.getDiskPricePerHour());
+            return runPrice;
+        }
+
+        final RunPricePerHour snapshot = statusInfo.getPrice();
+        if (snapshot.getComputePricePerHour() == null) {
+            log.warn("[RUN_ID:{}] Invalid price snapshot (computePricePerHour is null), "
```
SilinPavel: for inactive period wouldn't it be always absent? If yes, probably we should rewrite this somehow.

api/src/main/resources/dao/run-status-dao.xml
```diff
@@ -110,5 +116,25 @@
                 ]]>
             </value>
         </property>
+        <property name="updatePriceInLastRunStatusChangeQuery">
+            <value>
+                <![CDATA[
+                    UPDATE pipeline.run_status_change
+                    SET run_status_info = jsonb_build_object(
+                        'price', jsonb_build_object(
```
SilinPavel: take a look how we work with jsonb in other cases, we don't need to map each field of json as a separate value, we can work with json as with an object on java side and here treat it just as a jsonb

SilinPavel (changes requested):

